PR #4, opened 2016-02-08
Fix issue 2

wence-:


miklos1:
This is OK. Is there any way we can test this?

wence-:
So what bit do you want to test?  That the answer is correct?  Or that the evaluation of bits are merged?

miklos1:
That they are actually merged.

wence-:
OK, I will have a go.

PR #6, opened 2016-02-09
Separate the handling of reference geometry from finite elements

miklos1:


tsfc/fem.py
```diff
@@ -437,11 +436,24 @@ def __init__(self, weights, quadrature_index, argument_indices, tabulation_manag
         self.quadrature_index = quadrature_index
         self.argument_indices = argument_indices
         self.tabulation_manager = tabulation_manager
+        self.integral_type = tabulation_manager.integral_type
         self.coefficient_map = coefficient_map
         self.cell_orientations = False
         self.facet = tabulation_manager.facet
         self.index_cache = index_cache
 
+    def get_cell_orientations(self):
```
wence-: Could be a memoized property?

miklos1: Technically it could, but I prefer to function call syntax instead of the property / attribute syntax, because calling this function has a side effect that we actually use: `self.cell_orientations` is set `True`, so we know we need cell orientations in the kernel. (Though this is still a violation of command-query separation.)

wence-:
OK.

PR #8, opened 2016-02-10
Add an interpreter for GEM

wence-:
Takes a GEM tree with some optional data bindings for Variable and
VariableIndex nodes and evaluates it.

Lightly tested!

### DIFF
--- a/tsfc/interpreter.py
+++ b/tsfc/interpreter.py
@@ -1,0 +1,306 @@
+"""
+An interpreter for GEM trees.
+"""
+from __future__ import absolute_import
+
+import numpy
+import operator
+import math
+from singledispatch import singledispatch
+import itertools
+
+from tsfc import gem, node
+
+__all__ = ("evaluate", )
+
+
+class Result(object):
+    """An array object that tracks which axes of the array correspond to
+    gem free indices (and what those free indices are).
+
+    :arg arr: The array.
+    :arg fids: The free indices.
+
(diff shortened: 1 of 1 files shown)


wence-:
I haven't thought hard about where to allow hooking this in to the compilation pipeline.  It works on the tree that comes out of fem.process either before or after simplification passes.

miklos1:
We could hook it up with the PyOP2 interpreter.

tsfc/interpreter.py
```diff
@@ -0,0 +1,306 @@
...
+    idx = val.fids.index(e.index)
+    return Result(val.arr.sum(axis=idx),
+                  val.fids[:idx] + val.fids[idx+1:])
+
+
+@_evaluate.register(gem.ListTensor)  # noqa: not actually redefinition
+def _(e, self):
+    """List tensors just turn into arrays."""
+    ops = [self(o) for o in e.children]
+    assert all(ops[0].fids == o.fids for o in ops)
```
miklos1: This is true for ListTensors where were translated from UFL. GEM doesn't have this restriction though.

miklos1: Or is it true? It may not be when the FormSplitter is involved.

wence-: I can fix this by relaxing the assertion.

wence-: But that requires me to think too hard in my hazy state.

PR #11, opened 2016-02-11
Some clean up in fem.py

miklos1:
Mostly clean up and some new docstrings. Other changes:
 * Cellwise constant Arguments are now treated as cellwise constant.
 * Cellwise constant FormArguments are the same on all facets, thus they don't a facet number to look up the right tabulation matrix.
 * Cellwise constant Coefficient evaluation is expanded if the tabulation vector contains at most 2 nonzeros. This way the generated code for the evaluation of the Jacobian on P1 simplices is equivalent to the code snippets we earlier had.
 * Simplification of absolute values and cell orientations is rewritten, using the saner DAG visitor infrastructure from `tsfc.node`.  This also demonstrates the generality of the DAG visitor infrastructure, as it can handle both UFL and GEM.
 * No more `fabs(sqrt(...))`

### DIFF
--- a/tsfc/fem.py
+++ b/tsfc/fem.py
@@ -12,10 +12,13 @@ from ufl.corealg.multifunction import MultiFunction
 from ufl.classes import (Argument, Coefficient, FormArgument,
                          GeometricQuantity, QuadratureWeight,
                          ReferenceValue, Zero)
+from ufl.classes import (Abs, CellOrientation, Expr, FloatValue,
+                         Division, Product, ScalarValue)
 
 from tsfc.constants import PRECISION
 from tsfc.fiatinterface import create_element, as_fiat_cell
 from tsfc.modified_terminals import is_modified_terminal, analyse_modified_terminal
+from tsfc.node import MemoizerArg
 from tsfc import gem
 from tsfc import ufl2gem
 from tsfc import geometric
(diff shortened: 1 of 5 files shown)


tsfc/fem.py
```diff
@@ -28,7 +29,12 @@
 
 
 class ReplaceSpatialCoordinates(MultiFunction):
+    """Replace SpatialCoordinate nodes with the ReferenceValue of a
+    Coefficient.  Assumes that the coordinate element only needs
+    affine mapping.
```
wence-: One could assert this by doing `assert o.ufl_coordinate_element().mapping() == "identity"` in the `spatial_coordinate` handler.

tsfc/fem.py
```diff
@@ -157,421 +169,354 @@ def spanning_degree(element):
...
-            else:
-                result = ein.Variable("cell_orientations", (1, 1))
-            self.cell_orientations = True
-            self._cell_orientations = result
-            return result
+        if integral_type in ['exterior_facet', 'exterior_facet_vert']:
+            self.facet = {None: gem.VariableIndex('facet[0]')}
+        elif integral_type in ['interior_facet', 'interior_facet_vert']:
+            self.facet = {'+': gem.VariableIndex('facet[0]'),
+                          '-': gem.VariableIndex('facet[1]')}
```
wence-: I think I noticed this before.  Is there a reason you don't use `gem.Indexed(gem.Variable("facet"), (0,))`, etc... here?

tsfc/fem.py
miklos1: Thanks for reminding, I'll add that.

tsfc/fem.py
miklos1: I agree it's ugly, and I plan to go in that direction. However, I don't plan to use GEM expressions as indices directly. Indices are special, and there are three kinds of them: fixed indices (int), free indices (Index), and variable indices (VariableIndex; they are runtime constant for a single execution of the kernel). I might introduce a common base class (IndexBase) for them. What I plan to do at one point is to make VariableIndex wrap a GEM expression (with no free indices and no shape) instead of string. That is, to use

    gem.VariableIndex(gem.Indexed(gem.Variable("facet"), (0,)))

instead of

    gem.VariableIndex('facet[0]')

wence-: OK, that makes sense.

tsfc/fem.py
miklos1: Done.

tsfc/node.py
```diff
@@ -98,7 +98,8 @@ def get_hash(self):
 def traversal(expression_dags):
     """Pre-order traversal of the nodes of expression DAGs."""
     seen = set(expression_dags)
-    lifo = list(expression_dags)
+    # Some roots might be same, but they must be visited only once
+    lifo = list(seen)
```
wence-: Careful, `lifo` might come out in a different order on different executions.  Does that matter?

miklos1: Not sure, it might. Better be safe, I'll fix this.

Issue #12, opened 2016-02-12
compile_form not idempotent


wence-:
Consider the following:

```python
from firedrake import *
from tsfc import compile_form
m = UnitCubeMesh(1, 1, 1)
V = VectorFunctionSpace(m, 'CG', 1)
f = Function(V)

k1 = compile_form(dot(f, f)*dx)[0]
k2 = compile_form(dot(f, f)*dx)[0]

assert k1.ast.gencode() == k2.ast.gencode()
```

This is because indices that are not given a `name` are initialised with a class-based incremented `count`.  Now, the on-disk kernel caches don't actually use the generated code, so this might not matter.  It's a bit funky though.

miklos1:
I was aware of this, but did not find it critical.

wence-:
So it's a bit funky, because imagine I clear out the kernel cache, but not the generated code cache.  And then compile some forms in a different order.  I end up needing to compile wrapper code again.

PR #13, opened 2016-02-12
Idempotency

wence-:
Make compile_form idempotent, and add a few tests thereof.  c.f. #12.

### DIFF
--- a/tests/test_idempotency.py
+++ b/tests/test_idempotency.py
@@ -1,0 +1,54 @@
+import ufl
+from tsfc import compile_form
+import pytest
+
+
+@pytest.fixture(params=[ufl.interval,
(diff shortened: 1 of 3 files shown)


miklos1:
Nice test, we don't actually need any data to compile a form, so no data is created.

miklos1:
I see you're changing `Index.__repr__` and `Index.__slots__`.

wence-:
One could avoid changing `Index` and instead do the following in `coffee.py`

```python
 def _index_name(index, parameters):
     if index.name.startswith("i_"):
         return parameters.index_names[index]
     return index.name
```

But this felt somewhat fragile.

miklos1:
Ah, I see. OK, go ahead.

PR #14, opened 2016-02-12
Optimise generated code

miklos1:
A few optimisations that help to achieve assembly performance on par with our old FFC on affine cells:
 * Accumulating directly into the output tensor instead of using a temporary and then later copying into the output tensor.
 * Alternative mode of handling mixed element coefficients in interior facet integrals: assemble ListTensor instead "manually" generating COFFEE loops.
 * This alternative way is enabled on the coordinates on simplex cells. Combined with the previous optimisation expanding the evaluation of cellwise constant coefficients, this gives "code snippet"-equivalent code even in interior facet integrals.
 * Don't expand quadrature loops.

### DIFF
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -173,14 +173,13 @@ def compile_integral(idata, fd, prefix, parameters):
                              tabulation_manager, quad_rule.weights,
                              quadrature_index, argument_indices,
                              coefficient_map, index_cache)
+        nonfem = opt.unroll_indexsum(nonfem, max_extent=3)
         nonfem_.append([(ein.IndexSum(e, quadrature_index) if quadrature_index in e.free_indices else e)
                         for e in nonfem])
 
     # Sum the expressions that are part of the same restriction
     nonfem = list(reduce(ein.Sum, e, ein.Zero()) for e in zip(*nonfem_))
-
     simplified = opt.remove_componenttensors(nonfem)
-    simplified = opt.unroll_indexsum(simplified, max_extent=3)
 
     cell_orientations = False
     refcount = sch.count_references(simplified)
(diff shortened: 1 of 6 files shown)


tsfc/driver.py
```diff
@@ -100,7 +100,10 @@ def compile_integral(idata, fd, prefix, parameters):
 
     mesh = idata.domain
     coordinates = fem.coordinate_coefficient(mesh)
-    funarg, prepare_, expression = prepare_coefficient(integral_type, coordinates, "coords")
+    if mesh.ufl_cell().cellname() in ["interval", "triangle", "tetrahedron"]:
+        funarg, prepare_, expression = prepare_coefficient(integral_type, coordinates, "coords", mode='list_tensor')
```
wence-: If I understand the description correctly, you want to do this for affine coordinates?  But this is not the correct check for that.

tsfc/driver.py
```diff
@@ -241,7 +243,11 @@ def compile_integral(idata, fd, prefix, parameters):
     return kernel
 
 
-def prepare_coefficient(integral_type, coefficient, name):
+def prepare_coefficient(integral_type, coefficient, name, mode=None):
+    if mode is None:
+        mode = 'manual_loop'
+
+    assert mode in ['manual_loop', 'list_tensor']
```
wence-: Is there a reason not to always use one over the other (and when should one make the choice?)

tsfc/driver.py
```diff
@@ -291,30 +297,54 @@ def prepare_coefficient(integral_type, coefficient, name):
         return funarg, [], expression
 
     # Interior facet integral + mixed / vector element
-    name_ = name + "_"
-    shape = (2, fiat_element.space_dimension())
```
wence-: Maybe a short comment on what this packing is doing?  And why there are two options?

miklos1:
@wence-: I've added clarifications and a fix, please have a look.

wence-:
Yes, thanks.

Issue #17, opened 2016-02-15
numpy version


tkarna:
In `tsfc/fem.py` there are two calls to something like
```python
assert numpy.allclose(table, table.mean(axis=0, keepdims=True), equal_nan=True)
```
where the `equal_nan=True` argument requires numpy-1.10 or newer. As far as I know these are the only calls in the firedrake stack that do not work with numpy-1.9.1.

This may be an issue on machines where system wide numpy is ~1.9.1 and where `pip install numpy` doesn't work (TACC Stampede in my case, I was able to compile 1.10.4 manually in the end).

I wonder if there's a 1.9.1 compatible workaround to this? Also numpy version should probably be mentioned in the requirements: Now firedrake installation goes through with numpy-1.9.1 but results in a broken system.

miklos1:
I wouldn't be surprised if things would just work without the `equal_nan=True`, but I have to test it first.

miklos1:
So 1 hybridisation test fails with `equal_nan=False` on this assertion. First derivative of Trace(RT1) is estimated to be constant on the cell, but yields an all NaN table which is then probably not used after the simplifications are applied. I'm tagging #15 as related.

wence-:
How about including a backwards compat module:
```python
"""
Backwards compatibility for some functionality.
"""
import numpy
from distutils.version import StrictVersion


if StrictVersion(numpy.__version__) < StrictVersion("1.10"):
    def allclose(a, b, rtol=1e-5, atol=1e-8, equal_nan=False):
        """Wrapper around ``numpy.allclose``, which see.

        If ``equal_nan`` is ``True``, consider nan values to be equal
        in comparisons.
        """
        if not equal_nan:
            return numpy.allclose(a, b, rtol=rtol, atol=atol)
        nana = numpy.isnan(a)
        if not nana.any():
            return numpy.allclose(a, b, rtol=rtol, atol=atol)
        nanb = numpy.isnan(b)
        equal_nan = numpy.allclose(nana, nanb)
        return equal_nan and numpy.allclose(a[~nana], b[~nanb],
                                            rtol=rtol, atol=atol)
else:
     allclose = numpy.allclose
```

In any case, we should certainly bump the numpy required version.

dorugeber:
Depending on how dangerously you want to live, you could kill line 160 of
FIAT/trace_hdiv.py, and just tabulate zeroes for derivatives instead.
There's already a commented-out check at line 102-103.

[...]

PR #19, opened 2016-02-16
Compat

wence-:
Addresses #17, by providing a backwards compat function for `allclose` with equal_nan keyword argument.


miklos1:
Great.

PR #21, opened 2016-02-18
Lots of docstrings and minor cleanups

miklos1:
Two other notable changes:
- Some compiler optimisations are now configurable
- `gem.VariableIndex` wraps GEM expressions instead of a string

The configurable options are:
- `coffee_licm`: Trust COFFEE to do LICM. Disabled by default as COFFEE doesn't work on TSFC generated code yet. When enabled, it allows the inlining of expressions even if that pull calculations into inner loops.
- `unroll_indexsum`: Maximum extent to unroll index sums. Default is 3, so that loops over geometric dimensions are unrolled; this improves assembly performance. Can be disabled by setting it to `None`, `False` or `0`; that makes compilation time much shorter.

@wence-: can you have a look at the interpreter changes?

### DIFF
--- a/tsfc/constants.py
+++ b/tsfc/constants.py
@@ -13,7 +13,9 @@ SCALAR_TYPE = {numpy.dtype("double"): "double",
 
 PARAMETERS = {
     "quadrature_rule": "auto",
-    "quadrature_degree": "auto"
+    "quadrature_degree": "auto",
+    "coffee_licm": False,
+    "unroll_indexsum": 3,
 }
 
 
--- a/tsfc/constants.py
+++ b/tsfc/constants.py
@@ -14,7 +14,17 @@ SCALAR_TYPE = {numpy.dtype("double"): "double",
 PARAMETERS = {
     "quadrature_rule": "auto",
     "quadrature_degree": "auto",
+
+    # Trust COFFEE to do loop-invariant code motion. Disabled by
+    # default as COFFEE does not work on TSFC-generated code yet.
+    # When enabled, it allows the inlining of expressions even if that
+    # pulls calculations into inner loops.
     "coffee_licm": False,
+
+    # Maximum extent to unroll index sums. Default is 3, so that loops
+    # over geometric dimensions are unrolled; this improves assembly
+    # performance.  Can be disabled by setting it to None, False or 0;
+    # that makes compilation time much shorter.
     "unroll_indexsum": 3,
 }
 
--- a/tsfc/constants.py
+++ b/tsfc/constants.py
@@ -11,8 +11,6 @@ SCALAR_TYPE = {numpy.dtype("double"): "double",
                numpy.dtype("float32"): "float"}[NUMPY_TYPE]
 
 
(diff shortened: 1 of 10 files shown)


tsfc/constants.py
```diff
@@ -11,11 +11,11 @@
...
 
 
-RESTRICTION_MAP = {"+": 0, "-": 1}
-
 PARAMETERS = {
     "quadrature_rule": "auto",
-    "quadrature_degree": "auto"
+    "quadrature_degree": "auto",
+    "coffee_licm": False,
+    "unroll_indexsum": 3,
```
wence-: Documentation for `PARAMETERS` (it's sort of outside visible?)

tsfc/fem.py
```diff
@@ -522,11 +543,12 @@ def callback(key):
...
 
 def coordinate_coefficient(domain):
+    """Create a fake coordinate coefficient for a domain."""
     return ufl.Coefficient(ufl.FunctionSpace(domain, domain.ufl_coordinate_element()))
 
 
 def replace_coordinates(integrand, coordinate_coefficient):
-    # Replace SpatialCoordinate nodes with Coefficients
+    """Replace SpatialCoordinate nodes with Coefficients."""
     return map_expr_dag(ReplaceSpatialCoordinates(coordinate_coefficient), integrand)
```
wence-: I wonder if we should map `replace_coordinates` over all the integrands in an integral (rather than just singly) ?

tsfc/interpreter.py
```diff
@@ -238,11 +238,10 @@ def _(e, self):
             # Free index, want entire extent
             idx.append(Ellipsis)
         elif isinstance(i, gem.VariableIndex):
-            try:
-                # Variable indices must be provided in bindings
-                idx.append(self.bindings[i])
-            except KeyError:
-                raise ValueError("Binding for %s not found" % i)
+            # Variable index, evaluate inner expression
+            result, = evaluate(i.expression, self.bindings)
```
wence-: Use `self(i.expression)` ?

tsfc/constants.py
miklos1: I can copy here the info from the PR description.

tsfc/interpreter.py
miklos1: Good idea.

tsfc/fem.py
miklos1: Do we expect a significant amount of shared* expressions between integrands? I believe otherwise it shouldn't really matter.

*_Shared_ as in _object identity_, not as in _equal_.

tsfc/constants.py
wence-: Perfect.

tsfc/fem.py
wence-: Probably not.  If it turns out to be a problem we can fix it later.

miklos1:
@wence-: is it okay now?

wence-:
Yep, go go go.

PR #22, opened 2016-02-23
Interpolate UFL expressions

miklos1:
Refactoring in the driver for increased code sharing and loosening of an assertion.

### DIFF
--- a/tsfc/impero.py
+++ b/tsfc/impero.py
@@ -90,13 +90,13 @@ class Return(Terminal):
     __front__ = ('variable', 'expression')
 
     def __init__(self, variable, expression):
-        assert set(variable.free_indices) == set(expression.free_indices)
+        assert set(variable.free_indices) >= set(expression.free_indices)
 
         self.variable = variable
         self.expression = expression
 
     def loop_shape(self, free_indices):
-        return free_indices(self.expression)
+        return free_indices(self.variable)
 
 
 class ReturnAccumulate(Terminal):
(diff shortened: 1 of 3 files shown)


tsfc/impero.py
```diff
@@ -90,13 +90,13 @@ class Return(Terminal):
     __front__ = ('variable', 'expression')
 
     def __init__(self, variable, expression):
-        assert set(variable.free_indices) == set(expression.free_indices)
+        assert set(variable.free_indices) >= set(expression.free_indices)
```
wence-: OK, so now the lvalue of a return is allowed to have free indices that the rvalue does not.

tsfc/impero.py
```diff
@@ -90,13 +90,13 @@ class Return(Terminal):
...
 
         self.variable = variable
         self.expression = expression
 
     def loop_shape(self, free_indices):
-        return free_indices(self.expression)
+        return free_indices(self.variable)
```
wence-: And so the loop shape changes.

wence-:
This looks fine

tsfc/impero.py
miklos1: This comes up, for example, when we interpolate constants onto Functions.

wence-: Is this because we broadcast the constant, or because it's inlined, it doesn't have free indices?

miklos1: A constant is an expression that doesn't depend on the reference point. Yet we have to assign its to each point.

PR #23, opened 2016-02-23
Add COFFEE code generation for gem.MinValue and gem.MaxValue

miklos1:
William Booker will run into this problem as soon as he updates his Firedrake.


wence-:
tick.

PR #26, opened 2016-03-15
Fix code generation for gem.IndexSum

### LINKED ISSUE #25: Code generation merges loops incorrectly
Consider the following expression, where `A` is a vector:

    A[i] * IndexSum(A[i], i)

The generated C code looks like this:
```
double t0  = 0.0;
for (int i  = 0; i < 6; i += 1) {
    t0 += A[i];
    double t1  = A[i] * t0;
}
```
where `t1` is the "result". Obviously the value of `t0` incorrect in all but the last iteration.

If we instead start with the following, semantically equivalent GEM:

    A[i] * IndexSum(A[j], j)

then we get a correct code:
```
double t0  = 0.0;  
for (int j  = 0; j < 6; j += 1) {
  t0 += A[j];
}  
for (int i  = 0; i < 6; i += 1) {
  double t1  = (A[i] * t0);
}
```

miklos1:
Fixes #25, and protects the fix with a test case.

Changes:

1. A new Impero terminal `Noop` (no-op) is introduced, does nothing, but has a loop shape, thus it can prevent loop fusion. This terminal is thrown away during the construction of the Impero AST, i.e. when turning a list of Impero terminals into a tree.

2. An IndexSum over free index `r`, with free indices `(q,)`, formerly generated two Impero terminals:
   * `Initialise` with loop shape `(q,)`
   * `Accumulate` with loop shape `(q, r)`

 With this change, a third one is generated as well:
   * `Noop` with loop shape `(q,)`

 This is to guarantee that the accumulate loop is finished before the summed value is used.

### DIFF
--- a/tsfc/impero_utils.py
+++ b/tsfc/impero_utils.py
@@ -88,14 +88,10 @@ def collect_temporaries(ops):
         # IndexSum temporaries should be added either at Initialise or
         # at Accumulate.  The difference is only in ordering
         # (numbering).  We chose Accumulate here.
-        if isinstance(op, (imp.Initialise, imp.Return, imp.ReturnAccumulate)):
-            pass
-        elif isinstance(op, imp.Accumulate):
+        if isinstance(op, imp.Accumulate):
             result.append(op.indexsum)
         elif isinstance(op, imp.Evaluate):
             result.append(op.expression)
-        else:
-            raise AssertionError("unhandled operation: %s" % type(op))
     return result
 
 
@@ -117,6 +113,8 @@ def make_loop_tree(ops, get_indices, level=0):
             statements.append(imp.For(first_index[0], inner_block))
         else:
             statements.extend(op_group)
+    # Remove no-op terminals from the tree
+    statements = filter(lambda s: not isinstance(s, imp.Noop), statements)
     return imp.Block(statements)
 
 
@@ -130,7 +128,6 @@ def place_declarations(ops, tree, temporaries, get_indices):
     :arg get_indices: callable mapping from GEM nodes to an ordering
                       of free indices
     """
-
     temporaries_set = set(temporaries)
     assert len(temporaries_set) == len(temporaries)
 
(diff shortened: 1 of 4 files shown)


tsfc/impero_utils.py
```diff
@@ -88,14 +88,10 @@ def collect_temporaries(ops):
...
         # (numbering).  We chose Accumulate here.
-        if isinstance(op, (imp.Initialise, imp.Return, imp.ReturnAccumulate)):
-            pass
-        elif isinstance(op, imp.Accumulate):
+        if isinstance(op, imp.Accumulate):
             result.append(op.indexsum)
         elif isinstance(op, imp.Evaluate):
             result.append(op.expression)
-        else:
-            raise AssertionError("unhandled operation: %s" % type(op))
```
wence-: So previously you asserted (effectively) that you were passed on of `Initialise`, `Return`, `ReturnAccumulate`, `Accumulate` or `Evaluate`.  Now you silently pass through everything not explicitly handled.  I guess this is because now you have these `Noop` nodes too?

tsfc/impero_utils.py
```diff
@@ -206,10 +203,8 @@ def recurse_block(expr, loop_indices):
             e = op.expression
         elif isinstance(op, imp.Initialise):
             e = op.indexsum
-        elif isinstance(op, (imp.Accumulate, imp.Return, imp.ReturnAccumulate)):
-            continue
         else:
-            raise AssertionError("unhandled operation type %s" % type(op))
+            continue
```
wence-: The same here, I guess.

wence-:
Looks fine, I think.

tsfc/impero_utils.py
miklos1: I could have added `imp.Noop` to the first `if`, but then I decided to simplify the code by not listing all the accepted types, just the ones which are treated specially. This removes an assertion, but Impero nodes don't seem to change very often.

tsfc/impero_utils.py
miklos1: Yes, it's the same.

wence-:
OK, go.

Issue #1, opened 2016-02-04
Handle CellVolume


wence-:
On non-affine cells, we should compute `CellVolume` by integrating `detJ` over the reference cell (presumably using the same quadrature scheme as the rest of the form).  To do this, we need to emit a separate quadrature loop for the cell volume that occurs before the main quad-loop that uses it.

miklos1:
We just need to get a translated `|J|` somehow when we translate `CellVolume`, then it's just:

    IndexSum(|J|, quadrature_index)

if `|J|` depends on quadrature_index, otherwise:

    |J|

tkarna:
Any progress on this? Is FacetArea going to be supported as well? Many stabilization methods use the ratio CellVolume/FacetArea as a mesh length scale estimate.

miklos1:
It's good that you mention facets, because I was wondering whether to integrate `|J|` on the same quadrature points as the rest of the form, or just using the "correct" degree. However, when one uses `CellVolume` in a facet integral, then it cannot be sufficient to use the same quadrature points that we use to integrate on the facet. On a similar note, I believe `FacetArea` only makes sense in a facet integral, doesn't it?

tkarna:
> On a similar note, I believe FacetArea only makes sense in a facet integral, doesn't it?

I think so.

miklos1:
I'll look at this tomorrow.

miklos1:
I believe all we need to do is to compile the integration of 1 in the physical space down to GEM, and then during the UFL -> GEM translation phase of the actual form we just insert the previously generated GEM every time we encounter a `CellVolume` or `FacetArea` node.

I have to refactor the driver before this could actually be as easy as it sounds.

wence-:
Careful to make sure you integrate against the correct measure for `FacetArea`.

miklos1:
Yes.
 * `CellVolume` -> always `dx`
 * `FacetArea` -> same measure as for the form, only allowed with facet integrals

miklos1:
Branch `cellvolume` seems to support both `CellVolume` and `FacetArea`. I just need to clean up, test etc. and then it can go to master.

miklos1:
Working on this seems to uncover quite a few different problems in our stack:
 * [Degree estimation doesn't cater for non-affine cell geometry](https://bitbucket.org/fenics-project/ufl/issues/80/degree-estimation-doesnt-cater-for-non)
[...]

PR #27, opened 2016-03-18
Refactor kernel interface and UFL futzing

miklos1:
This is a refactoring step before we could have support for `CellVolume` and `FacetArea` (see #1).

Major changes:
 * Kernel interface is split off from the driver.
 * UFL routines are moved to `ufl_utils.py` (mostly from `fem.py`)
 * GEM -> ImperoC pipeline moves to `impero_utils.py` (basically from `driver.py`)

Requires firedrakeproject/firedrake#736.


tsfc/kernel_interface.py
```diff
@@ -0,0 +1,365 @@
...
+        self.ast = ast
+        self.integral_type = integral_type
+        self.oriented = oriented
+        self.subdomain_id = subdomain_id
+        self.coefficient_numbers = coefficient_numbers
+        super(Kernel, self).__init__()
+
+
+class Interface(object):
+    def __init__(self, integral_type, subdomain_id):
```
miklos1: Need some docstrings in this class.

tsfc/kernel_interface.py
```diff
@@ -0,0 +1,365 @@
...
+    def coefficient_mapper(self):
+        return lambda coefficient: self.coefficient_map[coefficient]
+
+    def require_cell_orientations(self):
+        self.kernel.oriented = True
+
+    def construct_kernel(self, name, body):
+        args = [self.local_tensor, self.coordinates_arg]
+        if self.kernel.oriented:
+            args.append(2, coffee.Decl("int *restrict *restrict",
```
wence-: Don't need the `2, ` I think.

miklos1: Correct.

tsfc/driver.py
```diff
@@ -183,24 +120,40 @@ def compile_integral(idata, fd, prefix, parameters):
...
+            break
+
+    impero_c = impero_utils.compile_gem([gem.partial_indexed(rv, argument_indices)
+                                         for rv in return_variables], ir,  # TODO
+                                        tuple(quadrature_indices) + argument_indices,
+                                        coffee_licm=parameters["coffee_licm"],
+                                        remove_zeros=True)
+
+    # Generate COFFEE
     index_names = zip(argument_indices, ['j', 'k'])
```
wence-: Could use `index_names = tuple(a.name for a in argument_indices)`

tsfc/fem.py
```diff
@@ -383,18 +145,17 @@ def __getitem__(self, key):
...
         integral_type = tabulation_manager.integral_type
         self.weights = gem.Literal(weights)
         self.quadrature_index = quadrature_index
-        self.argument_indices = argument_indices
         self.tabulation_manager = tabulation_manager
         self.integral_type = integral_type
-        self.coefficient_map = coefficient_map
+        self.coefficient_mapper = coefficient_mapper
         self.index_cache = index_cache
+        self.argument_indices = (gem.Index(name='j'), gem.Index(name='k'))
```
wence-: Is there a reason you *don't* pass argument_indices in any more and instead build new index instances here?

tsfc/impero_utils.py
```diff
@@ -27,6 +28,90 @@
...
+    apply_ordering = make_index_orderer(index_ordering)
+
+    get_indices = lambda expr: apply_ordering(expr.free_indices)
+
+    # Build operation ordering
+    ops = scheduling.emit_operations(zip(return_variables, expressions), get_indices)
+
+    # Empty kernel
+    if len(ops) == 0:
+        raise NoopError()
```
wence-: I don't really like exceptions for control flow, but I don't see a better way of doing this!

wence-:
Mostly looks fine (it's predominantly code movement).

tsfc/driver.py
miklos1: Done.

tsfc/fem.py
miklos1: I have changed my mind again. I think the current version is less convoluted than both the original and the first proposed.

miklos1:
I have revised this pull request, hitting it to most of the other unmerged branches. I'm now confident that this is a step in the right direction, please have a final review.

tsfc/kernel_interface.py
```diff
@@ -0,0 +1,467 @@
...
+        """Construct a COFFEE function declaration with the
+        accumulated glue code.
+
+        :arg name: function name
+        :arg args: function argument list
+        :arg body: function body (:class:`coffee.Block` node)
+        :returns: :class:`coffee.FunDecl` object
+        """
+        body.open_scope = False
+        body_ = coffee.Block(self.prepare + [body] + self.finalise)
```
wence-: Eww, use `coffee.Block(self.prepare + body.children + self.finalise)`?

miklos1: Now that I officially assume that `body` is a `coffee.Block`, I guess I can do just that. Maybe I can even add an assertion.

wence-:
OK.

miklos1:
Let's wait for [Travis](https://travis-ci.org/firedrakeproject/firedrake/builds/117716922), then I will merge.

PR #28, opened 2016-04-04
Activate COFFEE

miklos1:
This patch attaches the necessary pragmas onto the COFFEE AST such that COFFEE will understand what and how to optimise.

Depends on coneoproject/COFFEE#67.

### DIFF
--- a/tsfc/fem.py
+++ b/tsfc/fem.py
@@ -6,7 +6,6 @@ import itertools
 import numpy
 from singledispatch import singledispatch
 
-import ufl
 from ufl.corealg.map_dag import map_expr_dag, map_expr_dags
 from ufl.corealg.multifunction import MultiFunction
 from ufl.classes import (Argument, Coefficient, FormArgument,
@@ -235,11 +234,6 @@ def iterate_shape(mt, callback):
         return result[0]
 
 
-def zero_tracking_literal(table, ufl_element):
-    track_zeros = isinstance(ufl_element, (ufl.EnrichedElement, ufl.MixedElement))
-    return gem.Literal(table, track_zeros)
-
-
 @singledispatch
 def translate(terminal, mt, params):
     """Translates modified terminals into GEM.
@@ -270,9 +264,9 @@ def _(terminal, mt, params):
         table = params.tabulation_manager[key]
         if len(table.shape) == 1:
             # Cellwise constant
-            row = zero_tracking_literal(table, key[0])
+            row = gem.Literal(table)
         else:
-            table = params.select_facet(zero_tracking_literal(table, key[0]), mt.restriction)
+            table = params.select_facet(gem.Literal(table), mt.restriction)
             row = gem.partial_indexed(table, (params.quadrature_index,))
         return gem.Indexed(row, (argument_index,))
 
@@ -293,7 +287,7 @@ def _(terminal, mt, params):
         table = params.tabulation_manager[key]
         if len(table.shape) == 1:
             # Cellwise constant
-            row = zero_tracking_literal(table, key[0])
+            row = gem.Literal(table)
             if numpy.count_nonzero(table) <= 2:
                 assert row.shape == ka.shape
                 return reduce(gem.Sum,
@@ -301,7 +295,7 @@ def _(terminal, mt, params):
                                for i in range(row.shape[0])],
                               gem.Zero())
         else:
-            table = params.select_facet(zero_tracking_literal(table, key[0]), mt.restriction)
+            table = params.select_facet(gem.Literal(table), mt.restriction)
             row = gem.partial_indexed(table, (params.quadrature_index,))
 
         r = params.index_cache[terminal.ufl_element()]
--- a/tsfc/fem.py
+++ b/tsfc/fem.py
@@ -6,6 +6,7 @@ import itertools
 import numpy
 from singledispatch import singledispatch
 
+import ufl
 from ufl.corealg.map_dag import map_expr_dag, map_expr_dags
 from ufl.corealg.multifunction import MultiFunction
 from ufl.classes import (Argument, Coefficient, FormArgument,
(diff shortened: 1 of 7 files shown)


tsfc/gem.py
```diff
@@ -97,10 +97,11 @@ def value(self):
 class Literal(Terminal):
     """Tensor-valued constant"""
 
-    __slots__ = ('array',)
+    __slots__ = ('array', 'track_zeros')
```
miklos1: I don't like this at all.

tsfc/fem.py
```diff
@@ -233,6 +234,11 @@ def flat_index(ordered_deriv):
         return result[0]
 
 
+def zero_tracking_literal(table, ufl_element):
+    track_zeros = isinstance(ufl_element, (ufl.EnrichedElement, ufl.MixedElement))
```
miklos1: So I did this to exactly mimic the old FFC behaviour.
@FabioLuporini: Is there any reason not to track zeros for all matrices? Code would be much simpler and with better separation of concerns.

wence-: Unless there are strong reasons not to, I too would be in favour of always tracking zeros.

FabioLuporini: There is not, you're correct. The only thing I don't want is to track zero-valued columns due to taking derivatives

dorugeber: ......why not?

FabioLuporini: impact on low-level performance (roughly speaking, playing with loop indices is very very bad if you are "just" skipping very few ops on zeros)

wence-: Shurely that is a choice that COFFEE should be making based on the nonzero bounds it is fed.  Rather than trying to pre-guess which tables would require zero tracking or not.

dorugeber: Maybe I missed something.  I thought Fabio only ever stripped away initial and final columns.  I didn't think there was any risk of looping over, e.g. columns 1-4, 6, 8-10, 12-16 [if cols 5, 7 and 11 'happen' to be 0].

FabioLuporini: Andrew: No, coffee can skip blocks of zeros irrespective of where they are located
Lawrence: COFFEE could do that in theory, yeah...

wence-: OK, so the argument in favour of removing this special casing is that tsfc doesn't know about what constitutes a "small loop" and therefore should not be eliminated.  So it should always annotate zero columns.  Rather than using the heuristic that "if it's a mixed or enriched element then there will probably be *enough* zero columns".

The argument in favour of keeping it is that coffee currently doesn't check how many zero columns it's eliminating, and always does so?

FabioLuporini: Correct. As I said, the idea, perhaps naive at that time I don't know, was to exploit the "knowledge" that in non-scalar function spaces we have "non-degenerate" blocks of zero-valued entries (non-degenerate ---> not blocks of size Nx1, i.e. spare columns)

dorugeber: @wence- I think it's more like "mixed [and vector] elements are constructed in a way that *guarantees* a block of contiguous zero columns".  Then Enriched is there so you catch the HDiv(....) + HDiv(....) elements.  I guess Enriched just passed up information from the subelements rather than doing any inspection of its own.

FabioLuporini: @dorugeber to clarify my comment about the impact on performance: basically, if you have a loop, it is always better that accesses are adjacent in memory (spatial locality), e.g. A[i], A[i+1], A[i+2] etc. If you have jumps you may well screw a few things in the fetch-decode-execute pipeline (including vectorisation).  Essentially, if we sacrifice this property, it has to be for a good cause; for example, saving on a significant amount of operations on zeros. A single column (unless a very, very big one) is not a significant amount. Does this help ?

dorugeber: @FabioLuporini Yes, I understand that.  But I think your mention of derivatives before was *misleading*, perhaps?  You're not *not interested* in derivatives producing zero columns... if derivatives made the first 8 columns zero then you'd like to exploit that, I'm sure?

The thing you're not interested in is knocking out single columns, for the reason you posted, therefore you don't do exhaustive table-sniffing.  Right?

FabioLuporini: You're correct

wence-: So as I understand the code, the zero tracking that is added here only squeezes zeros out of the start or end of the table. e.g. if I provide an array that is `[1, 0, 0, 0, 0, 0, 0, 1]` it won't notice the zero cols in the middle.  So your only worry is if you had something like `[0, 1, 1, 1]`, you don't want squashing to `[1, 1, 1]`, right?

FabioLuporini: Correct.

One more comment: COFFEE could handle zeros in the middle too, it's just that the performance experiments I run time ago showed that vectorisation was much worse by creating jumps "right in the middle" on the iteration space. I actually had forgotten about this until I looked again at the code

wence-: OK, so we could always track zeros and just put in the whole loop bounds if the number of zero columns is less than what?

dorugeber: Depends on location and structure, surely

FabioLuporini: Clearly we should leave this as a parameter. But heuristically I'd say not to track them if <= 2 (I can run some quick tests later on for confirmation)

wence-: Is it trivial to make this change in coffee itself?

FabioLuporini: yeah probably you're right and coffee is the right place... should be relatively simple. Keep you posted

miklos1: Maybe we can drop fca01bc completely then?

miklos1:
@FabioLuporini: Simplified the `SparseArrayInit` stuff on this side.

FabioLuporini:
I've done my job to. Tests are running as I write

miklos1:
I think this is now ready to go (once the corresponding COFFEE branch is merged).

FabioLuporini:
I believe I have to cook up something to warn the users we are about to re-integrate coffee into the system. We need to think carefully about when is the best time to land it. I'll be away 4 days (Tue-Fri) next week and won't be able to work on potential issues (hopefully none!)

Btw, we could also try, as I was saying yesterday, removing crap Par objects entirely in COFFEE.

miklos1:
I just tried that (removing `Par`) nodes, and it seems to work. I will push very soon.

PR #29, opened 2016-04-08
Spin off GEM as a separate Python package

miklos1:
We need this to break a reference cycle between TSFC and FInAT. By spinning GEM off:
 * GEM becomes standalone Python package.
 * FInAT will depend on GEM.
 * TSFC depends on GEM, and will depend on FInAT.

For now GEM remains in the TSFC repository. We can change this later if it proves to be unsatisfactory.

Must be merged at the same time as firedrakeproject/firedrake#749.


wence-:
Go.

PR #30, opened 2016-04-09
Delta in gem

dham:


miklos1:
I refreshed the branch, but I cannot change the merge target, so closing this to open another one.

PR #31, opened 2016-04-11
Abstract base classes IndexBase and Constant

miklos1:
Currently GEM has three types of indices:
 * `int`: fixed index
 * `VariableIndex`: indices which are constants for a single execution of the kernel, but not compile-time constant, for example: facet number.
 * `Index`: running free index, will generate a loop.

`IndexBase` is introduced by this pull request as an abstract base class for these index types.

GEM will have three types of compile-time constant nodes:
 * `Literal`: tensor of floating-point values with arbitrary shape (general case: the data is a numpy array)
 * `Zero`: special case of `Literal`, all elements are zero, thus no need to store array of all zeros
 * `Identity`: identity matrix (also a special case `Literal`, but no need store the matrix)

`Identity` is introduced by this pull request, to map Identity from UFL, and for future use with Delta.
This pull request also introduces the abstract base class `Constant`, which is the base class of `Literal`, `Identity` and `Zero`.

The reason I typed also these out, it that if some of the names are misleading, better suggestions are welcome.

### DIFF
--- a/gem/gem.py
+++ b/gem/gem.py
@@ -16,6 +16,7 @@ indices.
 
 from __future__ import absolute_import
 
+from abc import ABCMeta
 from itertools import chain
 from numpy import asarray, eye, unique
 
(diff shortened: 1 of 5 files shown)


wence-:
fine

PR #32, opened 2016-04-11
Nascent Delta support for FInAT

miklos1:
@rckirby, @dham: This is against the `finat` branch.

This is on top of #31 (subset that can go to master right now), the only new commit is 150e6a5.


dham:
Looks fine to me.

PR #33, opened 2016-04-20
Building ListTensor from non-scalars

miklos1:
In GEM, ListTensor is built from scalar components (although the tensor itself can have arbitrary rank). This is not the case in UFL, where a ListTensor can be built from tensor components. Thus we flatten this composition when translating UFL to GEM. This pull request moves this flattening to the constructor of ListTensor, so that people (e.g. @dham) writing GEM can express more structure in their code (even though that structure is immediately thrown away for now).

In addition this pull request improves constant propagation: a ListTensor built from solely from constants will construct a Literal instead. Previously this only worked for Zeros.

### DIFF
--- a/gem/gem.py
+++ b/gem/gem.py
@@ -18,7 +18,9 @@ from __future__ import absolute_import
 
 from abc import ABCMeta
 from itertools import chain
-from numpy import asarray, eye, unique
+import numpy
+from numpy import asarray, unique
+from operator import attrgetter
 
 from gem.node import Node as NodeBase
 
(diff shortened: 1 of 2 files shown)


gem/gem.py
```diff
@@ -492,11 +494,23 @@ class ListTensor(Node):
 
     def __new__(cls, array):
         array = asarray(array)
-
-        # Zero folding
-        if all(isinstance(elem, Zero) for elem in array.flat):
-            assert all(elem.shape == () for elem in array.flat)
-            return Zero(array.shape)
+        assert numpy.prod(array.shape)
```
wence-: This disallows, for example shape `(6, 0)`.  I suppose that is deliberate.

miklos1: Yes. I don't think we need those shapes.

wence-:
This looked fine,  I think.

PR #34, opened 2016-04-26
AffineIndex and IndexIterator

dham:
Implement affine index groups and an iterator for multi-indices over sets of Index and AffineIndex.


gem/gem.py
```diff
@@ -566,6 +582,55 @@ def __new__(cls, i, j):
...
+        for group, pos in self.affine_groups:
+            if sum(multiindex[p] for p in pos) >= group[0].extent:
+                return False
+        return True
+        
+    def __iter__(self):
+        # Fix this for affine index groups.
+        while True:
+            multiindex = self.ndindex.next()
+            if self._affine_groups_legal(multiindex):
```
miklos1: This isn't the best way to do it, but since it doesn't go to master yet, we can fix it up later.

PR #36, opened 2016-05-03
CellCoordinate and FacetCoordinate support

miklos1:
Refactors some of the facet management utilities into the `FacetManager` class. Implements UFL geometric nodes `CellCoordinate` and `FacetCoordinate`.


wence-:
Fine.

PR #37, opened 2016-05-10
Split mixed coefficients

miklos1:
This will become useful to support mixed problems with FInAT, but I think the change is solid enough to go directly to master.

After this change, `tsfc.MixedElement` (formerly known as `ffc.MixedElement`) will only be used to implement `ufl.VectorElement`s and `ufl.TensorElement`, because actual mixed coefficients are split.

This changes the kernel contract between TSFC and "PyOP2", so it must be merged at the same time as firedrakeproject/firedrake#775.


wence-:
I think this looks good.

wence-:
This is also step 1 on the path to removing the transposition of data between the in memory firedrake layout and what the kernel expects.

PR #38, opened 2016-05-11
Fix COFFEE Symbol creation

miklos1:
Required for a new Thetis test case.


tkarna:
This seems to fix the issue.

For the record the issue was that Thetis test
```
py.test test_steady_state_basin_mms.py::test_steady_state_basin_convergence[Setup9-P1DGP2]
```
failed with a compilation error (undeclared variable `i` on line 204).

FabioLuporini:
Looks sane to me

PR #39, opened 2016-05-24
Fix symmetric TensorElements

miklos1:
Reference value already has the symmetry subtracted.

### DIFF
--- a/tsfc/fem.py
+++ b/tsfc/fem.py
@@ -44,7 +44,7 @@ def _tabulate(ufl_element, order, points):
     """
     element = create_element(ufl_element)
     phi = element.space_dimension()
-    C = ufl_element.reference_value_size() - len(ufl_element.symmetry())
+    C = ufl_element.reference_value_size()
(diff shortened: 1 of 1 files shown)


tsfc/fem.py
```diff
@@ -44,7 +44,7 @@ def _tabulate(ufl_element, order, points):
     """
     element = create_element(ufl_element)
     phi = element.space_dimension()
-    C = ufl_element.reference_value_size() - len(ufl_element.symmetry())
+    C = ufl_element.reference_value_size()
```
miklos1: From `ufl.finiteelement.mixedelement.TensorElement.__init__`:
```python
        # Compute reference value shape based on symmetries
        if symmetry:
            # Flatten and subtract symmetries
            reference_value_shape = (product(shape)-len(symmetry),)
            mapping = "symmetries"
        else:
            # Do not flatten if there are no symmetries
            reference_value_shape = shape
            mapping = "identity"
```

PR #42, opened 2016-05-29
fem: Handle ConstantValue in modified terminal translation

### LINKED ISSUE #41: Compiling interior facet integrals containing grads and normals fails for vector function spaces in mixed function spaces
See:
https://gist.github.com/colinjcotter/0011563c10e8dd7065f42c21ad2cb2dd

wence-:
Could happen if zero-simplification in UFL didn't pick up that we had
grad of a cellwise constant sub coefficient in a mixed form.  Fixes #41.

### DIFF
--- a/tsfc/fem.py
+++ b/tsfc/fem.py
@@ -8,8 +8,9 @@ from singledispatch import singledispatch
 
 from ufl.corealg.map_dag import map_expr_dag, map_expr_dags
 from ufl.corealg.multifunction import MultiFunction
-from ufl.classes import (Argument, Coefficient, FormArgument,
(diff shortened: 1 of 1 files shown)


tsfc/fem.py
```diff
@@ -328,6 +329,11 @@ def callback(key):
     return iterate_shape(mt, callback)
 
 
+@translate.register(Zero)
+def _translate_zero(terminal, mt, params):
```
miklos1: I will make this a bit more general, so that not just `Zero`s, but all constant values are handled.

miklos1:
I have updated this PR a bit. I think it can be merged once it is tested against the Firedrake test suite.

PR #44, opened 2016-06-06
CellVolume and FacetArea

miklos1:
Implements `CellVolume` and `FacetArea` for all supported cell types, including non-affine. Higher-order cells are still under-integrated, but that will be fixed by [upstream UFL](https://bitbucket.org/fenics-project/ufl/issues/80/degree-estimation-doesnt-cater-for-non).

Must be merged at the same time with firedrakeproject/firedrake#800.

### DIFF
--- a/tests/test_create_quadrature.py
+++ b/tests/test_create_quadrature.py
@@ -37,24 +37,6 @@ def test_select_degree(cell, degree, itype):
 
 
 @pytest.mark.parametrize("degree",
-                         [(1, 1), (2, 2)])
-@pytest.mark.parametrize("itype",
-                         ["interior_facet", "exterior_facet"])
-def test_select_degree_facet_quad(degree, itype):
-    selected = q.select_degree(degree, ufl.quadrilateral, itype)
-    assert selected == degree[0]
-
-
-@pytest.mark.parametrize("degree",
-                         [(1, 2), (2,)])
-@pytest.mark.parametrize("itype",
-                         ["interior_facet", "exterior_facet"])
-def test_select_invalid_degree_facet_quad(degree, itype):
-    with pytest.raises(ValueError):
-        q.select_degree(degree, ufl.quadrilateral, itype)
-
-
-@pytest.mark.parametrize("degree",
                          [(1, 2), (2, 3)])
 @pytest.mark.parametrize("itype",
                          ["interior_facet_horiz", "exterior_facet_top",
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -117,7 +117,6 @@ def compile_integral(integral_data, form_data, prefix, parameters):
         # Check if the integral has a quad degree attached, otherwise use
         # the estimated polynomial degree attached by compute_form_data
         quadrature_degree = integral.metadata()["estimated_polynomial_degree"]
-        quad_rule = create_quadrature(cell, 'cell', quadrature_degree)
 
         integrand = ufl_utils.replace_coordinates(integral.integrand(), coordinates)
         quadrature_index = gem.Index(name='q')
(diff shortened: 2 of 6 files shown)


tests/test_create_quadrature.py
```diff
@@ -37,24 +37,6 @@ def test_select_degree(cell, degree, itype):
 
 
 @pytest.mark.parametrize("degree",
-                         [(1, 1), (2, 2)])
-@pytest.mark.parametrize("itype",
-                         ["interior_facet", "exterior_facet"])
-def test_select_degree_facet_quad(degree, itype):
```
miklos1: Firedrake never does this.

tsfc/driver.py
```diff
@@ -42,6 +42,7 @@ def compile_form(form, prefix="form", parameters=None):
                            do_apply_integral_scaling=True,
                            do_apply_geometry_lowering=True,
                            do_apply_restrictions=True,
+                           preserve_geometry_types=(CellVolume, FacetArea),
```
miklos1: On affine cells, UFL lowers `CellVolume` and `FacetArea` to reference quantities by default, while leaving `CellVolume` and `FacetArea` nodes intact on non-affine cells. This line leaves the handling of these nodes for TSFC even for affine cells. The TSFC pipeline seems to be more robust.

tsfc/driver.py
```diff
@@ -102,6 +103,76 @@ def compile_integral(integral_data, form_data, prefix, parameters):
     # evaluation can be hoisted).
     index_cache = collections.defaultdict(gem.Index)
 
+    def cellvolume(restriction):
```
miklos1: I will still refactor these.

tsfc/fem.py
```diff
@@ -191,35 +192,90 @@ def select_facet(self, tensor, restriction):
             return gem.partial_indexed(tensor, (f,))
 
 
+class cached_property(object):
```
miklos1: Copy-pasta from PyOP2, thinking about a better approach.

tsfc/fem.py
```diff
@@ -191,35 +192,90 @@ def select_facet(self, tensor, restriction):
...
-        self.select_facet = facet_manager.select_facet
-
-        if self.integral_type.startswith("interior_facet"):
+
+        if parameters.integral_type.startswith("interior_facet"):
             self.cell_orientations = gem.Variable("cell_orientations", (2, 1))
         else:
             self.cell_orientations = gem.Variable("cell_orientations", (1, 1))
 
+        parameters.__dict__.update(self.__dict__)
```
miklos1: I probably don't need this any more.

tsfc/geometric.py
```diff
@@ -155,6 +163,11 @@ def translate_reference_cell_volume(terminal, mt, params):
     return gem.Literal(reference_cell_volume[reference_cell(terminal)])
 
 
+@translate.register(ReferenceFacetVolume)
+def translate_reference_facet_volume(terminal, mt, params):
```
miklos1: This (and the previous two snippets) is only used if there is no `FacetArea` in `preserve_geometry_types=` when calling `compute_form_data`.

tsfc/quadrature.py
```diff
@@ -391,19 +391,14 @@ def select_degree(degree, cell, integral_type):
             raise ValueError("Integral type '%s' invalid for cell '%s'" %
                              (integral_type, cell.cellname()))
         if cell.cellname() == "quadrilateral":
-            try:
-                d1, d2 = degree
```
miklos1: Never happens in Firedrake.

wence-:
Note as per discussion in #1 underintegration may be a reasonable thing to do.  This is a "there is no right general answer" case, so we should decide on what we want and do that I think.

tsfc/quadrature.py
```diff
@@ -391,19 +391,14 @@ def select_degree(degree, cell, integral_type):
...
-                    raise ValueError("tuple degree must have matching values")
-                return d1
-            except TypeError:
-                return degree
+            assert isinstance(degree, int)
         return degree
     if not isinstance(cell, ufl.TensorProductCell):
         raise ValueError("Integral type '%s' invalid for cell '%s'" %
                          (integral_type, cell.cellname()))
+    # Fix degree on TensorProductCell when not tuple
```
miklos1: Workaround for firedrakeproject/firedrake#744.

tsfc/driver.py
wence-: It looks like the best thing to do is to refactor in a separate round.

wence-:
I think go ahead.

miklos1:
I'll just add some `TODO`s or `FIXME`s and then merge.

PR #45, opened 2016-06-13
Support for Python logging

miklos1:
Instead of printing messages to standard output, use Python logging, so that these messages can be disabled.

Current configuration leaves the default behaviour unchanged. Maybe we don't want that?

### DIFF
--- a/tsfc/logging.py
+++ b/tsfc/logging.py
@@ -1,0 +1,8 @@
+"""Logging for TSFC."""
+
+from __future__ import absolute_import
+
+import logging
+
+logger = logging.getLogger('tsfc')
+logger.addHandler(logging.StreamHandler())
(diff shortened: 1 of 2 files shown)


tsfc/logging.py
```diff
@@ -0,0 +1,11 @@
+"""Logging for TSFC."""
+
+from __future__ import absolute_import
+
+import logging
+
+logger = logging.getLogger('tsfc')
+logger.setLevel(logging.INFO)
```
miklos1: This is to keep the current behaviour.

tsfc/logging.py
```diff
@@ -0,0 +1,11 @@
...
+
+default_handler = logging.StreamHandler()
```
miklos1: A warning is printed if no handler is registered.

Firedrake will probably want to remove this handler, and register an "`MPISteamHandler`". However, I also don't TSFC to rely on Firedrake doing some stuff for it.

wence-: Yes, this can be done like so:
```python
from firedrake import *

import logging

logger = logging.getLogger("tsfc")

if op2.MPI.comm.rank == 0:
    ch = logging.StreamHandler()

    ch.setFormatter(logging.Formatter("Banananas %(message)s"))
else:
    ch = logging.NullHandler()

for hdlr in logger.handlers:
    logger.removeHandler(hdlr)
logger.addHandler(ch)
```

PR #46, opened 2016-07-13
Tsfc trace

thomasgibson:
This branch depends on the following other branches: firedrakeproject/firedrake#824, firedrakeproject/ufl#3, and firedrakeproject/fiat#6.

Summary of changes:
1. A gem failure node called `Failure` is created to handle any exceptions that arise during the form compiling process, such as illegal tabulations on trace elements.
2. The `facetmanager` class in `fem.py` has a generator function called `entity_ids` that creates the appropriate entity arguments for the `TabulationManager`.
3. Changes were made in `fiatinterface.py` to address a problem when tabulating a tensor product element defined on a quad. Specifically with tabulating an HDiv element defined on a quadrilateral mesh.

### DIFF
--- a/tsfc/fem.py
+++ b/tsfc/fem.py
@@ -25,10 +25,9 @@ from tsfc.parameters import PARAMETERS
 from tsfc.ufl_utils import (CollectModifiedTerminals,
                             ModifiedTerminalMixin, PickRestriction,
                             spanning_degree, simplify_abs)
-from FIAT.hdiv_trace import TraceError
 
 
-def _tabulate(ufl_element, order, points, entity):
+def tabulate(ufl_element, order, points, entity, epsilon):
     """Ask FIAT to tabulate ``points`` up to order ``order``, then
     rearranges the result into a series of ``(c, D, table)`` tuples,
     where:
@@ -47,41 +46,28 @@ def _tabulate(ufl_element, order, points, entity):
     phi = element.space_dimension()
     C = ufl_element.reference_value_size()
     q = len(points)
-    try:
-        for D, fiat_table in iteritems(element.tabulate(order, points, entity)):
+    for D, fiat_table in iteritems(element.tabulate(order, points, entity)):
+        if isinstance(fiat_table, Exception):
+            # In the case an exception is found in the fiat table, do not
+            # perform any rounding
+            gem_fail = gem.Failure((q, phi), fiat_table)
+            for c in range(C):
+                yield c, D, gem_fail
+        else:
             reordered_table = fiat_table.reshape(phi, C, q).transpose(1, 2, 0)  # (C, q, phi)
             for c, table in enumerate(reordered_table):
-                yield c, D, table
-    except TraceError as TE:
-        # import sys
-        # shape = (q, phi)
-        # TODO: Incorporate a proper Failure Table into the tabulation manager.
-        # Currently we are returned appropriately sized zeros arrays for gradient
-        # tabulations on trace element.
-        # gemTable = gem.Failure(shape, sys.exc_info())
-        for D, fiat_table in TE.zeros.iteritems():
-            reordered_table = fiat_table.reshape(phi, C, q).transpose(1, 2, 0)  # (C, q, phi)
-            for c, table in enumerate(reordered_table):
-                yield c, D, table
+                # Copied from FFC (ffc/quadrature/quadratureutils.py)
+                table[abs(table) < epsilon] = 0
+                table[abs(table - 1.0) < epsilon] = 1.0
+                table[abs(table + 1.0) < epsilon] = -1.0
+                table[abs(table - 0.5) < epsilon] = 0.5
+                table[abs(table + 0.5) < epsilon] = -0.5
 
+                if spanning_degree(ufl_element) <= sum(D) and ufl_element.family() != "HDiv Trace":
+                    assert compat.allclose(table, table.mean(axis=0, keepdims=True), equal_nan=True)
+                    table = table[0]
 
-def tabulate(ufl_element, order, points, entity, epsilon):
-    """Same as the above, but also applies FFC rounding and recognises
-    cellwise constantness.  Cellwise constantness is determined
-    symbolically, but we also check the numerics to be safe."""
-    for c, D, table in _tabulate(ufl_element, order, points, entity):
-        # Copied from FFC (ffc/quadrature/quadratureutils.py)
-        table[abs(table) < epsilon] = 0
-        table[abs(table - 1.0) < epsilon] = 1.0
-        table[abs(table + 1.0) < epsilon] = -1.0
-        table[abs(table - 0.5) < epsilon] = 0.5
-        table[abs(table + 0.5) < epsilon] = -0.5
-
-        if spanning_degree(ufl_element) <= sum(D) and ufl_element.family() != "HDiv Trace":
-            assert compat.allclose(table, table.mean(axis=0, keepdims=True), equal_nan=True)
-            table = table[0]
-
-        yield c, D, table
+                yield c, D, gem.Literal(table)
 
 
 def make_tabulator(points, entity):
@@ -120,11 +106,11 @@ class TabulationManager(object):
                 store[(ufl_element, c, D)].append(table)
 
         for key, tables in iteritems(store):
-            table = numpy.array(tables)
+            table = gem.ListTensor(tables)
             if len(table.shape) == 2:
                 # Cellwise constant; must not depend on the facet
-                assert compat.allclose(table, table.mean(axis=0, keepdims=True), equal_nan=True)
-                table = table[0]
+                assert compat.allclose(table.array, table.array.mean(axis=0, keepdims=True), equal_nan=True)
+                table = gem.Literal(table.array[0])
             self.tables[key] = table
 
     def __getitem__(self, key):
@@ -332,9 +318,10 @@ def translate_argument(terminal, mt, ctx):
         table = ctx.tabulation_manager[key]
         if len(table.shape) == 1:
             # Cellwise constant
-            row = gem.Literal(table)
+            row = table
         else:
-            table = ctx.index_selector(lambda i: gem.Literal(table[i]), mt.restriction)
+            table = ctx.index_selector(lambda i: gem.partial_indexed(table, (i,)),
+                                       mt.restriction)
             row = gem.partial_indexed(table, (ctx.point_index,))
         return gem.Indexed(row, (argument_index,))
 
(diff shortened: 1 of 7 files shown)


tsfc/fiatinterface.py
```diff
@@ -58,7 +58,8 @@
     "Nedelec 2nd kind H(curl)": FIAT.NedelecSecondKind,
     "TensorProductElement": FIAT.TensorProductElement,
     "Raviart-Thomas": FIAT.RaviartThomas,
-    "TraceElement": FIAT.HDivTrace,
+    "TraceElement": FIAT.TraceHDiv,
```
miklos1: I assume you have this for compatibility with old code. It would probably be better to register that alias in UFL.

tsfc/mixedelement.py
```diff
@@ -95,7 +96,10 @@ def tabulate(self, order, points):
         crange = numpy.cumsum(sub_cmps)
 
         for i, e in enumerate(self.elements()):
-            table = e.tabulate(order, points)
+            try:
+                table = e.tabulate(order, points, entity)
+            except TraceError as TE:
+                table = TE.zeros
```
miklos1: Are you swallowing a potential error here? Why?

thomasgibson: In the case that a gradient of a trace element is tabulated, the exception `TE` will return the order=0 function evaluations as well as appropriately sized zeros in the dictionary for all orders of derivatives. That's what `TE.zeros` returns - it includes zeros for all gradient evaluations.

miklos1:
We should review together the situation with the `TraceError` exception, but first I would like you to fix the quad patching.

thomasgibson:
Yes I agree we should. I'll have a look at this more closely

tsfc/fiatinterface.py
```diff
@@ -122,19 +123,30 @@ def _(element, vector_is_mixed):
...
-        flat_entity_ids[0] = entity_ids[(0, 0)]
-        flat_entity_ids[1] = dict(enumerate(entity_ids[(0, 1)].values() +
-                                            entity_ids[(1, 0)].values()))
-        flat_entity_ids[2] = entity_ids[(1, 1)]
-
-        element.dual = DualSet(nodes, ref_el, flat_entity_ids)
-        element.ref_el = ref_el
-        return element
+
+        def cell_to_quad(element):
```
thomasgibson: I am also somewhat convinced that this is causing the problems I'm experiencing with the caching errors described in Firedrake issue #823. This was added to address some problems I experienced with `RTCF` elements on a `FiredrakeQuadrilateral`.

I managed to work out exactly which particular `pytest` parameters are having the caching issue: they all are tests related to `RTCF` elements. I think I need to discuss this in more detail with you to create a fix for this. Perhaps fixing this might solve my problem in Firedrake issue #823.

thomasgibson:
Quadrilateral patching is done. This should fix firedrakeproject/firedrake#823

tsfc/fiatinterface.py
```diff
@@ -121,32 +198,8 @@ def _(element, vector_is_mixed):
         # Can't use create_element here because we're going to modify
```
miklos1: This is no longer true, I guess.

wence-: Yes, that's right.  We should use `create_element` now.

tsfc/fem.py
```diff
@@ -153,32 +168,53 @@ def __init__(self, integral_type, ufl_cell):
...
         elif self.integral_type in ['exterior_facet', 'interior_facet']:
             for entity in range(self.ufl_cell.num_facets()):
-                t = as_fiat_cell(self.ufl_cell).get_entity_transform(dim-1, entity)
+                t = as_fiat_cell(self.ufl_cell).get_facet_transform(entity)
                 yield numpy.asarray(map(t, points))
 
         elif self.integral_type in ['exterior_facet_bottom', 'exterior_facet_top', 'interior_facet_horiz']:
             for entity in range(2):  # top and bottom
-                t = as_fiat_cell(self.ufl_cell).get_entity_transform((dim-1, 0), entity)
+                t = as_fiat_cell(self.ufl_cell).get_horiz_facet_transform(entity)
```
miklos1: I don't think this function exists anymore.

tsfc/mixedelement.py
miklos1: I doubt this is still relevant after #37.

tsfc/fiatinterface.py
miklos1: @thomasgibson: This is still outstanding.

tsfc/fiatinterface.py
```diff
@@ -52,14 +52,13 @@
     "Discontinuous Lagrange": FIAT.DiscontinuousLagrange,
     "Discontinuous Taylor": FIAT.DiscontinuousTaylor,
     "Discontinuous Raviart-Thomas": FIAT.DiscontinuousRaviartThomas,
-    "Discontinuous Lagrange Trace": FIAT.DiscontinuousLagrangeTrace,
```
miklos1: Let's keep this if we kept the FIAT element.

tsfc/fem.py
```diff
@@ -319,9 +317,10 @@ def callback(key):
         table = ctx.tabulation_manager[key]
         if len(table.shape) == 1:
             # Cellwise constant
-            row = gem.Literal(table)
+            row = table
         else:
-            table = ctx.index_selector(lambda i: gem.Literal(table[i]), mt.restriction)
+            table = ctx.index_selector(lambda i: gem.partial_indexed(table, (i,)),
```
miklos1: This explains the rank change we have seen in the generated code.

tsfc/fem.py
```diff
@@ -340,15 +339,16 @@ def callback(key):
...
+            row = table
+            if numpy.count_nonzero(table.array) <= 2:
                 assert row.shape == vec.shape
                 return reduce(gem.Sum,
                               [gem.Product(gem.Indexed(row, (i,)), gem.Indexed(vec, (i,)))
                                for i in range(row.shape[0])],
                               gem.Zero())
         else:
-            table = ctx.index_selector(lambda i: gem.Literal(table[i]), mt.restriction)
+            table = ctx.index_selector(lambda i: gem.partial_indexed(table, (i,)),
```
miklos1: Same here...

tsfc/fem.py
```diff
@@ -14,7 +14,6 @@
 from ufl.classes import (Argument, Coefficient, CellVolume,
                          FacetArea, FormArgument,
                          GeometricQuantity, QuadratureWeight)
-
```
miklos1: Don't remove this blank here.

tsfc/fem.py
thomasgibson: So this bit (as well as the comment above) is an issue, yes? I'm not sure what the rank-change issue is.

miklos1: To some extent this is just code generation cosmetics. But `gem.Literal(table.array[i])` would produce nicer code I think than `gem.partial_indexed(table, (i,))`.

tsfc/fem.py
wence-: I'd prefer the former code, I think.

tsfc/fem.py
```diff
@@ -41,57 +40,56 @@ def _tabulate(ufl_element, order, points):
...
-    also check the numerics to be safe."""
-    for c, D, table in _tabulate(ufl_element, order, points):
-        # Copied from FFC (ffc/quadrature/quadratureutils.py)
-        table[abs(table) < epsilon] = 0
-        table[abs(table - 1.0) < epsilon] = 1.0
-        table[abs(table + 1.0) < epsilon] = -1.0
-        table[abs(table - 0.5) < epsilon] = 0.5
-        table[abs(table + 0.5) < epsilon] = -0.5
+    for D, fiat_table in iteritems(element.tabulate(order, points, entity)):
+        if isinstance(fiat_table, Exception):
```
wence-: I hadn't realised we're sending back an exception instance to signal the failure :-o.  Can you be more specific with the instance check?

tsfc/fem.py
```diff
@@ -41,57 +40,56 @@ def _tabulate(ufl_element, order, points):
...
+            # In the case an exception is found in the fiat table, do not
+            # perform any rounding
+            gem_fail = gem.Failure((q, phi), fiat_table)
```
wence-: OK, so we just use this to raise an error later, and the `(q, phi)` is something of the right shape.

tsfc/fem.py
```diff
@@ -41,57 +40,56 @@ def _tabulate(ufl_element, order, points):
...
+                table[abs(table) < epsilon] = 0
+                table[abs(table - 1.0) < epsilon] = 1.0
+                table[abs(table + 1.0) < epsilon] = -1.0
+                table[abs(table - 0.5) < epsilon] = 0.5
+                table[abs(table + 0.5) < epsilon] = -0.5
 
-        if spanning_degree(ufl_element) <= sum(D):
-            assert compat.allclose(table, table.mean(axis=0, keepdims=True), equal_nan=True)
-            table = table[0]
+                if spanning_degree(ufl_element) <= sum(D) and ufl_element.family() != "HDiv Trace":
```
wence-: Why can one not collapse tables in the trace case?

tsfc/fem.py
wence-: Oh, but previously `table` was a numpy array and now it's a `Literal`.  So this would have to be `ctx.index_selector(lambda i: gem.Literal(table.array[i]), mt.restriction)` I think.

tsfc/fem.py
thomasgibson: What do you mean by "more specific?" Like check explicitly for a `TraceError` rather than a general exception?

tsfc/fem.py
thomasgibson: Exactly

tsfc/fem.py
thomasgibson: While most tsfc tests pass with `gem.Literal(table.array[i])`, it completely fails the `gem.Failure` tests. It never makes it to the `raise exception` stage and gives me a `TypeError`. However, with `gem.partial_indexed(table, (i,))` everything works just fine.

tsfc/fem.py
wence-: Oh, I see, because tables aren't always tables any more.  How about updating `partial_indexed` to return a lower-rank literal if the `tensor` argument is a `Literal`?  Is that not too ugly?

thomasgibson: So it's a `ListTensor` at this point that contains an array of `Indexed` objects where the `gem.Failure` is in. When tsfc gets to this stage, `table.array` of the `ListTensor` has the form: `array([[[ Indexed(Failure(shape, TraceError()), (0, 0)), Indexed(Failure(shape, TraceError()), (0, 1)), ... ]]], dtype=object)`

I managed to find where it goes wrong. It goes wrong in the constructor for `Literal` when it forces the array as `self.array = asarray(array, dtype=float)`. I get a `TypeError` here in the case the `ListTensor` contains indexed gem failures.

miklos1: I think an `as_gem` akin to `as_ufl` would be right here. `table.array[i]` is a numpy array is `table` was a `Literal`, but a GEM expression if `table` was a `ListTensor`. Let me push a change.

tsfc/fem.py
miklos1: I prefer not making this part of TSFC know too much about specific finite elements.

wence-: OK, so you're just looking for an error of any kind.

thomasgibson: Right. At the moment, the only errors ever returned in a `fiat_table` will be a `TraceError`, but perhaps in the future there may be other kinds. I kept it as an `Exception` to keep it general.

miklos1:
I think this is now good, running the Firedrake test suite right now...

miklos1:
It didn't break Firedrake.

PR #48, opened 2016-07-21
travis: Notifications via slack

wence-:


miklos1:
OK.

wence-:
Let's see that it works first.

PR #20, opened 2016-02-16
Clean up code generation

miklos1:
Mostly clean up and docstrings for the code generation (GEM -> COFFEE) part of the pipeline. Partial clean up for `driver.py`.

PR #49, opened 2016-07-25
Don't directly use invasive integral type for quadrature rule creation

miklos1:
OK, so there are two connected things here. First, it's a TSFC/FIAT refactoring related to quadrature rule creation. This is mostly pushing things to the FIAT side. Second, `integral_type` annoyingly appears everywhere in TSFC, and this is the first step to moderate this. Specially, we dispatch on integral type to:

1. Select the integration cell
* Determine the quadrature degree we need to use on the integration cell
* Map the quadrature points back from the integration cell to the cell
* Distinguish between horizontal and vertical facets for various geometric terms
* Just know whether it's a cell or a facet integral

The idea is to lower the `integral_type` into an *integration dimension* and a list of *entity IDs* (inspired by what @thomasgibson and @dham are doing for tabulation on trace elements). For example, for a prism cell, the integration dimension is (2, 1) for a cell integral, (2, 0) for a horizontal facet integral and (1, 1) for a vertical facet integral. The entity IDs for different horizontal facet integrals are: [0] for `'exterior_facet_bottom'`, [1] for `'exterior_facet_top'`, and [0, 1] for `'interior_facet_horiz'` (these are all the facet numbers we need to consider).

So we will have switch case once to decompose the integral type, and then just use integration dimension and/or entity IDs without repeating the switch case code smell again. This pull request solves 1-2 of the issues above.

Depends on [FIAT pull request #20](https://bitbucket.org/fenics-project/fiat/pull-requests/20).


tsfc/driver.py
```diff
@@ -243,3 +250,30 @@ def facetarea():
...
+            raise NotImplementedError("integral type %s not supported" % integral_type)
+
+    if integral_type == 'exterior_facet_bottom':
+        entity_ids = [0]
+    elif integral_type == 'exterior_facet_top':
+        entity_ids = [1]
+    else:
+        entity_ids = range(len(fiat_cell.get_topology()[integration_dim]))
+
+    return integration_dim, entity_ids
```
wence-: So by `integration_dim` you mean `integration_entity` really, no? i.e. the return value is the description of the topological entity you're integrating over as fiat understands it, and the ids of those entities?

miklos1: Not sure what you mean by `integration_entity`. Let's say if we are on a triangle, then `integration_dim` is 2 for cell integrals, and 1 for facet integrals, rather than a triangle cell for cell integrals and an interval cell for facet integrals.

I guess I'm missing a docstring here anyway.

dorugeber: Aber how do you distinguish between over a triangle integrating and overintegrating a square?  Or do you not need to?

miklos1: So what really happens is this:

    (cell, integral_type) -> (cell, integration_dim, entity_ids)

This function doesn't return the cell, because that doesn't change, just decomposes `integral_type` into `(integration_dim, entity_ids)`, but it needs the cell to do that decomposition. The cell still maintains the distinction between a triangle and a quadrilateral. Basically, integration dimension and entity IDs are only used to pass them to methods of the FIAT cells.

PR #50, opened 2016-08-12
Reference geometry improvements

miklos1:
This change represents TSFC/FIAT refactoring related to reference geometry. Furthermore, dispatching on the integral type is now limited to the kernel interface and `lower_integral_type` in the driver.

We also retire a few superseded solutions:
 * "`ufc_geometry.h`" (in `tsfc/geometric.py`)
 * The `FacetManager`.

Depends on [FIAT pull request #24](https://bitbucket.org/fenics-project/fiat/pull-requests/24).

Must be merged at the same time with firedrakeproject/firedrake#855.

### DIFF
--- a/requirements-ext.txt
+++ b/requirements-ext.txt
@@ -1,3 +1,2 @@
 numpy
 singledispatch
-sympy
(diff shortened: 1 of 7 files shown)


miklos1:
Added some docstrings and a test case for cell-facet Jacobian generation.

PR #51, opened 2016-08-17
Support for forms with multiple domains

wence-:
compute_form_data provides us with the index into form.ufl_domains()
that allows us to extract the correct domain for a given integral.  We
need to remember this in the Kernel and send it back to the outside
world.


miklos1:
This is fine.

PR #52, opened 2016-09-12
Flexible indexing

miklos1:
This change introduces a new indexed node type which can code generate things like `w[12 + 3*i + j]`. This indexed type can only be applied to terminal variables, and can implement reshapes (splitting dimensions only, no flattening) and views on tensors.

- This eliminates the need for the C hacks TSFC still uses for reshapes in `kernel_interface.py`.
- This eliminates the need for C syntax features that aren't supported by COFFEE (e.g. `const double w[restrict 10]` not supported by COFFEE, but `const double *restrict w` works).
- Improves compatibility with C++. (Though not compiling with C++ is a feature.)
- @dham: helps with `ufl.TensorElement`s on the FInAT branch.
- @blechta: helps avoiding [this hack](https://bitbucket.org/fenics-project/ffc/commits/2053f4592414b9bf5ee461f58701a1f200d621c1?at=master).

### DIFF
--- a/gem/gem.py
+++ b/gem/gem.py
@@ -479,6 +479,10 @@ class ComponentTensor(Node):
     def __new__(cls, expression, multiindex):
         assert not expression.shape
 
+        # Empty multiindex
+        if not multiindex:
+            return expression
+
         # Collect shape
         shape = tuple(index.extent for index in multiindex)
         assert all(shape)
@@ -613,7 +617,4 @@ def reshape(variable, *shapes):
             indices.append(i)
         dim2idxs.append((0, tuple(idxs)))
     expr = FlexiblyIndexed(variable, tuple(dim2idxs))
-    if indices:
-        return ComponentTensor(expr, tuple(indices))
-    else:
-        return expr
+    return ComponentTensor(expr, tuple(indices))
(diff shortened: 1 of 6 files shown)


gem/gem.py
```diff
@@ -30,7 +30,7 @@
            'MaxValue', 'Comparison', 'LogicalNot', 'LogicalAnd',
            'LogicalOr', 'Conditional', 'Index', 'VariableIndex',
            'Indexed', 'ComponentTensor', 'IndexSum', 'ListTensor',
-           'partial_indexed']
+           'partial_indexed', 'reshape']
```
miklos1: TODO: export `FlexiblyIndexed` as well.

tsfc/coffee.py
```diff
@@ -314,3 +314,54 @@ def _expression_indexed(expr, parameters):
...
+
+        if len(iss) == 0:
+            rank.append(off)
+            offset.append((1, 0))
+        elif len(iss) == 1:
+            (i, s), = iss
+            rank.append(parameters.index_names[i])
+            offset.append((s, off))
+        else:
+            raise NotImplementedError("COFFEE may break in this case.")
```
miklos1: This code path is exercised in FInAT.

miklos1:
I think this is now fine to merge.

PR #53, opened 2016-09-18
Support UFC kernel interface

blechta:
UFC tested by FFC regression tests. PyOP2 not tested for correctness of generated code. Can you test it, @miklos1. 

`kernel_interface.py`, `pyop2.py`, `ufc.py` need some sort of rewrite, streamlining the codeflow. Hence no need for merging now, just throwing a partial work in for testing and asking politely to avoid conflicts, if possible, as I'm not able to fix it earlier than in three weeks.

Also `FlexiblyIndexed` to be employed in `ufc.py` later.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,4 +4,4 @@ version = "0.0.1"
 
 setup(name="tsfc",
       version=version,
-      packages=["gem", "tsfc"])
+      packages=["gem", "tsfc", "tsfc.backends"])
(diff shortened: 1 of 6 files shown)


blechta:
I can't understand the failure on Travis. Locally it works for me.

miklos1:
My immediate guess (I haven't checked) for the Travis failure is that the new package (`tsfc.backends`) needs to be added to `setup.py`.

miklos1:
@blechta: I have given you write access to this repo, so that you can push branches here. (But don't push to master.)

blechta:
Thanks :wink:. Will create a new PR from the branch in the repo.

PR #54, opened 2016-09-18
Support UFC kernel interface

blechta:
UFC tested by FFC regression tests. PyOP2 not tested for correctness of generated code. Can you test it, @miklos1?

`kernel_interface.py`, `pyop2.py`, `ufc.py` need some sort of rewrite, streamlining the codeflow. Hence no need for merging now, just throwing a partial work in for testing and asking politely to avoid conflicts, if possible, as I'm not able to fix it earlier than in three weeks.

Also `FlexiblyIndexed` to be employed in `ufc.py` later.

Must be merged at the same time with firedrakeproject/firedrake#888.


miklos1:
I have just run this against Firedrake:
```=== 93 failed, 3541 passed, 49 skipped, 18 xfailed, 1 pytest-warnings in 849.16 seconds ===```
I will look into the failures.

In the meantime, I am trying to learn how to test this against UFC/FEniCS. I created a Python virtualenv where I installed FFC + TSFC, and trying to run the FFC test suite. Will I need DOLFIN tests as well, i.e. are there any DOLFIN tests that exercise the form compiler, or are those tests all in FFC?

**Comment by miklos1** [comment]
So it seems that most Firedrake failures are because UFL interpolation in Firedrake uses TSFC internal APIs which are changed here. Other than that, the only failure is getting an unexpected number when manually specifying a quadrature degree:
```
======================================== FAILURES =========================================
_________________________________ test_manual_quadrature __________________________________

    def test_manual_quadrature():
        mesh = UnitSquareMesh(4, 4)
        V = FunctionSpace(mesh, "CG", 3)
        f = Function(V).interpolate(Expression("pow(x[0], 3)"))
    
        f1 = f*dx(degree=3)
        f2 = f*dx(degree=1)
        f3 = f*dx
    
        assert np.allclose(assemble(f1), assemble(f3))
    
        assert np.allclose(assemble(f1), 0.25)
    
>       assert np.allclose(assemble(f2), 0.244791666666)
E       assert <function allclose at 0x7f1e768a7f50>(0.24999999999967257, 0.244791666666)
E        +  where <function allclose at 0x7f1e768a7f50> = np.allclose
E        +  and   0.24999999999967257 = assemble(Form([Integral(Coefficient(WithGeometry(FunctionSpace(<firedrake.mesh.MeshTopo...nge', triangle, 1), dim=2), 1), 'everywhere', {'quadrature_degree': 1}, None)]))

tests/regression/test_manual_quadrature.py:19: AssertionError
================================ 1 failed in 1.67 seconds =================================
```

I have been thinking about moving UFL interpolation from Firedrake to TSFC, so that Firedrake would not have to touch TSFC internals, but I have not figured out a proper contract between them yet. Maybe it is time I should do that.

**Comment by miklos1** [comment]
On the other hand, I get some failures testing against FFC, because the floating-point numbers in the source code slightly differ compared to the reference. This can mean that

- the last printed digit is different, or
- the number is printed with one more or less digits.

Consequently, the source code differs, and in one case the computed results differ slightly.

There is another legitimate change in generated code due to fixing #55 in coneoproject/COFFEE@8ba0fda8a0b3126beae08a1c6a125b8d8ec6d8b5 which prints more parenthesis.

**Comment by blechta** [comment]
> Will I need DOLFIN tests as well, i.e. are there any DOLFIN tests that exercise the form compiler, or are those tests all in FFC?

FFC regression tests are important part. They compare results of `tabulate_tensor` (with some random numbers as input) to reference produced with `auto` representation, mostly `quadrature`. DOLFIN tests are somewhat useful as well but they don't have an ambition to exercise numerical correctness of all the elements/forms.

> On the other hand, I get some failures testing against FFC, because the floating-point numbers in the source code slightly differ compared to the reference.

I expect you've tested this with branch `jan/tsfc-rebase-2`.

Numerical precision in the FFC tests needs some attention. I plan to look on this. But I don't think that this is important now. The tests essentially pass (up to some parentheses and numerical precision). I guess that most concern is now with rewriting the patch into reasonable piece of code. Once you're able to run `cd <ffc>/demo && ffc -r tsfc LessOrMoreComplicatedFormFile.ufl` you're probably fine. For numerical tests `cd <ffc>/test/regression && python test.py --skip-download --skip-auto --skip-quad --skip-uflacs LessOrMoreComplicatedFormFile.ufl` or the same without the form file.

> which prints more parenthesis.

I'll try to sort out the reference soon.

> I am trying to learn how to test this against UFC/FEniCS

I'll try setup testing again tsfc on http://fenics-bamboo.simula.no:8085 so that it is more convenient.

**Comment by miklos1** [comment]
I added a commit which alongside the `tsfc/support-ufc` branch in firedrake fixes most of the Firedrake test suite. The only remaining failure is the manual quadrature above.

File: tsfc/driver.py [hunk 1]
```diff
@@ -60,21 +53,17 @@ def compile_form(form, prefix="form", parameters=None):
...
     """Compiles a UFL integral into an assembly kernel.
 
     :arg integral_data: UFL integral data
     :arg form_data: UFL form data
     :arg prefix: kernel name will start with this string
     :arg parameters: parameters object
+    :arg backend: output format
     :returns: a kernel, or None if the integral simplifies to zero
     """
-    # Remove these here, they're handled below.
```
miklos1: Can you explain your tinkering with the parameters in this function? In Firedrake, the global parameters (coming from the `parameters` argument of this function) generally override the parameters specified as integral metadata, with these exceptions here: if `quadrature_degree` or `quadrature_rule` are set to `auto` then they don't override the per-integral specified quadrature degree (if any). Since the `auto` values are no longer removed in this patch, they will always override the per-integral degree in [line 177](https://github.com/firedrakeproject/tsfc/pull/54/files#diff-4906bef44a1f7f6a1cbf093c68d66c37R177), causing the test failure with `f*dx(degree=1)` for example.

Can you explain what is the expected behaviour in FFC and what values are coming from the `parameters` argument?

blechta: Sorry, haven't noticed of the side effect. Good that your tests spotted it. I had similar problems in FFC. Degrees `[None, "auto", "default", -1, "-1"]` and rules `[None, "auto", "default"]` all mean default. I wanted to be able to pass integral data to `tsfc.driver.compile_integral` without a need to copy it and translate metadata to the form tsfc accepts, i.e., only `"auto"` as degree and scheme.

It would be very helpful if you could support those aliases to `"auto"`. In any case, feel free to fix according to your needs now and I will check later what the consequence to FFC is.

tsfc/driver.py
```diff
@@ -89,21 +78,24 @@ def compile_integral(integral_data, form_data, prefix, parameters):
...
+        builder.set_coordinates(coordinates, mode='list_tensor')
     else:
         # Otherwise we use the approach that might be faster (?)
-        builder.set_coordinates(coordinates, "coords")
+        builder.set_coordinates(coordinates)
 
     builder.set_coefficients(integral_data, form_data)
 
+    builder.set_cell_orientations()
+    builder.set_facets()
```
miklos1: I think these could be just set up in the constructor of the builder without explicitly asking.

tsfc/driver.py
```diff
@@ -143,7 +130,7 @@ def coefficient(ufl_coefficient, r):
                              point_index=quadrature_index,
                              coefficient=coefficient,
                              index_cache=index_cache)
-        if parameters["unroll_indexsum"]:
+        if parameters.get("unroll_indexsum"):
```
miklos1: I guess we could just move the sanitization of parameters to `compile_integral` from `compile_form`, then everything would be here when expected.

tsfc/ufl_utils.py
```diff
@@ -21,6 +23,32 @@
                                      construct_modified_terminal)
 
 
+def compute_form_data(form,
```
miklos1: Ah, so this is here so that FFC does not need to duplicate the exact same options that TSFC works with.

tsfc/driver.py
```diff
@@ -196,12 +178,13 @@ def facetarea():
 
         # Check if the integral has a quad degree attached, otherwise use
         # the estimated polynomial degree attached by compute_form_data
-        quadrature_degree = params.get("quadrature_degree",
-                                       params["estimated_polynomial_degree"])
+        quad_degree = params.get("quadrature_degree")
+        if quad_degree in [None, "auto", "default", -1, "-1"]:
```
miklos1: I would prefer to have this great variety of equivalent default values reduced to one in the `tsfcrepr` part of FFC, before calling this function.

tsfc/backends/ufc.py
```diff
@@ -0,0 +1,392 @@
...
+                # if interior_facet:
+                #     i = gem.Index()
+                #     expression = gem.ComponentTensor(
+                #         gem.Indexed(gem.Variable(name, (num_coefficients,) + (2,)),
+                #                     (coefficient_numbers[j], i,)),
+                #         (i,))
+            else:
+                # Mixed/vector/tensor constant/real
+                # FIXME: Tensor case is incorrect. Gem wants shaped expression, UFC requires flattened.
+                indices = tuple(gem.Index() for i in six.moves.xrange(len(shape)))
```
miklos1: We will have to add `six` to the dependencies (`requirements-ext.txt`).

tsfc/driver.py
```diff
@@ -115,12 +107,7 @@ def compile_integral(integral_data, form_data, prefix, parameters):
     def cellvolume(restriction):
         from ufl import dx
         form = 1 * dx(domain=mesh)
-        fd = compute_form_data(form,
-                               do_apply_function_pullbacks=True,
-                               do_apply_integral_scaling=True,
-                               do_apply_geometry_lowering=True,
-                               do_apply_restrictions=True,
-                               do_estimate_degrees=True)
+        fd = ufl_utils.compute_form_data(form, preserve_geometry_types=())
```
miklos1: Just a note, but whether `CellVolume` and `FacetArea` are specified to be preserved does not matter much. In this case, and similarly for `facetarea` you can see that the form does not contain any of these geometric types. When the mesh geometry is non-affine, UFL will not lower these types anyway. And when UFL lowers these to reference quantities, TSFC still handles the result correctly (though this case has not been tested recently).

The reason I decided to always preserve these geometry types is because

- for some reason I get slightly more common subexpressions eliminated.
- restriction propagation for these types was bugged in UFL, which is worked around by handling them always in TSFC.

miklos1:
I will try to fix the quadrature degree problem and push a fix.

tsfc/driver.py
blechta: For sure.

tsfc/driver.py
blechta: Ok. Note that we don't use `compile_form`. We call `compile_integral` from FFC.

tsfc/driver.py
blechta: That should be possible. I've done it only to avoid copying integral data and form data because I'm not sure I'm allowed to change their metadata.

tsfc/ufl_utils.py
blechta: Exactly. I've spotted that you've changed them in the mean time so it scared me up.

tsfc/backends/ufc.py
blechta: Please do.

tsfc/driver.py
```diff
@@ -179,6 +179,8 @@ def facetarea():
         params = {}
         # Record per-integral parameters
         params.update(integral.metadata())
+        if params.get("quadrature_rule") == "default":
```
miklos1: I want to flag this up as a topic for further discussion. While this commit helps to satisfy both the Firedrake and the FFC test suite, the usage the `quadrature_rule` parameter in FEniCS/Firedrake is quite unclear. TSFC currently expects a `QuadratureRule` object here, while FFC seems to provide strings. (Maybe the `scheme` argument of `ffc.fiatinterface.create_quadrature`?) However, none is these usages seems to have tested in either Firedrake or FFC.

blechta: Agree. I had this on mind before but that was not so important.

tsfc/kernel_interface.py
```diff
@@ -119,400 +107,66 @@ def arguments(self, arguments, indices):
...
-        for multiindex in numpy.ndindex(shape):
-            references.append(coffee.Symbol(name, multiindex))
-
-    restriction_shape = []
-    for e in elements:
-        if isinstance(e, MixedElement):
-            restriction_shape += [len(e.elements()),
-                                  e.elements()[0].space_dimension()]
+class KernelBuilder(KernelBuilderBase):
+    def __new__(cls, backend, *args, **kwargs):
```
miklos1: I don't like factory functions of this kind, and elaborate switch case generally smell. Fortunately, unless one has to read external representations, these are easy to avoid. In this case one could just pass in the backend module to `compute_integral` directly. I'm testing this change now.

tsfc/driver.py
wence-: My reasoning for being able to provide a `QuadratureRule` object is that a third-party user can directly provide the quadrature without needing to go in and add some information to a mapping table.

tsfc/backends/ufc.py
```diff
@@ -274,29 +274,33 @@ def _prepare_arguments(arguments, indices, interior_facet=False):
...
-    prepare = [zero, reshape]
-
-    return funarg, prepare, expressions, []
+        shape = space_dimensions
+        expressions = [gem.FlexiblyIndexed(varexp, ((0, tuple(zip(indices, space_dimensions))),))]
+
+    zero = coffee.FlatBlock(
+        str.format("memset({name}, 0, {size} * sizeof(*{name}));{nl}",
+                   name=funarg.sym.gencode(), size=numpy.product(shape),
+                   nl=os.linesep)
```
blechta: Small comment: `os.linesep` is in principle wrong as it generates different code on different architectures. Should be just `'\n'`

blechta:
@miklos1, please note that FFC regression tests are not run with TSFC representation yet on either bitbucket pipelines or bamboo. TSFC is just skipped on import error.

To make it work on bamboo, is it just about adding `tsfc` to `requirements.txt`, @johannesring? Note that there might be troubles, if requirements.txt is resolved recursively - tsfc pulls firedrake versions of FIAT and UFL.

tsfc/backends/ufc.py
miklos1: Thanks, fix is on the way. I'm cleaning things up a bit now, and hopefully we can merge this soon.

miklos1:
We should discuss a few naming conventions here. (I know, code works irrespective of how things are named, but in the long term things become unintuitive and painful if we get this wrong.) For this reason, I would like to summon @dham and @wence- here.

Names of the "backends":

- **Firedrake/PyOP2:** It is currently called PyOP2, but I think the name Firedrake would be more appropriate here. PyOP2 does not have specific calling conventions for finite element forms, it is a framework for running local kernels over unstructured meshes, and how the kernel is called is very configurable. For finite element assembly kernels, Firedrake configures how this is called, and TSFC matches the same calling conventions in its kernel interface.
- **UFC/DOLFIN/FEniCS:** I guess the name UFC is fine here, because UFC has well defined calling conventions. DOLFIN would probably not be very off either, since AFAIK DOLFIN is the only consumer of UFC kernels, while at the other side of the interface there have been multiple form compilers.

The term "backend" itself is something we should think about in the light of @rckirby's and @inducer's work. So there are two kinds of things that we might call "backend", and they are different and completely orthogonal to each other:

- The *output format*, or rather the code generation module for that output format. Currently, this is COFFEE, but Andreas Klöckner has also sketched a [loo.py backend](https://github.com/inducer/tsfc/tree/loopy-output) which may be merged eventually.
- How the generated code interfaces the outside world, formerly known as *kernel interface*. Firedrake and UFC fall in this category.

So any comments or ideas on what the right names should be?

johannesring:
@blechta - no I don't think so because we use `fenics-build` from `dockerfiles/dev-env/bin/fenics-build` in the `docker` repo, which runs `pip install` with `--no-deps`.

blechta:
@johannesring, so can we add `https://github.com/firedrakeproject/tsfc`, `https://github.com/coneoproject/COFFEE`, `singledispatch`, `networkx`, `pulp` to FFC dependencies? Push to `tsfc` and `COFFEE` could trigger the build. @miklos1, is `pulp` needed? I don't remember installing it locally.

miklos1:
@johannesring, @blechta: PuLP and GLPK (glpsol) are needed once the COFFEE optimisations are enabled. If COFFEE is only used to create and print the AST, then these are not necessary. For glpsol, one needs the glpk-utils package on Ubuntu.

blechta:
We can forget it now. Optimization is not in place yet. So we need to install just `tsfc`, `COFFEE`, `singledispatch`, `networkx`.

wence-:
So the two distinct things we have are:

- kernel interface: Firedrake and UFC prescribe a data layout for incoming arguments, along with what the signature of the called kernel is.  This has some bearing on the gem that tsfc generated because some of the indexing changes.  Is that right?  The gem tree doesn't currently describe the input and output arguments (the codegen for those happens "by hand", right?).  During discussions at PDESoft, it seemed that various DUNE people might also be interested, they presumably have their own data ordering requirements.

- codegen engine: Currently COFFEE, maybe loopy.  This unfortunately cuts into the kernel interface right now, because when prescribing the function arguments and so forth, one must describe them in the codegen language, not gem.  Is that right?

So this would suggest we would like to have the two things separate: _kernel_interface_ and _codegeneration_.  To do so, does that mean we need gem-like way of describing these things that the codegeneration engines can interpret?

blechta:
> two things separate: kernel_interface and codegeneration

Shouldn't codegeneration know about a layout of arguments in order to do the optimization?

wence-:
>> two things separate: kernel_interface and codegeneration

> Shouldn't codegeneration know about a layout of arguments in order to do the optimization?

COFFEE does optimisation of loop nests, but no data transformations.  Loopy can do data transformations, but in that case, you've told it about the data layout I think (@inducer ?).

But the point is more that the caller of the kernel that comes out of tsfc is the one who is going to pass in the data.  And therefore they prescribe the data ordering that the kernel has to obey (as part of the kernel interface).  The current contract with the code generation/optimisation is that it can't change the kernel signature, but can do whatever it likes inside the kernel body.  AIUI, a putative loopy interface would have loopy describing both the local kernel iteration, *and also* the iteration over a mesh, so this is a bit different.

johannesring:
I have installed `tsfc`, `COFFEE`, `singledispatch` and `networkx` for `FFC` on Bamboo now. I'm not sure about adding a trigger for push to `tsfc` and `COFFEE` though. This will trigger a rebuild of all `FFC` and `DOLFIN` branches, which is quite time consuming.

blechta:
Ok, let's keep it without a trigger for now. @miklos1, do you have an account on bamboo so that you can start a build manually?

miklos1:
@blechta, @johannesring: I don't have a bamboo account yet, if I sign up, will it be a FEniCS-only bamboo account?

blechta:
I think http://fenics-bamboo.simula.no:8085 is FEniCS specific. I think atlassian/bitbucket account works there.

There's an import error http://fenics-bamboo.simula.no:8085/browse/FFC-FD30-7. I can debug it in the evening. (The way to fetch the container with the build https://bitbucket.org/fenics-project/testing/issues/1.)

johannesring:
@miklos1 - I have given you access to manually start builds of FIAT and FFC on out Bamboo server now. Let me know if you need permission for anything else.

miklos1:
Some of the FFC regression tests could be easily fixed, but I think we can merge this first and deal with those after. There seem to be files on the FFC side which are copies of TSFC and are probably not used anymore, but I guess @blechta can take care of that branch.

blechta:
I could overlook this in the evening hopefully. But from overlooking your commits it seems good, I think it can be merged immediately. I'll try to debug bamboo in the evening. And push the new reference - but that might be a bit involved.

blechta:
I think I've already remove TSFC files from FFC when backporting the changes here by this PR. But I'll check.

blechta:
I'm an idiot. Bamboo fails, of course, because it pulls tsfc master. I think you can merge it, @miklos1.

blechta:
@miklos, just caugh a sight of `memset(A, 0, 1.0 * sizeof(*A));` in the generated code http://fenics-bamboo.simula.no:8085/browse/FFC-FD30-12/artifact/RT/error.log/error.log. Why does it happen?

blechta:
Of course, it's because `numpy.product(()) == 1.0`. Can you add `dtype`?

miklos1:
@blechta: Fixed on master.

PR #58, opened 2016-09-20
Adding a wrapper class for quad flattening

thomasgibson:
This PR adds a wrapper class the flattens the entity dofs of an element defined on an Interval x Interval to a `FiredrakeQuadrilateral`. We need this for the new tabulate API in the FIAT pull request [here](https://bitbucket.org/fenics-project/fiat/pull-requests/28/add-optional-entity-argument-to-the/diff).

### DIFF
--- a/tsfc/fiatinterface.py
+++ b/tsfc/fiatinterface.py
@@ -90,10 +90,13 @@ class FlattenToQuad(FIAT.FiniteElement):
         nodes = element.dual.nodes
         self.ref_el = FiredrakeQuadrilateral()
         entity_ids = element.dual.entity_ids
+
         flat_entity_ids = {}
         flat_entity_ids[0] = entity_ids[(0, 0)]
-        flat_entity_ids[1] = dict(enumerate(entity_ids[(0, 1)].values() +
-                                            entity_ids[(1, 0)].values()))
+        flat_entity_ids[1] = dict(enumerate(
+            [v for k, v in sorted(entity_ids[(0, 1)].items())] +
+            [v for k, v in sorted(entity_ids[(1, 0)].items())]
+        ))
         flat_entity_ids[2] = entity_ids[(1, 1)]
         self.dual = DualSet(nodes, self.ref_el, flat_entity_ids)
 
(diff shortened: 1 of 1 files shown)


tsfc/fiatinterface.py
```diff
@@ -145,21 +227,8 @@ def _(element, vector_is_mixed):
         # Can't use create_element here because we're going to modify
         # it, so if we pull it from the cache, that's bad.
```
miklos1: This is no longer true.

tsfc/fiatinterface.py
```diff
@@ -76,6 +76,88 @@
...
+        :arg element: a fiat element
+        """
+        self.element = element
+        nodes = element.dual.nodes
+        self.ref_el = FiredrakeQuadrilateral()
+        entity_ids = element.dual.entity_ids
+        flat_entity_ids = {}
+        flat_entity_ids[0] = entity_ids[(0, 0)]
+        flat_entity_ids[1] = dict(enumerate(entity_ids[(0, 1)].values() +
+                                            entity_ids[(1, 0)].values()))
```
miklos1: I recently spotted that even though this has been working very well for a while, it is actually wrong. Asking the `.values()` of a `dict` returns the values in an unspecified order. Since the keys are `range(n)` for a small `n`, by pure implementation luck this returns the ordering that we need.

miklos1:
@thomasgibson: I can fix these things if you don't have time to deal with them now. Or we can do it together if you need help.

thomasgibson:
@miklos1 Thanks for the comments. I have no problem fixing these! I'll have a good look at it tomorrow morning when I'm in the office

tsfc/fiatinterface.py
thomasgibson: So we no longer need `element = convert(element, vector_is_mixed)`?

wence-: Instead you can use `element = create_element(element, vector_is_mixed)` I think

miklos1:
This will be merged with [FIAT pull request #28](https://bitbucket.org/fenics-project/fiat/pull-requests/28).

PR #63, opened 2016-10-13
Use precision and epsilon from parameters

### LINKED ISSUE #61: Use precision and epsilon from parameters
The default values of `PRECISION` in `constants.py` and `epsilon` in `fem.py` are good. But could those be overridden by parameters `precision` and `epsilon`? (Note that `PRECISION` is used in `coffee.py` and `epsilon` in `fem.py`.)

This serves for making FFC regression test output reproducible.

blechta:
Resolves #61.

### DIFF
--- a/tsfc/coffee.py
+++ b/tsfc/coffee.py
@@ -16,7 +16,7 @@ import coffee.base as coffee
 
 from gem import gem, impero as imp
 
-from tsfc.constants import SCALAR_TYPE, PRECISION
+from tsfc.constants import SCALAR_TYPE
 from tsfc.logging import logger
 
 
@@ -24,11 +24,12 @@ class Bunch(object):
     pass
 
 
-def generate(impero_c, index_names, roots=(), argument_indices=()):
+def generate(impero_c, index_names, parameters, roots=(), argument_indices=()):
     """Generates COFFEE code.
 
     :arg impero_c: ImperoC tuple with Impero AST and other data
     :arg index_names: pre-assigned index names
+    :arg parameters: TSFC parameters
     :arg roots: list of expression DAG roots for attaching
         #pragma coffee expression
     :arg argument_indices: argument indices for attaching
@@ -36,21 +37,23 @@ def generate(impero_c, index_names, roots=(), argument_indices=()):
         to the argument loops
     :returns: COFFEE function body
     """
-    parameters = Bunch()
-    parameters.declare = impero_c.declare
-    parameters.indices = impero_c.indices
-    parameters.roots = roots
-    parameters.argument_indices = argument_indices
+    params = Bunch()
+    params.declare = impero_c.declare
+    params.indices = impero_c.indices
+    params.roots = roots
+    params.argument_indices = argument_indices
 
-    parameters.names = {}
+    params.names = {}
     for i, temp in enumerate(impero_c.temporaries):
-        parameters.names[temp] = "t%d" % i
+        params.names[temp] = "t%d" % i
 
     counter = itertools.count()
-    parameters.index_names = defaultdict(lambda: "i_%d" % next(counter))
-    parameters.index_names.update(index_names)
+    params.index_names = defaultdict(lambda: "i_%d" % next(counter))
+    params.index_names.update(index_names)
 
-    return statement(impero_c.tree, parameters)
+    params.precision = parameters["precision"]
+
+    return statement(impero_c.tree, params)
 
 
 def _coffee_symbol(symbol, rank=()):
@@ -171,7 +174,7 @@ def statement_evaluate(leaf, parameters):
             return coffee.Decl(SCALAR_TYPE,
                                _decl_symbol(expr, parameters),
                                coffee.ArrayInit(array_expression(expr.array),
-                                                precision=PRECISION))
+                                                precision=parameters.precision))
         else:
             ops = []
             for multiindex, value in numpy.ndenumerate(expr.array):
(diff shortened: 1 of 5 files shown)


tsfc/coffee.py
```diff
@@ -296,7 +299,7 @@ def _expression_scalar(expr, parameters):
     if isnan(expr.value):
         return coffee.Symbol("NAN")
     else:
-        return coffee.Symbol(("%%.%dg" % (PRECISION - 1)) % expr.value)
+        return coffee.Symbol(("%%.%dg" % (parameters.precision - 1)) % expr.value)
```
miklos1: I don't quite remember why this minus one here. It seems to mimic `epsilon`, but `epsilon` is used as an absolute tolerance for rounding, while with `%g` this means a relative tolerance. Maybe I wasn't aware of this distinction?

tsfc/fem.py
```diff
@@ -75,23 +71,26 @@ def tabulate(ufl_element, order, points):
         yield c, D, table
 
 
-def make_tabulator(points):
-    """Creates a tabulator for an array of points."""
-    return lambda elem, order: tabulate(elem, order, points)
+def make_tabulator(points, epsilon):
+    """Creates a tabulator for an array of points with rounding
+    parameter epsilon."""
+    return lambda elem, order: tabulate(elem, order, points, epsilon)
```
miklos1: This parametrization/configuration could have a much shorter diff if Python had support for dynamic scoping.

tsfc/coffee.py
```diff
@@ -16,41 +16,44 @@
...
-    parameters = Bunch()
-    parameters.declare = impero_c.declare
-    parameters.indices = impero_c.indices
-    parameters.roots = roots
-    parameters.argument_indices = argument_indices
+    params = Bunch()
+    params.declare = impero_c.declare
+    params.indices = impero_c.indices
+    params.roots = roots
+    params.argument_indices = argument_indices
```
miklos1: Hmm... diff could be shorter if `params` / `parameters` would be the other way around.

tsfc/coffee.py
miklos1: It seems that the minus one was added precisely in af4145ab. I guess it may be needed to get the expected error in some test cases, maybe @wence- remembers better? The date suggests it was a release day change.

tsfc/fem.py
blechta: Yeah, but that would make the code less obvious to follow. Now it's explicit.

tsfc/coffee.py
blechta: Isn't it `1` valid digit before comma and `precision - 1` valid digits after comma?

tsfc/coffee.py
blechta: Seemed to me a bit more consistent with some other piece of code: `parameters` for passed tsfc parameters and `params` for module internals...

tsfc/coffee.py
miklos1: http://stackoverflow.com/questions/30580481/why-does-e-behave-different-than-g-in-format-strings claims 'no' for `%g`.

tsfc/coffee.py
miklos1: OK

tsfc/driver.py
```diff
@@ -254,17 +254,25 @@ def facetarea():
     return builder.construct_kernel(kernel_name, body)
 
 
-def compile_expression_at_points(expression, points, coordinates):
+def compile_expression_at_points(expression, points, coordinates, parameters=None):
```
blechta: Is it ok to add parameters kwarg?

tsfc/coffee.py
blechta: I could compare this with uflacs and quadrature. Do you a form which goes through this codepath?

blechta: Analogically for `0.23456789` tsfc does `0.23457` while quad and uflacs do `0.234568`.

miklos1: So I guess it would be the same without the minus one?

blechta: You mean same output of tsfc and quad/uflacs? Yes. COFFEE generated arrays seem to have correct number of significant digits. Sometimes they're pone digit shorter but that could be trailing zero. Minus sign or number of digits before comma do not seem to play a role.

`return coffee.Symbol(("%%.%dg" % (parameters.precision - 1)) % expr.value)` yields one less digit than expected.

miklos1:
I'm going to test this branch and the minus one thing against Firedrake now.

tsfc/driver.py
miklos1: We don't use it, but since `compile_ufl` now needs parameters, I guess it's fine. I might undo this if I manage simplify the argument list of `compile_ufl` during some ongoing refactorings.

blechta: Who uses it?

miklos1: I meant that the newly added `parameters` keyword argument is not being used anywhere yet. This function is called [here](https://github.com/firedrakeproject/firedrake/blob/master/firedrake/interpolation.py#L168) for [interpolating UFL expressions](http://www.firedrakeproject.org/interpolation.html#ufl-expressions).

miklos1:
Report:

1. Minus one precision: no failures in Firedrake without the minus one. I wonder if @wence- might have seen numbers like `3.000000000001` or `4.99999999999` when committing this change.
2. This branch against Firedrake: there are some failures due to importing `tsfc.constants.PRECISION` in some legacy FFC-like code, but the fix is trivial.

blechta:
:smiley:

blechta:
Can you ping me with minus one issue where it takes place further? I need to update FFC regression test output then.

miklos1:
@blechta: OK

PR #65, opened 2016-10-13
Recognize numpy.integer as gem.IndexBase

blechta:
Allows to run FFC regression tests with Py3.

### DIFF
--- a/gem/gem.py
+++ b/gem/gem.py
@@ -344,6 +344,7 @@ class IndexBase(with_metaclass(ABCMeta)):
     pass
 
 IndexBase.register(int)
+IndexBase.register(numpy.integer)
 
 
 class Index(IndexBase):


miklos1:
Can we just use `numbers.Integral`?

blechta:
Does not seem to work.

```
Python 2.7.9 (default, Apr  2 2015, 15:33:21) 
[GCC 4.9.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy, numbers
>>> issubclass(int, numbers.Integral)
True
>>> issubclass(numpy.int64, numbers.Integral)
True
>>> issubclass(numpy.int32, numbers.Integral)
False
>>> issubclass(numpy.integer, numbers.Integral)
False
```

```
Python 3.4.3 (default, Mar 26 2015, 22:03:40) 
[GCC 4.9.2] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy, numbers
>>> issubclass(int, numbers.Integral)
True
>>> issubclass(numpy.int64, numbers.Integral)
False
>>> issubclass(numpy.int32, numbers.Integral)
False
>>> issubclass(numpy.integer, numbers.Integral)
False
```

miklos1:
This is odd. I'm getting:
```
Python 2.7.12 (default, Jul  1 2016, 15:12:24) 
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy, numbers
>>> issubclass(int, numbers.Integral)
True
>>> issubclass(numpy.int64, numbers.Integral)
True
>>> issubclass(numpy.int32, numbers.Integral)
True
>>> issubclass(numpy.integer, numbers.Integral)
True
```
```
Python 3.5.2 (default, Sep 10 2016, 08:21:44) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy, numbers
>>> issubclass(int, numbers.Integral)
True
>>> issubclass(numpy.int64, numbers.Integral)
True
>>> issubclass(numpy.int32, numbers.Integral)
True
>>> issubclass(numpy.integer, numbers.Integral)
True
```

blechta:
Maybe CPython and NumPy versions? Seems that we should stay here explicit. `issubclass(numpy.int<foo>, numpy.integer)` hopefully holds on every implementation...

blechta:
Both examples were with NumPy 1.8.2.

miklos1:
I have 1.11.0 and 1.11.1.

miklos1:
There are a couple other places with `isinstance(..., int)` which are checking for fixed indices. We adjust can all those instance check (not so good), create `FixedIndex` abstract class and register `int` and `numpy.integer` and update those instace checks to `FixedIndex`, or just cast fixed indices to `int` as they come. But where are `numpy.integer`s coming from?

blechta:
> But where are numpy.integers coming from?

Let me check...

blechta:
It's `indices` at https://github.com/firedrakeproject/tsfc/blob/master/tsfc/kernel_interface/ufc.py#L208. Already `numpy.int64` with Py2:

```
Python 2.7.9 (default, Apr  2 2015, 15:33:21) 
[GCC 4.9.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy
>>> num_nodes, gdim = 3, 2
>>> indices = numpy.arange(num_nodes * gdim).reshape(num_nodes, gdim).transpose().flatten()
>>> type(indices[0])
<type 'numpy.int64'>
```

miklos1:
[A sneaky comment on StackOverflow](http://stackoverflow.com/questions/22471644/how-do-i-check-if-a-numpy-dtype-is-integral#comment34487455_22476290) suggests that the `numbers.Integral` vs. `numpy.integer` problem was solved in numpy 1.9.

miklos1:
> It's indices at https://github.com/firedrakeproject/tsfc/blob/master/tsfc/kernel_interface/ufc.py#L208.

I think I would just convert them right there then, if that's sufficient to fix the issue at hand.

blechta:
Ok, no problem.

miklos1:
Superseded by #67.

Issue #64, opened 2016-10-13
Formatting precision of floats seems to be buggy


blechta:
```
element = FiniteElement("Lagrange", triangle, 1)
v = TestFunction(element)
L = 1.23456789*v*dx
```
with `precision=6` gives `1.2346` while uflacs and quadrature do `1.23457`. On the other hand COFFEE formatting of arrays (different code path) produces right and consistent answer.

See the discussion at https://github.com/firedrakeproject/tsfc/pull/63#discussion_r83212244 for details.

blechta:
Further formatting of zero is not reproducible. One can get `0.0` or `-0.0`, see http://fenics-bamboo.simula.no:8085/browse/FFC-FD30-38/artifact/RT/error.log/error.log.

blechta:
Would this patch
```
diff --git a/coffee/base.py b/coffee/base.py
index b50b798..5a46112 100644
--- a/coffee/base.py
+++ b/coffee/base.py
@@ -269,6 +269,7 @@ class ArrayInit(Expr):
         elif isnan(v):
             return "NAN"
         elif abs(v - round(v, 1)) < eps:
+            if v == -0.0: v = 0.0
             return f_int % v
         else:
             return f % v
```
be acceptable to solve plus/minus zero issue?

miklos1:
This `f_int` is confusing both in `coffee.base.ArrayInit._formatter` and in `ffc.cpp.set_floating_formatting`. While the name suggests that it is formatting integers, it rounds to a multiple of 0.1 and formats with 1 decimal digit (after the decimal point).

miklos1:
Moreover a quick grep in FFC for `"epsilon"` suggests that the epsilon form compiler parameter is only being used by the tensor representation to skip small values, while for rounding tables and for all other uses `format["epsilon"]` is being used. `format["epsilon"]` does not come from `parameters["epsilon"]`, but it comes from [`parameters["precision"]`](https://bitbucket.org/fenics-project/ffc/src/2bfdac344b4edac50d478ad84d0a1c7ea2556a3f/ffc/cpp.py?at=master&fileviewer=file-view-default#cpp.py-729).

miklos1:
The reason I was a bit surprised about the minus zeros, is that if a number rounds to `-0.0`, then it must be close to zero, and if it is formatted in COFFEE, then it probably comes from a tabulation matrix, but tabulation matrix entries have already been rounded to exactly `0.0`, so a minus sign could not have arisen.

[...]

PR #66, opened 2016-10-14
Fix epsilon numerical tolerance

miklos1:
Turns out that the `format["epsilon"]` in FFC does not come from `parameters["epsilon"]`, but is calculated from `parameters["precision"]`. This change reflects this discovery.

Related to the discussion in #64, but does not fix that issue.

### DIFF
--- a/tsfc/constants.py
+++ b/tsfc/constants.py
@@ -7,8 +7,6 @@ NUMPY_TYPE = numpy.dtype("double")
 
 _PRECISION = numpy.finfo(NUMPY_TYPE).precision
 
-_EPSILON = eval("1e-%d" % (_PRECISION - 1))
-
 SCALAR_TYPE = {numpy.dtype("double"): "double",
                numpy.dtype("float32"): "float"}[NUMPY_TYPE]
 
@@ -25,9 +23,6 @@ PARAMETERS = {
 
     # Precision of float printing (number of digits)
     "precision": _PRECISION,
-
-    # Threshold for rounding FE tables to 0, +/- 1/2, +/- 1
-    "epsilon": _EPSILON,
 }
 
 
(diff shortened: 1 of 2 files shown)


miklos1:
@blechta: Currently testing this against Firedrake, but I don't expect any problems. The FFC regression test seem not to produce minus zeros with this change, but have a look yourself.

blechta:
I confirm FFC is now fine. But I'm not sure that the situation is perfectly good. Consider https://bitbucket.org/fenics-project/ffc/issues/127 and Martin also has doubts in https://bitbucket.org/fenics-project/ffc/src/2bfdac344b4edac50d478ad84d0a1c7ea2556a3f/ffc/uflacs/language/format_value.py?at=master&fileviewer=file-view-default#format_value.py-33.

blechta:
To be clear I'm for merging this for now.

miklos1:
It might be an unnecessary complexity in TSFC, but I don't think we are affected with [FFC issue 117](https://bitbucket.org/fenics-project/ffc/issues/127), or at least not in the form that is presented in the example.

We "round" numbers at three occasions:
- After receiving the FIAT tabulation, we round entries in the tabulation matrices to -1, -0.5, 0, 0.5, 1, with an absolute tolerance epsilon.
- When generating COFFEE, scalars are printed up to precision-1 digits, which is equivalent to a rounding with a relative precision (which does not destroy numbers like 1e-12).
- All non-scalar values are left for COFFEE to be printed, where they are printed up to precision accuracy, but rounding to multiples of 0.1 if close enough within an absolute tolerance epsilon = 10 * 10^(-precision).

blechta:
Ok then, thanks for a nice summary :wink:

PR #67, opened 2016-10-14
Cast numpy indices to int

miklos1:
@blechta: can you check if this solves the Python 3 problem that #65 was intended to fix?


blechta:
Yes :wink:

miklos1:
Yes, it fixes, or yes, you can check?

blechta:
Fixes...

Issue #60, opened 2016-10-03
Can't use CellVolume in interpolation


wence-:
The firedrake interpolation code calls `fem.compile_ufl` directly.  This means it doesn't have access to the `cellvolume` callback that is set up by `driver.compile_integral`.  I see comments in this region that say "TODO: refactor!" and I think @miklos1 is thinking about pulling all the interpolation kernel generation into tsfc (so that the firedrake interface doesn't have to be touched when anything changes).

We'd like to use this in thetis though for various reasons, so noting the issue here.

wence-:
@tkarna here's the relevant tsfc issue for cell volume interpolation.

miklos1:
I hope it isn't urgent, since I'm not planning to work on it this week.

wence-:
No no

PR #68, opened 2016-10-17
Refactor CellVolume and FacetArea implementation

miklos1:
- Long overdue refactoring of the implementation of `CellVolume` and `FacetArea`
- Workaround for [the UFL degree estimation bug](https://bitbucket.org/fenics-project/ufl/issues/80), so `CellVolume`/`FacetArea` should calculate the correct value even with higher-order coordinates.
- Wires up `CellVolume` in UFL interpolation (#60)
- New `gem.utils` module with useful utilities
- Some post-#63 clean up

Requires firedrakeproject/firedrake#913.

### DIFF
--- a/gem/gem.py
+++ b/gem/gem.py
@@ -354,12 +354,11 @@ class Index(IndexBase):
 
     __slots__ = ('name', 'extent', 'count')
 
-    def __init__(self, name=None):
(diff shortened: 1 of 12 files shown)


miklos1:
@wence-: will merge this soon if no objections.

PR #70, opened 2016-10-20
Optimise evaluation for cellwise constant quantities

miklos1:
This takes advantage of FInAT/FInAT#13 to optimise the evaluation of cellwise constant quantities -- eliminating one regression w.r.t. mainline TSFC.

Note this pull request is to `finat-old`, not `master`.

### DIFF
--- a/tsfc/fem.py
+++ b/tsfc/fem.py
@@ -165,7 +165,7 @@ def translate_argument(terminal, mt, ctx):
     M = ctx.entity_selector(callback, mt.restriction)
     vi = tuple(gem.Index(extent=d) for d in mt.expr.ufl_shape)
     argument_index = ctx.argument_indices[terminal.number()]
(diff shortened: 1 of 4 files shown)


tsfc/fem.py
```diff
@@ -165,7 +160,7 @@ def callback(entity_id):
     M = ctx.entity_selector(callback, mt.restriction)
     vi = tuple(gem.Index(extent=d) for d in mt.expr.ufl_shape)
     argument_index = ctx.argument_indices[terminal.number()]
-    result = gem.Indexed(M, ctx.point_multiindex + argument_index + vi)
+    result = gem.Indexed(M, argument_index + vi)
```
wence-: Removing the point indices because the tabulation tensor has lost them?

miklos1: Yes.

PR #71, opened 2016-10-21
Wire up Q/DQ elements in the FInAT interface

miklos1:
Requires FInAT/FInAT#14, avoids coneoproject/COFFEE#97.

### DIFF
--- a/tsfc/coffee.py
+++ b/tsfc/coffee.py
@@ -182,11 +182,6 @@ def statement_evaluate(leaf, parameters):
             return coffee.Block(ops, open_scope=False)
     elif isinstance(expr, gem.Constant):
         assert parameters.declare[leaf]
-        # Take all axes except the last one
-        axes = tuple(range(len(expr.array.shape) - 1))
-        nz_indices, = expr.array.any(axis=axes).nonzero()
-        nz_bounds = tuple([(i, 0)] for i in expr.array.shape[:-1])
-        nz_bounds += ([(max(nz_indices) - min(nz_indices) + 1, min(nz_indices))],)
         table = numpy.array(expr.array)
         # FFC uses one less digits for rounding than for printing
         epsilon = eval("1e-%d" % (parameters.precision - 1))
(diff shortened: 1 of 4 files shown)


tsfc/coffee.py
```diff
@@ -195,7 +190,7 @@ def statement_evaluate(leaf, parameters):
         table[abs(table + 1.0) < epsilon] = -1.0
         table[abs(table - 0.5) < epsilon] = 0.5
         table[abs(table + 0.5) < epsilon] = -0.5
-        init = coffee.SparseArrayInit(table, parameters.precision, nz_bounds)
+        init = coffee.ArrayInit(table, parameters.precision)
```
wence-: We assume this basically doesn't matter because the big zero blocks are gone in this implementation anyway.

tsfc/finatinterface.py
```diff
@@ -117,6 +130,17 @@ def convert_tensorelement(element):
...
+@convert.register(ufl.TensorProductElement)
+def convert_tensorproductelement(element):
+    cell = element.cell()
+    if type(cell) is not ufl.TensorProductCell:
+        raise ValueError("TensorProductElement not on TensorProductCell?")
+    return finat.TensorProductElement([create_element(elem)
+                                       for elem in element.sub_elements()])
+
+
+quad_opc = ufl.TensorProductCell(ufl.Cell("interval"), ufl.Cell("interval"))
```
wence-: I guess you could take this opportunity to rename this variable as `quad_tpc`?

tsfc/coffee.py
miklos1: Actually what I have on the link is something I might be able to optimise away.

miklos1: Well, [this](https://gist.github.com/miklos1/ad89ebe7bdec1b0763e36f609209f928) has more compact tables, but code actually looks worse. I think it is better to just leave it like above for now.

miklos1:
Refactoring FInAT API before merging this...

PR #76, opened 2016-10-28
Revise scalar formatting and rounding rules

### LINKED ISSUE #64: Formatting precision of floats seems to be buggy

miklos1:
Resolves #64, according to the following suggestion:
>
- Revert commit af4145ab65b68c4609377b8130597cb7d952fee1, i.e. format scalars up to precision digits, rather than precision minus one.
- On the other hand, round floating-point numbers if they are within `epsilon` (= 10 * 10^(-precision)) absolute difference to a **non-zero** integer. This kinda emulates what I think the minus one was supposed to achieve, yet doesn't destroy small (close to zero) numbers, as they could be user input ([FFC #127](https://bitbucket.org/fenics-project/ffc/issues/127)).

### DIFF
--- a/tsfc/coffee.py
+++ b/tsfc/coffee.py
@@ -41,6 +41,7 @@ def generate(impero_c, index_names, precision, roots=(), argument_indices=()):
     params.declare = impero_c.declare
     params.indices = impero_c.indices
     params.precision = precision
+    params.epsilon = 10.0 * eval("1e-%d" % precision)
     params.roots = roots
     params.argument_indices = argument_indices
 
@@ -298,7 +299,11 @@ def _expression_scalar(expr, parameters):
     if isnan(expr.value):
         return coffee.Symbol("NAN")
     else:
-        return coffee.Symbol(("%%.%dg" % (parameters.precision - 1)) % expr.value)
+        v = expr.value
+        r = round(v, 1)
+        if r and abs(v - r) < parameters.epsilon:
+            v = r  # round to nonzero
+        return coffee.Symbol(("%%.%dg" % parameters.precision) % v)
 
 
 @_expression.register(gem.Variable)


miklos1:
@blechta: This necessitates minor changes to the FFC reference data (one more digit appears in scalars). I pushed the new reference data as `d61e08e98035bac98d3e39bf1d57b6baf0644e20`. Any suggestions on the order in which following should be pushed?

- Merging this patch to blechta/tsfc@next
- Pushing the new ffc-reference-data-id to FFC/jan/tsfc-rebase-2
- Syncing blechta/tsfc/master with firedrakeproject/tsfc/master after this patch is merged

miklos1:
@blechta: I merged master into blechta/tsfc/next, then restarted the FFC tsfc next build on Bamboo, but I have seen no change.

blechta:
Reference `d61e08e98035bac98d3e39bf1d57b6baf0644e20` was based on wrong changes. Tip of master of reference data is not guaranteed to contain anything related to a tip of any FFC's branch. In fact, the tip contained changes for Martin's umerged topic branches. I've fixed that. FFC tsfc next plan was not triggeref probably because FFC next plan failed. Now everything is green and our tsfc and COFFEE masters are in sync with yours. Thanks!

blechta:
>  then restarted the FFC tsfc next build on Bamboo

I'm not sure if manual restart of FFC tsfc next does what intended. FFC tsfc next depends on FFC next building and pushing docker image and then pulls it. I think the right way is to restart FFC next. That should start FFC tsfc next as its child once it completes. Sorry that it's not documented so far. The same holds for "master+branches" plan.

Issue #75, opened 2016-10-27
FInAT transition plan


miklos1:
Outstanding tasks before switching to FInAT:
- [x] Make the whole test suite pass.
  - [x] Fix the rounding of FIAT tabulation matrices.
- [x] Optimise evaluation of quantities that are constant on the integration entity.
- [x] Code snippet efficiency for the Jacobian of affine simplices.
- [x] Structured code generation for `VectorElement`/`TensorElement`.
  - [x] Optimise away Kronecker deltas for coefficients.
- [x] Structured code generation for `TensorProductElement`.
  - [x] Sum factorisation for coefficients.
  - [x] Sum factorisation for arguments.
- [x] Update the UFC kernel interface.

Things to do last before the switch:
- [x] Clean up the `finat` branch in TSFC and FInAT itself.
- [x] Have a look at the generated code for different problems to spot potential regressions. (The FFC regression tests can help with this.)
- [x] Look for regressions in form compilation speed and the performance of the generated code.
- [x] Test correctness and performance for a few critical projects (gusto, thetis).
- [x] Test the upgrade path (should just work).

Things that can wait until after the switch:
- Structured code generation for `EnrichedElement`/`MixedElement`.
  - Sum factorisation for RTCF, RTCE, NCF, NCE elements and for the appropriate H(div) and H(curl) elements on triangular prism cells.
- Use FInAT elements in Firedrake function spaces.
- Retire `tsfc.fiatinterface` and the supporting routines.
- Optimise away Kronecker deltas for arguments.

miklos1:
Moved "optimise away Kronecker deltas for arguments" to the backlog. Since sum factorisation of arguments and argument factorisation will happen in COFFEE, this point might be difficult to do.

PR #80, opened 2016-11-16
Sum factorisation and delta elimination for coefficients

miklos1:
Ticks two boxes in #75. Can't merge currently without disabling COFFEE because of coneoproject/COFFEE#100.

### DIFF
--- a/gem/gem.py
+++ b/gem/gem.py
@@ -553,26 +553,27 @@ class ComponentTensor(Node):
 
 
 class IndexSum(Scalar):
-    __slots__ = ('children', 'index')
-    __back__ = ('index',)
+    __slots__ = ('children', 'multiindex')
+    __back__ = ('multiindex',)
 
-    def __new__(cls, summand, index):
+    def __new__(cls, summand, multiindex):
         # Sum zeros
         assert not summand.shape
         if isinstance(summand, Zero):
             return summand
 
-        # Sum a single expression
-        if index.extent == 1:
-            return Indexed(ComponentTensor(summand, (index,)), (0,))
+        # No indices case
+        multiindex = tuple(multiindex)
+        if not multiindex:
+            return summand
 
         self = super(IndexSum, cls).__new__(cls)
         self.children = (summand,)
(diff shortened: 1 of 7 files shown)


gem/optimise.py
```diff
@@ -191,6 +196,168 @@ def select_expression(expressions, index):
...
+            if flops < best_flops:
+                expression = expr
+                best_flops = flops
+    else:
+        # Cheap heuristic
+        def key(factor):
+            return len(set(sum_indices) & set(factor.free_indices))
+        ordering = sorted(factors, key=key)
+
+        # FIXME: Log this unexpected case.
```
miklos1: GEM doesn't have a logger, and I don't want to introduce reverse dependency on TSFC.

wence-: Could you fix this by requiring the caller of gem to provide a logger?

gem/gem.py
```diff
@@ -553,26 +553,27 @@ def __new__(cls, expression, multiindex):
...
         # Sum zeros
         assert not summand.shape
         if isinstance(summand, Zero):
             return summand
 
-        # Sum a single expression
-        if index.extent == 1:
-            return Indexed(ComponentTensor(summand, (index,)), (0,))
+        # No indices case
+        multiindex = tuple(multiindex)
```
miklos1: The ordering of these indices logically shouldn't matter.

gem/optimise.py
```diff
@@ -191,6 +196,175 @@ def select_expression(expressions, index):
...
+    # Flatten a product tree
+    sum_indices = []
+    factors = []
+
+    queue = deque([expression])
+    while queue:
+        expr = queue.popleft()
+        if isinstance(expr, IndexSum):
+            queue.append(expr.children[0])
+            sum_indices.extend(expr.multiindex)
```
miklos1: However, presumably here, we rely on the indices coming out in the same order on all nodes for deterministic code generation.

miklos1:
This does not work well with derivatives.

PR #81, opened 2016-11-25
Better sum factorisation for coefficients

miklos1:
This hopefully does right what #80 did not. Summary of changes:

- Improved constant folding
- Multi-index `IndexSum` nodes
- Workaround for coneoproject/COFFEE#98
- Sum factorisation and delta elimination for coefficients, with a new sum factorisation algorithm
- Revised coefficient evaluation (and marginally argument evaluation)
- Code snippet generation for fast Jacobian evaluation on affine cells.

Ticks three boxes in #75, and closes coneoproject/COFFEE#97.

### DIFF
--- a/gem/optimise.py
+++ b/gem/optimise.py
@@ -16,7 +16,7 @@ from gem.gem import (Node, Terminal, Failure, Identity, Literal, Zero,
                      Product, Sum, Comparison, Conditional, Index,
                      VariableIndex, Indexed, FlexiblyIndexed,
                      IndexSum, ComponentTensor, ListTensor, Delta,
-                     partial_indexed)
+                     partial_indexed, one)
 
 
 @singledispatch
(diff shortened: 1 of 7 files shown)


gem/optimise.py
```diff
@@ -267,3 +394,19 @@ def unroll_indexsum(expressions, max_extent):
...
+
+
+def aggressive_unroll(expression):
+    """Aggressively unrolls all loop structures."""
+    # Unroll expression shape
+    if expression.shape:
+        tensor = numpy.empty(expression.shape, dtype=object)
+        for alpha in numpy.ndindex(expression.shape):
+            tensor[alpha] = Indexed(expression, alpha)
+        expression = ListTensor(tensor)
```
miklos1: This converts a numpy array to a GEM object.

PR #86, opened 2016-12-12
Switch to FInAT

### LINKED ISSUE #75: FInAT transition plan

miklos1:
Closes #75. Requires firedrakeproject/firedrake#964.


gem/gem.py
```diff
@@ -621,6 +647,31 @@ def get_hash(self):
...
+        # Fixed indices
+        if isinstance(i, int) and isinstance(j, int):
+            return Literal(int(i == j))
+
+        self = super(Delta, cls).__new__(cls)
+        self.i = i
+        self.j = j
+        # Set up free indices
+        free_indices = tuple(index for index in (i, j) if isinstance(index, Index))
+        self.free_indices = tuple(unique(free_indices))
```
wence-: The constant-folding of `i == j` makes the uniquifying unnecessary?  Unless indices hash differently from equality?

wence-: Oh, you care about ordering, making `Delta(j, i)` and `Delta(i, j)` have the same ordering of `free_indices` by calling unique here?

miklos1: This has evolved from `numpy.unique` which sorts **and** removes duplicates. This idea is to make `x.free_indices == y.free_indices` work (without having to cast to sets first).

gem/optimise.py
```diff
@@ -73,11 +129,234 @@ def filtered_replace_indices(node, self, subst):
...
+                return Zero(shape)
+        return Indexed(_select_expression(list(map(child, expressions)), index), multiindex)
+
+    if types <= {Literal, Zero, Failure}:
+        return partial_indexed(ListTensor(expressions), (index,))
+
+    if len(types) == 1:
+        cls, = types
+        if cls.__front__ or cls.__back__:
+            raise NotImplementedError
```
wence-: Descriptive error message?

gem/optimise.py
```diff
@@ -73,11 +129,234 @@ def filtered_replace_indices(node, self, subst):
...
+        delta_queue = [(f, index)
+                       for f in factors if isinstance(f, Delta)
+                       for index in (f.i, f.j) if index in sum_indices]
+
+    # Drop ones
+    return sum_indices, [e for e in factors if e != one]
+
+
+def sum_factorise(sum_indices, factors):
+    """Optimise a tensor product throw sum factorisation.
```
wence-: through?

gem/optimise.py
```diff
@@ -73,11 +129,234 @@ def filtered_replace_indices(node, self, subst):
...
+
+    :arg sum_indices: free indices for contractions
+    :arg factors: product factors
+    :returns: optimised GEM expression
+    """
+    if len(sum_indices) > 5:
+        raise NotImplementedError("Too many indices for sum factorisation!")
```
wence-: I guess if we care, we can probably figure out a way to cast this as a shortest path problem.

tsfc/finatinterface.py
```diff
@@ -0,0 +1,155 @@
...
+                                      element.degree())
+        return finat.QuadrilateralElement(create_element(element))
+    return lmbda(cell, element.degree())
+
+
+# VectorElement case
+@convert.register(ufl.VectorElement)
+def convert_vectorelement(element):
+    scalar_element = create_element(element.sub_elements()[0])
+    return finat.TensorFiniteElement(scalar_element, (element.num_sub_elements(),))
```
wence-: Presumably `element.reference_value_shape()` would work in this case too?

miklos1: No, because if IIRC `ufl.TensorElement` asserts that the inner element is scalar, while `ufl.VectorElement` gladly accepts an RT element for example.

wence-: Oh yeah

gem/optimise.py
wence-: OK, so we don't care.  What I meant was, you can probably recast this minimisation over permutations as some kind of graph problem...

tsfc/geometric.py
```diff
@@ -98,13 +98,25 @@ def translate_cell_edge_vectors(terminal, mt, params):
...
-                                                     mt.restriction),
-                               (params.point_index,))
+    ps = params.point_set
+    if params.integration_dim == params.fiat_cell.get_dimension():
+        return ps.expression
+
+    # This destroys the structure of the quadrature points, but since
+    # this code path is only used to implement CellCoordinate in facet
+    # integrals, hopefully it does not matter much.
+    point_shape = tuple(index.extent for index in ps.indices)
```
wence-: Can we think of a test case where doing so would mean that evaluating `CellCoordinate` would become the leading-order cost?

gem/optimise.py
miklos1: doi:10.1109/18.910572 does something like that, but that approach only yields an easy optimum if the factor graph is cycle-free which isn't the case for us. Cycle-breaking techniques can produce correct factorisations in this case, with no guarantee about optimality, so it's not really helpful for us.

tsfc/geometric.py
miklos1: I don't think so. Leading order terms contain both basis function indices and quadrature indices; this one has no basis functions, so it involves quadrature indices only.

gem/optimise.py
wence-: OK.

tsfc/finatinterface.py
```diff
@@ -0,0 +1,155 @@
...
+    "Nedelec 1st kind H(curl)": finat.Nedelec,
+    "Nedelec 2nd kind H(curl)": finat.NedelecSecondKind,
+    "Raviart-Thomas": finat.RaviartThomas,
+    "Regge": finat.Regge,
+    # These require special treatment below
+    "DQ": None,
+    "Q": None,
+}
+"""A :class:`.dict` mapping UFL element family names to their
+FIAT-equivalent constructors.  If the value is ``None``, the UFL
```
wence-: finat?

tsfc/finatinterface.py
```diff
@@ -0,0 +1,155 @@
...
+    return finat.TensorProductElement([create_element(elem)
+                                       for elem in element.sub_elements()])
+
+
+quad_tpc = ufl.TensorProductCell(ufl.interval, ufl.interval)
+_cache = weakref.WeakKeyDictionary()
+
+
+def create_element(element):
+    """Create a FIAT element (suitable for tabulating with) given a UFL element.
```
wence-: finat?

PR #94, opened 2017-01-27
Fix performance bug in code generation

miklos1:
Figuring out where to declare temporaries may take O(n^2) time, n = # of temporaries. The patch fixes this performance bug.

### DIFF
--- a/gem/impero_utils.py
+++ b/gem/impero_utils.py
@@ -194,15 +194,15 @@ def place_declarations(tree, temporaries, get_indices):
     :arg get_indices: callable mapping from GEM nodes to an ordering
                       of free indices
     """
-    temporaries_set = set(temporaries)
-    assert len(temporaries_set) == len(temporaries)
+    numbering = {t: n for n, t in enumerate(temporaries)}
+    assert len(numbering) == len(temporaries)
 
     # Collect the total number of temporary references
     total_refcount = collections.Counter()
     for node in traversal((tree,)):
         if isinstance(node, imp.Terminal):
-            total_refcount.update(temp_refcount(temporaries_set, node))
-    assert temporaries_set == set(total_refcount)
+            total_refcount.update(temp_refcount(numbering, node))
+    assert set(total_refcount) == set(temporaries)
 
     # Result
     declare = {}
@@ -223,7 +223,7 @@ def place_declarations(tree, temporaries, get_indices):
 
     @recurse.register(imp.Terminal)
     def recurse_terminal(expr, loop_indices):
-        return temp_refcount(temporaries_set, expr)
+        return temp_refcount(numbering, expr)
 
     @recurse.register(imp.For)
     def recurse_for(expr, loop_indices):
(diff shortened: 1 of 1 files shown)


gem/impero_utils.py
```diff
@@ -241,7 +241,7 @@ def recurse_block(expr, loop_indices):
             refcount.update(recurse(statement, loop_indices))
 
         # Visit :class:`collections.Counter` in deterministic order
-        for e in sorted(refcount.keys(), key=temporaries.index):
```
miklos1: `temporaries` is a `list`, `list.index` is linear search, which is called to get the key of every temporary.

gem/impero_utils.py
```diff
@@ -194,15 +194,15 @@ def place_declarations(tree, temporaries, get_indices):
     :arg get_indices: callable mapping from GEM nodes to an ordering
                       of free indices
     """
-    temporaries_set = set(temporaries)
-    assert len(temporaries_set) == len(temporaries)
+    numbering = {t: n for n, t in enumerate(temporaries)}
```
miklos1: For deterministic code generation: create a `dict` that maps temporaries to their ordering...

gem/impero_utils.py
```diff
@@ -241,7 +241,7 @@ def recurse_block(expr, loop_indices):
             refcount.update(recurse(statement, loop_indices))
 
         # Visit :class:`collections.Counter` in deterministic order
-        for e in sorted(refcount.keys(), key=temporaries.index):
+        for e in refcount.keys():
```
wence-: So here you used to sort the keys which temporaries.index, but that comparison key is `O(n)` since index must traverse the list of temporaries.

gem/impero_utils.py
```diff
@@ -256,6 +256,7 @@ def recurse_block(expr, loop_indices):
                 # Remove expression from the ``refcount`` so it will
                 # not be declared again.
                 del refcount[e]
+        declare[expr].sort(key=lambda t: numbering[t])
```
miklos1: ... and use the dict for sorting declarations in blocks.

gem/impero_utils.py
miklos1: Yes.

gem/impero_utils.py
wence-: Now you just shove stuff into the list of declarations in any old order and at the end sort with semantically the same as `temporaries.index`, but with an `O(1)` comparator.

gem/impero_utils.py
```diff
@@ -223,7 +223,7 @@ def recurse(expr, loop_indices):
 
     @recurse.register(imp.Terminal)
     def recurse_terminal(expr, loop_indices):
-        return temp_refcount(temporaries_set, expr)
+        return temp_refcount(numbering, expr)
```
wence-: Update docstring of `temp_refcount`?

miklos1: Meh, only `in` is used in `temp_refcount`, fast lookup works just as well on `dict`s as on `set`s.

PR #96, opened 2017-02-03
Fix pickling of GEM

### LINKED ISSUE #74: Make GEM expressions picklable
This currently doesn't work because of the presence of `__slots__` and the lack of `__getstate__` and `__setstate__`. Moreover, `gem.node.Node` defines a method named `__getinitargs__` with a different purpose than `pickle` expects such a method to work if exists.

Pickling GEM would be handy for experimenting with transformations on GEM expressions. One could simply write a UFL form, call `compile_form` on it, set a breakpoint just after the UFL is translated to GEM, dump the GEM expression to file, and then load it in a Jupyter notebook for experimentation. In similar manner, the semi-manually transformed GEM expression can be fed back to the compilation pipeline to see and test the result as C code.

miklos1:
Found a hidden detail in the Python documentation which reveals that pickling in Python need not be a messy chaos.

Testing and saving/restoring running indices is not immaculate yet, but workable. Fixing little annoyances will be possible when we have better textual representation (pretty-printing) tools.

Closes #74.

### DIFF
--- a/tests/test_pickle_gem.py
+++ b/tests/test_pickle_gem.py
@@ -1,0 +1,24 @@
+from __future__ import absolute_import, print_function, division
+from six.moves import cPickle as pickle, range
+import gem
+import numpy
+import pytest
+
+
+@pytest.mark.parametrize('protocol', range(3))
+def test_pickle_gem(protocol):
+    f = gem.VariableIndex(gem.Indexed(gem.Variable('facet', (2,)), (1,)))
+    q = gem.Index()
+    r = gem.Index()
+    _1 = gem.Indexed(gem.Literal(numpy.random.rand(3, 6, 8)), (f, q, r))
+    _2 = gem.Indexed(gem.view(gem.Variable('w', (None, None)), slice(8), slice(1)), (r, 0))
+    expr = gem.ComponentTensor(gem.IndexSum(gem.Product(_1, _2), (r,)), (q,))
+
+    unpickled = pickle.loads(pickle.dumps(expr, protocol))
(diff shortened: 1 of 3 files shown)


tests/test_pickle_gem.py
```diff
@@ -0,0 +1,24 @@
...
+    f = gem.VariableIndex(gem.Indexed(gem.Variable('facet', (2,)), (1,)))
+    q = gem.Index()
+    r = gem.Index()
+    _1 = gem.Indexed(gem.Literal(numpy.random.rand(3, 6, 8)), (f, q, r))
+    _2 = gem.Indexed(gem.view(gem.Variable('w', (None, None)), slice(8), slice(1)), (r, 0))
+    expr = gem.ComponentTensor(gem.IndexSum(gem.Product(_1, _2), (r,)), (q,))
+
+    unpickled = pickle.loads(pickle.dumps(expr, protocol))
+    assert repr(expr) == repr(unpickled)
+
```
thomasgibson: Are you comparing `repr(expr) == repr(unpickled)` here because there's no guarantee that two objects comparing equal with `==` serialize properly?

miklos1: They won't be equal and I would argue they shouldn't, and this is because of the semantics of running indices. Two running indices are the same only by object identity, so one cannot accidentally create an Index that is equal to an already existing Index... that is only possible by copying a reference. This object identity however infeasible to maintain through the pickling/unpickling operation.

One might also argue that `==` is overly discriminating, and that is true, but changing that would require much more complicated comparison algorithms.

PR #97, opened 2017-02-09
Attempt to do form splitting in TSFC for loo.py

miklos1:
Note that the target is `loopy-output`, not `master`.

This is quite a rude sketch to do form splitting and fused kernels inside TSFC. It doesn't work with Firedrake master since it changes the kernel API: due to fusion, kernels from mixed forms expect several arguments for return value (e.g. A0, A1, A2, A3 the four blocks of a mixed form).

@rckirby: please have a look and see if you can make use of this!

### DIFF
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -106,8 +106,9 @@ def compile_integral(integral_data, form_data, prefix, parameters,
     fake_args = [[Argument(FunctionSpace(arg.ufl_domain(), sub_elem), arg.number())
                   for sub_elem in ufl_utils.unmix_element(arg.ufl_element())]
                  for arg in arguments]
-    for bar, args in zip(itertools.product(*argument_indices), itertools.product(*fake_args)):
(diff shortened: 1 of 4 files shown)


rckirby:
So far I can confirm that this generates code that looks like what I expected for cell integrals on mixed spaces and that tsfc_to_loopy generates a loopy kernel without failure.  No attempt yet to process the loopy kernel.

rckirby:
We now have a collection of output tensor "Ai" for i=0, 1, ...
Will these always correspond to the pieces of the test functions?
Or are there cases (e.g. cell integrals) where we might have len(test_space.split()) not equal to the number of output Ai's?

wence-:
It will always, I think, be `len(test_space.split()) * len(trial_space.split())`

miklos1:
@rckirby: I just pushed here a proposed fix for your recent issue.

rckirby:
I've made the last changes I think this needs to provide for generating loopy code.  I think this can be merged back into loopy-output now.

PR #98, opened 2017-02-19
Coffee migration

tj-sun:
This commit migrates loop optimisations in coffee to gem.

### Test if functionalities are migrated

Results of flops comparison are in the link below

The result shows that gem.optimise() matches or betters coffee level 2 by and large. It also demonstrates that coffee level 2 cannot further optimise ASTs generated by gem.optimise().

There are some cases for which gem.optimise() are slightly worse. One example is mixed_poisson, dim=2, p=1, q=1, nf=0.

[tsfc output](https://gist.github.com/tj-sun/3efbc3aaff764fb8e74ed417d8838cad)
[coffee o2](https://gist.github.com/tj-sun/011d43e6ca47c62f634530a9b5820c26)
[this PR](https://gist.github.com/tj-sun/e5572e137d9aff2a68735a88030b7f82)

Notice COFFEE can merge temporaries t24 and t23. Such optimisation is not implemented here yet.

Compile time is generally OK, e.g. holzapfel takes 4.02s, vs tsfc 0.91s, COFFEE O2 13.80s.

### Minor problems:

- Several subtle cases to refine further, e.g. Power nodes are not analysed yet (might be beneficial to expand the product).
- latex printing not completed yet.

### TODOs:
- factorise non argument factors in the end (this will likely require expanding the products in `rest` which is probably expensive, maybe this is another optimisation level)

[Results link](https://drive.google.com/open?id=1Cl81qNyeZFJnf8hf-Wz48XudECzkoR5El8Ax6rTREdA)

PR #99, opened 2017-02-20
Coffee optimisation migration

tj-sun:
#### clean commit for PR to squash numerous commits and separate latex printing changes

This commit migrates loop optimisations in coffee to gem.

### Test if functionalities are migrated

Results of flops comparison are in the attached file result.txt (actually it is a csv file, which cannot be uploaded).

The result shows that gem.optimise() matches or betters coffee level 2 by and large. It also demonstrates that coffee level 2 cannot further optimise ASTs generated by gem.optimise(). There are some cases for which gem.optimise() are slightly worse. These can be solved by checking if factorisation is profitable at each step. As it tends to only affect simple forms with small number of flops, and the differences are not large, I have commented out the code for now. Will revisit this later.

### Major problem
pre-evaluations of tensors reduces numerical precision in some cases, as a result, these tests fails in the firedrake test suits

regression/test_geometric_strong_bcs.py::test_dg_advection[left-1-False]
regression/test_hybridisation.py::test_hybridisation[2]
regression/test_manifolds.py::test_manifold_parallel
regression/test_poisson_strong_bcs.py
regression/test_poisson_strong_bcs_nitsche.py::test_poisson_nitsche[1-False]
regression/test_python_parloop.py::test_python_parloop_vector
regression/test_slate_hybridization.py::test_slate_hybridization[2]

In addition, several others which use errornorm() fails due to taking square root of (tiny) negative numbers. Not sure if these can be eliminated by some clever tricks in choosing an order to contract the tensors, or purely bad luck with floating point numbers.

### Minor problems:

- Performance can be improved further. Currently 3D hyperelasticity takes ~20 seconds. There are significant scope for code optimisation to get that time down, such as reusing the internal data structure of list of factors.
- Several subtle cases to refine further, e.g. Power nodes are not analysed yet (might be beneficial to expand the product).
- Pre-evaluation only works when the inner IndexSum is expanded already.

[result.txt](https://github.com/firedrakeproject/tsfc/files/787955/result.txt)

### DIFF
--- a/gem/gem.py
+++ b/gem/gem.py
@@ -34,7 +34,8 @@ __all__ = ['Node', 'Identity', 'Literal', 'Zero', 'Failure',
            'LogicalNot', 'LogicalAnd', 'LogicalOr', 'Conditional',
            'Index', 'VariableIndex', 'Indexed', 'ComponentTensor',
            'IndexSum', 'ListTensor', 'Delta', 'index_sum',
-           'partial_indexed', 'reshape', 'view']
+           'partial_indexed', 'reshape', 'view', 'Scalar', 'Terminal',
+           'FlexiblyIndexed']
 
 
 class NodeMeta(type):
@@ -149,10 +150,11 @@ class Identity(Constant):
 class Literal(Constant):
     """Tensor-valued constant"""
 
-    __slots__ = ('array',)
+    __slots__ = ('array', 'name')
     __front__ = ('array',)
+    __back__ = ('name',)
 
-    def __new__(cls, array):
+    def __new__(cls, array, name=None):
         array = asarray(array)
         if (array == 0).all():
             # All zeros, make symbolic zero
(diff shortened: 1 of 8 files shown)


miklos1:
@tj-sun: Fix pull request title, please!

miklos1:
Ugh... not sure if new PR is a good idea: we have lost all pull request comment history...

wence-:
Comments applied to the previous PR still apply here, FWIW.

As to the error norm issue.  The problem is that we compute `sqrt(assemble((a-b)**2*dx))` which is guaranteed to be positive (or zero), but not in floating point if we expand into `(a^2 + b^2 - 2ab)`.  Dolfin's implementation of error norm has a workaround for this which is really ugly (interpolate both `a` and `b` into some (higher degree) space that spans the target space.  Then evaluate the pointwise difference, then compute the norm of this.

This trades off an interpolation error at quadrature points of the exact expression for an interpolation error at the node points of the (higher degree) space.

When one is computing the error norm, on paper, we don't mean pointwise differences.  So that's not very satisfactory.  @colinjcotter thoughts?  my approximation theory is pretty weak.

If we could turn off optimisation passes on a per-integral basis, then we can do so for the error norm computations in firedrake and this problem would go away as long as such products are not expanded.  Or we could slap an `abs` round the calculation in Firedrake, since that's "mathematically" safe?

miklos1:
We are staying with #98.

PR #101, opened 2017-03-15
Add GEM utilities for refactorisation

miklos1:
This is the common part of two feature branches:

- `argument-sketch`: applies argument factorisation and sum factorisation for arguments at the GEM level, rather than leaving it for COFFEE.
- `tensor`: Similar to the tensor representation of FFC, attempts to rearrange the GEM expression such that the quadrature loop is constant folded at compile time.

Refactorisation is parametrized by a _classifier_ function which classifies any GEM subexpression as

- `atomic`: an expression of particular interest that needs no further decomposition
- `compound`: an expression "containing" multiple atomics, application of distributivity rules (expansion) is necessary
- `other`: an expression of no particular interest for refactorisation

For example, for argument factorisation atomics have exactly one argument index as free index, compounds have more than one, and others have no argument indices. For the "tensor representation" atomics are literal tables with a quadrature index, compounds are quadrature index-dependent expressions, and others have no dependency on any quadrature index.

Additionally, this pull request optimises associativity during sum factorisation.

Includes a small test case.

### DIFF
--- a/gem/optimise.py
+++ b/gem/optimise.py
@@ -2,12 +2,12 @@
 expressions."""
 
 from __future__ import absolute_import, print_function, division
-from six import itervalues
+from six import iteritems, itervalues
 from six.moves import filter, intern, map, zip
 
-from collections import OrderedDict, namedtuple
+from collections import OrderedDict, defaultdict, namedtuple
 from functools import reduce
-from itertools import chain, combinations, permutations, product
+from itertools import combinations, permutations, product
 
 import numpy
 from singledispatch import singledispatch
@@ -409,8 +409,8 @@ OTHER = intern('other')
 
 
 Monomial = namedtuple('Monomial', ['sum_indices', 'atomics', 'rest'])
-"""Monomial type, used in the return type of
-:py:func:`collect_monomials`.
+"""Monomial type, representation of a tensor product with some
+distinguished factors (called atomics).
 
 - sum_indices: indices to sum over
 - atomics: tuple of expressions classified as ATOMIC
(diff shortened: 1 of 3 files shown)


miklos1:
@tj-sun: I wonder if 500a829 was really a good idea...

gem/optimise.py
```diff
@@ -230,13 +230,47 @@ def delta_elimination(sum_indices, factors):
     return sum_indices, [e for e in factors if e != one]
 
 
+def associate_product(factors):
+    """Apply associativity rules to construct an operation-minimal product tree.
+
+    For best performance give factors that have different set of free indices.
+    """
```
tj-sun: "Create product which has best performance (i.e. lowest number of flops) from a given list of factors with (potentially) different set of free indices" ?

gem/optimise.py
```diff
@@ -230,13 +230,47 @@ def delta_elimination(sum_indices, factors):
     return sum_indices, [e for e in factors if e != one]
 
 
+def associate_product(factors):
```
tj-sun: Also wondering if we will need to reassociate summation as well?

gem/refactorise.py
```diff
@@ -0,0 +1,232 @@
...
+            for key, value in iteritems(arg.ordering):
+                result.ordering.setdefault(key, value)
+        return result
+
+    @staticmethod
+    def product(*args):
+        """Product of multiple :py:class:`MonomialSum`s"""
+        result = MonomialSum()
+        for monomials in product(*args):
+            sum_indices = []
```
tj-sun: similar assertion as per sum()?

gem/refactorise.py
```diff
@@ -0,0 +1,232 @@
...
+        # Break up compounds only
+        return self.classifier(expr) == OTHER
+    common_indices, terms = traverse_product(expression, stop_at=stop_at)
+    common_indices = tuple(common_indices)
+
+    common_atomics = []
+    common_others = []
+    compounds = []
+    for term in terms:
+        cls = self.classifier(term)
```
tj-sun: cls not really a keyword but maybe should avoid using it?

gem/optimise.py
miklos1: Probably, but there is no obvious place to do that, so I would just make a separate, DAG-safe pass for that.

gem/refactorise.py
miklos1: Will change...

gem/optimise.py
miklos1: > For best performance give factors that have different set of free indices.

What this really means is that if you have multiple factors with the same set of free indices, then you are wasting optimisation time.

miklos1:
@tj-sun: I keep 500a829 for now because otherwise I get into trouble on `argument-sketch` when not expanding short loops.

PR #102, opened 2017-03-24
Mode infrastructure and multiple little changes

miklos1:
This is really a bag of changes, so I discuss them one by one:

- Renames: `argument_indices` -> `argument_multiindices` and `flat_argument_indices` -> `argument_indices`. This should have happened during the FInAT transition, but it did not, so better later than never. @tj-sun: you are using and propagating the old names, please understand the changes and update your branch accordingly.
- Index cache: another loss of the FInAT transition. On current master, the index cache exists in bits and pieces but it is never actually used. This change puts the index cache back into use.
- Internal API of `gem.impero_utils.compile_gem` changes.
- Per discussion with @blechta, per-integral metadata now has precedence over the `parameters=` keyword argument provided to `compile_form`/`compile_integral` (and also over the defaults). This will now match FFC's behaviour. @wence-: We miss this early history, but I think that the current master behaviour was your choice. Any particular reason? Everything still seems to pass with this change.
- Some basic infrastructure is set up for modes, more details below.

### What are modes?

Modes are different ways of transforming the GEM expressions that are the intermediate representation of TSFC. In some way, they are similar to representations in FFC. However, in FFC each representation is a distinct UFL -> C++ pipeline, while in TSFC there is a single UFL -> GEM pipeline and a single GEM -> COFFEE AST -> C pipeline, and modes are GEM -> GEM transformers that are inserted in the middle.

### Why do we need them?

The golden age of "just use COFFEE" is over. While current master relies on COFFEE for sum factorisation of arguments, it demonstrably fails to get the optimal algorithmic complexity on bilinear forms. This fuels the need to maintain the `argument-sketch` branch, yet that functionality is kept off master because it is on a branch. In addition, @tj-sun is working on porting the expression optimisations of COFFEE to GEM, so we will need a switch to turn his code on and off.

### What modes are expected in the near future?

- `vanilla`: This PR names the behaviour of current master *vanilla* mode: ["simplest version of something, without any optional extras, basic or ordinary."](https://en.wikipedia.org/wiki/Plain_vanilla)
- `coffee`: @tj-sun's work may be wired in as *coffee* mode.
- `spectral`: My `argument-sketch` branch could be brought into master for reliable sum factorisation, for @alsgregory (firedrakeproject/firedrake#972), and for further research on delta elimination in arguments.
- `tensor`: Based on the `tensor` branch which I made as a toy exercise in the winter break. Similar in spirit to the tensor representation of FFC, but rather than trying to reduce the whole expression to a single contraction between a reference tensor and a geometry tensor, it simply eliminates all quadrature loops by applying the contractions at compile time. The main difference is that the expression can be arbitrarily complex in quantities that are cell-wise constant, because this mode does not munge with those subexpressions, while the tensor representation would simply fail. (In particular, the `tensor` branch only fails 5 additional demos in the FFC regression test suite.)

All names are, of course, subject to discussion.

### Why are these modes not called representations like in FFC?

1. They are different under the hood, see "What are modes?" above.
2. So that they may be used from FFC. For example, compiling `u*v*dx(metadata={'representation': 'tensor'})` with FFC uses the FFC's tensor representation, while `u*v*dx(metadata={'representation': 'tsfc', 'mode': 'tensor'})` would use TSFC's tensor mode.

### DIFF
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -90,17 +90,16 @@ def compile_integral(integral_data, form_data, prefix, parameters,
     fiat_cell = as_fiat_cell(cell)
     integration_dim, entity_ids = lower_integral_type(fiat_cell, integral_type)
 
-    argument_indices = tuple(tuple(gem.Index(extent=e)
-                                   for e in create_element(arg.ufl_element()).index_shape)
-                             for arg in arguments)
-    flat_argument_indices = tuple(chain(*argument_indices))
+    argument_multiindices = tuple(create_element(arg.ufl_element()).get_indices()
+                                  for arg in arguments)
+    argument_indices = tuple(chain(*argument_multiindices))
     quadrature_indices = []
 
     # Dict mapping domains to index in original_form.ufl_domains()
     domain_numbering = form_data.original_form.domain_numbering()
     builder = interface.KernelBuilder(integral_type, integral_data.subdomain_id,
                                       domain_numbering[integral_data.domain])
-    return_variables = builder.set_arguments(arguments, argument_indices)
+    return_variables = builder.set_arguments(arguments, argument_multiindices)
 
     coordinates = ufl_utils.coordinate_coefficient(mesh)
     builder.set_coordinates(coordinates)
@@ -119,7 +118,7 @@ def compile_integral(integral_data, form_data, prefix, parameters,
                       precision=parameters["precision"],
                       integration_dim=integration_dim,
                       entity_ids=entity_ids,
-                      argument_indices=argument_indices,
+                      argument_multiindices=argument_multiindices,
                       index_cache=index_cache)
 
     kernel_cfg["facetarea"] = facetarea_generator(mesh, coordinates, kernel_cfg, integral_type)
@@ -153,7 +152,7 @@ def compile_integral(integral_data, form_data, prefix, parameters,
                              type(quad_rule))
 
         quadrature_multiindex = quad_rule.point_set.indices
-        quadrature_indices += quadrature_multiindex
+        quadrature_indices.extend(quadrature_multiindex)
 
         config = kernel_cfg.copy()
         config.update(quadrature_rule=quad_rule)
@@ -176,12 +175,12 @@ def compile_integral(integral_data, form_data, prefix, parameters,
         builder.require_cell_orientations()
 
     impero_c = impero_utils.compile_gem(return_variables, ir,
-                                        tuple(quadrature_indices) + flat_argument_indices,
+                                        tuple(quadrature_indices) + argument_indices,
                                         remove_zeros=True)
 
     # Generate COFFEE
     index_names = [(si, name + str(n))
-                   for index, name in zip(argument_indices, ['j', 'k'])
+                   for index, name in zip(argument_multiindices, ['j', 'k'])
                    for n, si in enumerate(index)]
     if len(quadrature_indices) == 1:
         index_names.append((quadrature_indices[0], 'ip'))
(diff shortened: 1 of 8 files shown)


wence-:
> Per discussion with @blechta, per-integral metadata now has precedence over the parameters= keyword argument provided to compile_form/compile_integral (and also over the defaults). This will now match FFC's behaviour. @wence-: We miss this early history, but I think that the current master behaviour was your choice. Any particular reason? Everything still seems to pass with this change.

I'd be happy to change this part.  I don't recall why I overrode per-integral metadata with parameters.  Maybe I just brain-farted.

tj-sun:
Thanks, will merge this into my branch

tsfc/driver.py
```diff
@@ -90,50 +89,47 @@ def compile_integral(integral_data, form_data, prefix, parameters,
...
-        params.update(integral.metadata())
+        params = parameters.copy()
+        params.update(integral.metadata())  # integral metadata overrides
         if params.get("quadrature_rule") == "default":
             del params["quadrature_rule"]
-        # parameters override per-integral metadata
-        params.update(parameters)
+
+        mode = pick_mode(params["mode"])
+        mode_irs.setdefault(mode, collections.OrderedDict())
```
wence-: OK, so there can be a different translation mode for every integral.

tsfc/driver.py
```diff
@@ -153,43 +149,52 @@ def compile_integral(integral_data, form_data, prefix, parameters,
...
-        ir = [gem.index_sum(expr, quadrature_multiindex) for expr in ir]
-        irs.append(ir)
-
-    # Sum the expressions that are part of the same restriction
-    ir = list(reduce(gem.Sum, e, gem.Zero()) for e in zip(*irs))
+        expressions = fem.compile_ufl(integrand,
+                                      interior_facet=interior_facet,
+                                      **config)
+        reps = mode.Integrals(expressions, quadrature_multiindex,
+                              argument_multiindices, params)
```
wence-: And it's the mode's job to translate the gem into gem here.

tsfc/driver.py
```diff
@@ -349,3 +354,12 @@ def lower_integral_type(fiat_cell, integral_type):
         entity_ids = list(range(len(fiat_cell.get_topology()[integration_dim])))
 
     return integration_dim, entity_ids
+
+
+def pick_mode(mode):
+    "Return one of the specialized optimisation modules from a mode string."
+    if mode == "vanilla":
+        import tsfc.vanilla as m
```
wence-: How do you (or do you?) envisage composing modes?  For example, it may well be beneficial for "optimised" modes to unroll index sums too.  Would they themselves import the vanilla mode and use it?

tsfc/driver.py
miklos1: It does not have to be GEM to GEM translation at this point yet, it only has to be GEM to GEM translation with `mode.Integrals` + `mode.flatten` combined. Modes may choose to do a GEM to GEM translation here (like `vanilla` here) and do a trivial concatenation later, or they may construct their representation here (perhaps just packing up these arguments here into an object), and then do the optimisation later at `mode.flatten` when all integrals are available. It is up to the mode to decide how to divide the work between `mode.Integrals` and `mode.flatten`.

tsfc/driver.py
miklos1: I do not envisage composing modes as such, strictly speaking they are merely meant for configuration. It does make sense to compose the underlying transformations, however. Unrolling `IndexSum`s is not really implemented by the vanilla mode, but by `gem.optimise.unroll_indexsum`. Any mode can just call this function if needed.

miklos1:
Augmented the description:

> ### Why are these modes not called representations like in FFC?
> 
> 1. They are different under the hood, see "What are modes?" above.
> 2. So that they may be used from FFC. For example, compiling `u*v*dx(metadata={'representation': 'tensor'})` with FFC uses the FFC's tensor representation, while `u*v*dx(metadata={'representation': 'tsfc', 'mode': 'tensor'})` would use TSFC's tensor mode.

Issue #103, opened 2017-03-30
Recent mode changes broke thetis


wence-:
7fcba4d changed something in some subtle way such that coffee's optimisations are invalid for a huge swathe of thetis tests.  e.g. run py.test test/tracerEq/test_v-diffusion_mes.py in the thetis repo.

We can workaround for now by winding back.   @tj-sun what stage are we at in migration?

tj-sun:
It's basically there (*) but need code review.  To use it we'd need to switch coffee to O0 (or Ov after coffee pr is merged).

(*) I'm doing argument factorisation unconditionally, since last week I been working on some alternatives, though I haven't seen improvements on my examples so far.

miklos1:
If you replace the inner for loop in vanilla.flatten with yield (variable, reduce(gem.Sum, expressions)), does that help?

(Sorry I'm on phone.)

wence-:
Seems to, thanks, was trying to figure out where to put that change in.

wence-:
Fixed in #104.

PR #104, opened 2017-03-30
Fix for #103

wence-:
Provide a Sum back from the list of expressions in vanilla.flatten.


miklos1:
You forgot to import reduce (Python 3).

PR #105, opened 2017-04-06
Spectral mode

miklos1:
Unconditionally argument factorises and applies sum factorisation for argument evaluation. **Not compatible with COFFEE.** Default mode remains `vanilla`.

Advantages:
- Correct algorithmic complexity for sum factorisation, protected by tests.
- @alsgregory used this branch because of firedrakeproject/firedrake#972. After merging this, he will have access to good behaviour on master with the form compiler option `{'mode': 'spectral'}`.

This is still work in progress, but it will not hurt because not enabled by default, but it offers an option that delivers the above goods.

### DIFF
--- a/gem/optimise.py
+++ b/gem/optimise.py
@@ -226,8 +226,7 @@ def delta_elimination(sum_indices, factors):
                        for f in factors if isinstance(f, Delta)
                        for index in (f.i, f.j) if index in sum_indices]
 
-    # Drop ones
-    return sum_indices, [e for e in factors if e != one]
+    return sum_indices, factors
 
 
 def associate_product(factors):
--- a/tsfc/spectral.py
+++ b/tsfc/spectral.py
@@ -1,0 +1,49 @@
+from __future__ import absolute_import, print_function, division
+from six.moves import map, zip
+
+from functools import partial
+from itertools import chain
+
+from gem import index_sum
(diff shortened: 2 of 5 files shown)


gem/optimise.py
```diff
@@ -226,8 +226,7 @@ def delta_elimination(sum_indices, factors):
                        for f in factors if isinstance(f, Delta)
                        for index in (f.i, f.j) if index in sum_indices]
 
-    # Drop ones
-    return sum_indices, [e for e in factors if e != one]
+    return sum_indices, factors
```
miklos1: The result of `delta_elimination` is only used for `sum_factorise` which won't be negatively affected by extra one factors.

tsfc/spectral.py
```diff
@@ -0,0 +1,77 @@
...
+from gem.refactorise import ATOMIC, COMPOUND, OTHER, MonomialSum, collect_monomials
+
+
+def delta_elimination(sum_indices, args, rest):
+    """IndexSum-Delta cancellation for monomials."""
+    factors = [rest] + list(args)  # construct factors
+    sum_indices, factors = _delta_elimination(sum_indices, factors)
+    # Destructure factors after cancellation
+    rest = factors.pop(0)
+    args = factors
```
miklos1: Not dropping one factors, however, allows to easily destructure here.

PR #106, opened 2017-04-07
Improve FEniCS compatibility

miklos1:
Adds:
- Vertex integrals (UFC only)
- Quadrature element (FFC and FInAT)
- Restricted element

@martinal, @dorugeber: what is the difference between `RestrictedElement` vs `FacetElement` and `InteriorElement`? They all map to the same FIAT element, feels like a duplicity in UFL.

Requires FInAT/FInAT#25.

### DIFF
--- a/tsfc/fiatinterface.py
+++ b/tsfc/fiatinterface.py
@@ -25,6 +25,7 @@ from __future__ import absolute_import, print_function, division
 
 from singledispatch import singledispatch
 from functools import partial
+import types
 import weakref
 
 import FIAT
@@ -175,7 +176,12 @@ def _(element, vector_is_mixed):
     if element.family() == "Quadrature":
         # Sneaky import from FFC
         from ffc.quadratureelement import QuadratureElement
-        return QuadratureElement(element)
+        ffc_element = QuadratureElement(element)
+
+        def tabulate(self, order, points, entity):
+            return QuadratureElement.tabulate(self, order, points)
+        ffc_element.tabulate = types.MethodType(tabulate, ffc_element)
+        return ffc_element
     cell = as_fiat_cell(element.cell())
     lmbda = supported_elements[element.family()]
     if lmbda is None:
--- a/tsfc/fiatinterface.py
+++ b/tsfc/fiatinterface.py
@@ -172,6 +172,10 @@ def _(element, vector_is_mixed):
         # Real element is just DG0
         cell = element.cell()
         return create_element(ufl.FiniteElement("DG", cell, 0), vector_is_mixed)
+    if element.family() == "Quadrature":
+        # Sneaky import from FFC
+        from ffc.quadratureelement import QuadratureElement
+        return QuadratureElement(element)
     cell = as_fiat_cell(element.cell())
     lmbda = supported_elements[element.family()]
     if lmbda is None:
--- a/tsfc/fiatinterface.py
+++ b/tsfc/fiatinterface.py
@@ -44,29 +44,23 @@ supported_elements = {
     # These all map directly to FIAT elements
     "Brezzi-Douglas-Marini": FIAT.BrezziDouglasMarini,
     "Brezzi-Douglas-Fortin-Marini": FIAT.BrezziDouglasFortinMarini,
-    "BrokenElement": FIAT.DiscontinuousElement,
     "Bubble": FIAT.Bubble,
     "Crouzeix-Raviart": FIAT.CrouzeixRaviart,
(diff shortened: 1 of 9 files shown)


martinal:
I believe FacetElement and InteriorElement were added by someone on your team because they didn't see that RestrictedElement was covering the same thing.

miklos1:
@martinal: Thanks for answering. I guess we should replace those classes then with deprecated functions that construct the appropriate `RestrictedElement`. @dorugeber?

dorugeber:
Fine.

tsfc/fiatinterface.py
```diff
@@ -178,6 +173,15 @@ def _(element, vector_is_mixed):
...
         return create_element(ufl.FiniteElement("DG", cell, 0), vector_is_mixed)
+    if element.family() == "Quadrature":
+        # Sneaky import from FFC
+        from ffc.quadratureelement import QuadratureElement
+        ffc_element = QuadratureElement(element)
+
+        def tabulate(self, order, points, entity):
+            return QuadratureElement.tabulate(self, order, points)
+        ffc_element.tabulate = types.MethodType(tabulate, ffc_element)
+        return ffc_element
```
miklos1: It's not really a bug from FFC's point of view, but only from TSFC's point of view.

TSFC always passes "FIAT elements" the points in the reference coordinates of the appropriate subentity, and uses the `entity=` parameter to specify which subentity the coordinates are to be interpreted on. FIAT elements have been updated to handle this entity parameter, and map points from facet to cell coordinates themselves if needed, or use the entity parameter directly (e.g. trace element).

Meanwhile, FFC itself translates quadrature points from facet to cell coordinates, and then asks FIAT elements to always tabulate on the cell, and thus never uses the optional entity parameter. Now, `QuadratureElement` resides in FFC, but its duck type is sort of a FIAT element.

blechta: If you find the monkey-patching approach less confusing for both TSFC and FFC, then ok.

miklos1: Not quite sure which one of these two approaches is better.

Perhaps we should just move all FIAT-like elements from FFC to FIAT. If we consider all of FFC, FIAT and TSFC, the current situation is a bit chaotic:

- Both FFC and FIAT have an `EnrichedElement`. FFC uses its own, TSFC uses the one from FIAT.
- Both FFC and TSFC have a `MixedElement`. Both using their own, however, due to form splitting, coefficient splitting and `finat.TensorFiniteElement`, Firedrake never actually uses the TSFC `MixedElement`, so the TSFC copy is only there for the sake of FEniCS.
- FFC has a `QuadratureElement`. FInAT has a smarter `QuadratureElement` which shall be satisfactory for Firedrake in virtually all cases. (Unless someone wants to "enrich" quadrature elements with other elements.) Because FEniCS does not practice form splitting and coefficient splitting, there many more cases in FEniCS when TSFC needs a FIAT-like `QuadratureElement`  which TSFC sneakily imports from FFC.

dorugeber: Just a history note.  The reason I put EnrichedElement in FIAT is so that I could do OPE((U1+U2), V) without having to expand this manually to OPE(U1, V) + OPE(U2, V).  Since I first did OPE at the FIAT level, I also needed to implement + there.

blechta: FFC elements are "FIAT elements" not-implementing the full FIAT element interface. But this might not be an obstacle for your suggestion. I agree that it is quite a chaos now. But I'm not sure, and you'll surely know more, how the suggestion fits an eventual FInAT transition and FIAT retirement.

miklos1: > FFC elements are "FIAT elements" not-implementing the full FIAT element interface.

The `FiniteElement`/`CiarletElement` split somewhat mitigates this problem.

> how the suggestion fits an eventual FInAT transition and FIAT retirement.

It is mostly orthogonal, though any transition is easier with less chaos.

I will file proposal issue.

blechta: > The FiniteElement/CiarletElement split somewhat mitigates this problem.

I don't think that this is fully correct. As of I remember, some FFC elements implement only tabulate and do not bother with, for instance, expansion sets, dmats, etc. But I agree, it would be good to fix this.

miklos1: > expansion sets, dmats

Neither do `FIAT.FiniteElement`s, these features belong to `FIAT.CiarletElement` now.

blechta: Ok, you might be right.

PR #108, opened 2017-04-11
Improve FEniCS compatibility

miklos1:
With these changes TSFC can solve the same FFC regression test examples as UFLACS. This pull request consists of two commits:

- Empty kernels trigger a `gem.impero_utils.NoopError` exception during the compilation, which was caught in `tsfc.compile_form`, skipping the empty kernel. FFC calls `tsfc.compile_integral` directly so this exception is never caught, causing a failure for zero integrals. It seems to me that FFC wants to have empty integral. (@blechta: Is that correct?) So with this change, `NoopError` is caught in `compile_integral`, which then ask the kernel interface construct an empty kernel. The UFC kernel interface complies and constructs a real "empty" kernel, while the Firedrake interface returns `None`, which is then filtered out `tsfc.compile_form`.
- Bindings for Bessel functions. Ordinary Bessel functions translate to the C math API, so they will work with both Firedrake and FEniCS. Modified Bessel functions translate to Boost math special function calls, so they will work with FEniCS but will fail with Firedrake.

### DIFF
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -86,6 +86,7 @@ def compile_integral(integral_data, form_data, prefix, parameters,
     mesh = integral_data.domain
     cell = integral_data.domain.ufl_cell()
     arguments = form_data.preprocessed_form.arguments()
+    kernel_name = "%s_%s_integral_%s" % (prefix, integral_type, integral_data.subdomain_id)
 
     fiat_cell = as_fiat_cell(cell)
     integration_dim, entity_ids = lower_integral_type(fiat_cell, integral_type)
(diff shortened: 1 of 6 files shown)


tsfc/driver.py
```diff
@@ -179,30 +178,34 @@ def compile_integral(integral_data, form_data, prefix, parameters,
...
+        else:
+            for i, quadrature_index in enumerate(quadrature_indices):
+                index_names.append((quadrature_index, 'ip_%d' % i))
+
+        # Construct kernel
+        body = generate_coffee(impero_c, index_names, parameters["precision"], expressions, argument_indices)
+        return builder.construct_kernel(kernel_name, body)
+    except impero_utils.NoopError:
+        # No operations, construct empty kernel
+        return builder.construct_empty_kernel(kernel_name)
```
blechta: Wow, that's quite a lot of lines where the exception can happen...

tsfc/driver.py
```diff
@@ -179,30 +178,34 @@ def compile_integral(integral_data, form_data, prefix, parameters,
...
-            index_names.append((quadrature_index, 'ip_%d' % i))
-
-    body = generate_coffee(impero_c, index_names, parameters["precision"], expressions, argument_indices)
-
     kernel_name = "%s_%s_integral_%s" % (prefix, integral_type, integral_data.subdomain_id)
-    return builder.construct_kernel(kernel_name, body)
+    try:
+        # Construct ImperoC
+        index_ordering = tuple(quadrature_indices) + argument_indices
+        impero_c = impero_utils.compile_gem(assignments, index_ordering, remove_zeros=True)
```
miklos1: @blechta: If the exception happens, it will happen here. However, I shall skip the following lines anyway if the exception happens here.

blechta:
Have to go now but have questions about empty integrals. Will get back to it in 30 minutes.

blechta:
> It seems to me that FFC wants to have empty integral.

In what sense? I can think of several. The unresolved (maybe rather UFL) issue of `0*v*dx`? New parameter `generate_dummy_tabulate_tensor`? Or some other, rather technical reason? (Those cases have different semantics, in the first one tensor needs to be zeroed.)

miklos1:
@blechta: Some kernels simplify to zero not at the UFL level, but only later. Is there a way TSFC could tell FFC that those kernels/integrals are "not needed"?

blechta:
@miklos1: I don't know. But in this case the UFC kernel must zero the tensor. Does it happen in `generate_empty_kernel`?

When does it happen that TSFC simplifies an integral to zero?

miklos1:
@blechta: It seems that both [uflacs](https://bitbucket.org/fenics-project/ffc-reference-data/src/bb91fa356ed3741d0472c3280e1e72143fb8c753/r_uflacs_O0/AdaptivePoisson.h?at=master&fileviewer=file-view-default#AdaptivePoisson.h-14354) and [tensor](https://bitbucket.org/fenics-project/ffc-reference-data/src/bb91fa356ed3741d0472c3280e1e72143fb8c753/r_tensor/AdaptivePoisson.h?at=master&fileviewer=file-view-default#AdaptivePoisson.h-14194) generate an empty/zero kernel, while unoptimised [quadrature](https://bitbucket.org/fenics-project/ffc-reference-data/src/bb91fa356ed3741d0472c3280e1e72143fb8c753/r_quadrature_O0/AdaptivePoisson.h?at=master&fileviewer=file-view-default#AdaptivePoisson.h-14560) tabulates a zero matrix and generates the rest of the form.

Do you mean UFC kernel interface `construct_empty_kernel`? Yes, it will zero out because the glue code is added when treating the arguments. (Since that's when we figure out how big the return buffer is.)

blechta:
@miklos1: Then it seems to me that you're doing the right thing. UFC now requires to zero the tensor every time. So you're doing the right thing.

PR #118, opened 2017-05-08
Fix problem with COMPOUND Conditionals

tj-sun:
Recursively factorise then and else branches of Conditional nodes.

Questions:
- Do we need to look into the condition node of Conditionals (and Comparison nodes as a result)
- The checks for Failure are duplicated, maybe need to move that code outside of main optimisation routine.

### DIFF
--- a/tsfc/coffee_mode.py
+++ b/tsfc/coffee_mode.py
@@ -9,9 +9,11 @@ from collections import defaultdict
 from gem.optimise import (replace_division, make_sum, make_product,
                           unroll_indexsum, replace_delta, remove_componenttensors)
 from gem.refactorise import Monomial, ATOMIC, COMPOUND, OTHER, collect_monomials
-from gem.node import traversal
-from gem.gem import Conditional, Indexed, IndexSum, Failure, one, index_sum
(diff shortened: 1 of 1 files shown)


tsfc/coffee_mode.py
```diff
@@ -48,6 +50,34 @@ def predicate(index):
...
+    raise AssertionError("cannot handle type %s" % type(node))
+
+
+_handle_conditional.register(Node)(reuse_if_untouched)
+
+
+@_handle_conditional.register(Conditional)
+def _handle_conditional_conditional(node, self):
+    condition, then, else_ = node.children
+    then, else_ = optimise_expressions([then, else_], self.argument_multiindices)
```
wence-: The condition could be arbitrarily complex, should we descend there too?

miklos1:
I am not convinced that this will produce a good optimisation.

miklos1:
We shall check out what UFLACS does in this case.

miklos1:
I decline this.

*Reason:* this is a very questionable "fix", and I am now convinced that we shall use the following substitution in some way:

    Conditional(a, b, c) => Conditional(a, 1, 0)*b + Conditional(a, 0, 1)*c

Issue #119, opened 2017-05-09
Factorisation problem with compound Conditional expressions


miklos1:
The default `"coffee"` mode applies unconditional argument factorisation; this, however, fails when the branches of a conditional contain either more than arguments, or an argument with a structured element (such as a vector element or a tensor product element).

First reported in #118.

miklos1:
**Suggestion:** use the following substitution rule to fix:

    Conditional(a, b, c) => Conditional(a, 1, 0)*b + Conditional(a, 0, 1)*c

PR #120, opened 2017-05-09
Rewrite Conditional nodes before optimisation

tj-sun:
Handle Conditional nodes by the following rewriting:

  Conditional(condition, 1, 0) * THEN + Conditional(condition, 0, 1) * ELSE

this ensures Conditional nodes do not have argument indices, and
also factorisation can occur across THEN and ELSE branches

issue #119 

BTW I notice a `NAN` from this test
`test tests/extrusion/test_cylinder.py::test_betti2_cylinder[horiz_complex0-vert_complex0]`
C code:
`double  t0  = (cell_orientations[0][0] == 1 ? -1 : (cell_orientations[0][0] == 0 ? 1 : NAN));`
not sure is that expected?

### DIFF
--- a/gem/refactorise.py
+++ b/gem/refactorise.py
@@ -11,7 +11,8 @@ from itertools import product
 from gem.node import Memoizer, traversal
 from gem.gem import Node, Zero, Product, Sum, Indexed, ListTensor, one
 from gem.optimise import (remove_componenttensors, sum_factorise,
-                          traverse_product, traverse_sum, unroll_indexsum)
+                          traverse_product, traverse_sum, unroll_indexsum,
+                          expand_conditional)
 
 
 # Refactorisation labels
@@ -228,6 +229,10 @@ def collect_monomials(expressions, classifier):
         expressions = unroll_indexsum(expressions,
                                       predicate=lambda i: i in must_unroll)
         expressions = remove_componenttensors(expressions)
+
+    # Expand Conditional nodes which are COMPOUND
+    conditional_predicate = lambda node: classifier(node) == COMPOUND
+    expressions = expand_conditional(expressions, conditional_predicate)
 
     # Finally, refactorise expressions
     mapper = Memoizer(_collect_monomials)
--- a/tsfc/coffee_mode.py
+++ b/tsfc/coffee_mode.py
@@ -37,7 +37,7 @@ def _handle_conditional_conditional(node, self):
     if self.predicate(node):
         return Sum(Product(Conditional(condition, one, Zero()), then),
                    Product(Conditional(condition, Zero(), one), else_))
-    return Conditional(condition, then, else_)
+    return node.reconstruct(condition, then, else_)
 
 
 def handle_conditional(expressions, predicate):
--- a/tsfc/coffee_mode.py
+++ b/tsfc/coffee_mode.py
@@ -46,7 +46,9 @@ def handle_conditional(expressions, argument_indices):
     """Rewrite :class:`Conditional` nodes as:
        Conditional(condition, 1, 0) * THEN + Conditional(condition, 0, 1) * ELSE
     so that factorisation can occur across THEN and ELSE branches.
-    :param expressions: list of GEM DAGs
+    :arg expressions: list of GEM DAGs
+    :arg argument_indices: tuple of argument indices
+
     :return: list of GEM DAGs with :class:`Conditional` nodes rewritten
     """
     mapper = Memoizer(_handle_conditional)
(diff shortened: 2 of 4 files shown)


miklos1:
```c
double t0 = (cell_orientations[0][0] == 1 ? -1 : (cell_orientations[0][0] == 0 ? 1 : NAN));
```
This is fine, and unrelated to the current issue. `cell_orientations` expected to have value 0 or 1, which is translated to +1 and -1. If `cell_orientations` have unexpected value, we continue with a `NaN`.

miklos1:
This is fine, however, can you decouple conditional factorisation from argument factorisation?

tsfc/coffee_mode.py
```diff
@@ -33,13 +33,11 @@ def _handle_conditional(node, self):
...
 @_handle_conditional.register(Conditional)
 def _handle_conditional_conditional(node, self):
-    if not self.predicate(node):
-        return node
-    condition, then, else_ = node.children
-    then = self(then)
-    else_ = self(else_)
-    return Sum(Product(Conditional(condition, one, Zero()), then),
-               Product(Conditional(condition, Zero(), one), else_))
+    condition, then, else_ = map(self, node.children)
```
miklos1: Great use of `map`!

tsfc/coffee_mode.py
```diff
@@ -33,13 +33,11 @@ def _handle_conditional(node, self):
...
+    if self.predicate(node):
+        return Sum(Product(Conditional(condition, one, Zero()), then),
+                   Product(Conditional(condition, Zero(), one), else_))
+    return Conditional(condition, then, else_)
```
miklos1: `reuse_if_untouched` better then "reconstruct always"

tj-sun: Ah I forgot. Amended.

tj-sun:
Something like this?
```
    node = reuse_if_untouched(node, self)
    if self.predicate(node):
        condition, then, else_ = node.children
        return Sum(Product(Conditional(condition, one, Zero()), then),
                   Product(Conditional(condition, Zero(), one), else_))
    return node
```

tsfc/coffee_mode.py
```diff
@@ -21,6 +23,39 @@
...
+    so that factorisation can occur across THEN and ELSE branches.
+    :arg expressions: list of GEM DAGs
+    :arg predicate: a predicate function on :class:`Conditional`s to determine
+                    whether to rewrite the node or not
+
+    :return: list of GEM DAGs with :class:`Conditional` nodes rewritten
+    """
+    mapper = Memoizer(_handle_conditional)
+    mapper.predicate = predicate
+    return list(map(mapper, expressions))
```
miklos1: Let's move this to `gem.optimise` under the name `factorise_conditional`.

tsfc/coffee_mode.py
```diff
@@ -21,6 +23,39 @@
...
+def handle_conditional(expressions, predicate):
+    """Rewrite :class:`Conditional` nodes as:
+       Conditional(condition, 1, 0) * THEN + Conditional(condition, 0, 1) * ELSE
+    so that factorisation can occur across THEN and ELSE branches.
+    :arg expressions: list of GEM DAGs
+    :arg predicate: a predicate function on :class:`Conditional`s to determine
+                    whether to rewrite the node or not
+
+    :return: list of GEM DAGs with :class:`Conditional` nodes rewritten
+    """
```
miklos1: Docstring needs some stylistic and content fixes.

tsfc/coffee_mode.py
```diff
@@ -45,15 +80,21 @@ def predicate(index):
     expressions = replace_delta(expressions)
     expressions = remove_componenttensors(expressions)
     expressions = replace_division(expressions)
-    return optimise_expressions(expressions, argument_multiindices)
+    argument_indices = tuple(itertools.chain(*argument_multiindices))
+
+    def conditional_predicate(argument_indices, node):
+        return len(set(node.free_indices) & argument_indices) >= 1
```
miklos1: If you can rewrite this predicate in terms of the classifier, then the `handle_conditional` call below may go to `collect_monomials`.

gem/refactorise.py
```diff
@@ -229,6 +230,10 @@ def collect_monomials(expressions, classifier):
                                       predicate=lambda i: i in must_unroll)
         expressions = remove_componenttensors(expressions)
 
+    # Expand Conditional nodes which are COMPOUND
+    conditional_predicate = lambda node: classifier(node) == COMPOUND
```
miklos1: Does this pass `flake8`?

tj-sun: Yes... why?

miklos1: There is a rule against assigning `lambda`s to variables. But I guess we ignore that then.

tj-sun: I see. Good to know.

PR #126, opened 2017-06-29
Implement ip and remove PuLP

### LINKED ISSUE #117: Use cvxopt instead of PuLP
Evaluate and potentially switch to using `cvxopt` instead of `PuLP`.

tj-sun:
This PR implements our own IP for coffee mode instead of calling PuLP

- Passes all firedrake tests.
- Matches the flop counts on my test suits.
- Time performance is largely the same as now, if not slightly better (e.g. holzapfel from 5.2s to 4.7s, hyperelasticity in 3D from 3.6s to 3.3s).

Inefficiencies in the algorithm:

- Can do better by keeping track of violated constraints, instead of repeated testing constraints that are already fulfilled across levels.

This supersedes, and thus closes #117.

### DIFF
--- a/tsfc/coffee_mode.py
+++ b/tsfc/coffee_mode.py
@@ -3,8 +3,7 @@ from __future__ import absolute_import, print_function, division
 import numpy
 import itertools
 from functools import partial
-from six import iteritems, iterkeys
-from collections import defaultdict
+from collections import OrderedDict
 from gem.optimise import make_sum, make_product, replace_division, unroll_indexsum
 from gem.refactorise import Monomial, collect_monomials
 from gem.node import traversal
@@ -94,42 +93,37 @@ def monomial_sum_to_expression(monomial_sum):
     return make_sum(indexsums)
 
 
-def solve_ip(size, is_feasible, compare):
+def solve_ip(variables, is_feasible, compare):
     """Solve a 0-1 integer programming problem. The algorithm tries to set each
     variable to 1 recursively, and stop the recursion early by comparing with
     the optimal solution at entry. At worst case this is 2^N (as this is a
     NP-hard problem), but recursions can be trimmed as soon as a reasonable
-    solution have been found. One potential improvement is to keep track of
-    violated constraints at each step, and only try to set the variables which
-    could help these constraints in the next step. This will help find decent
-    solutions faster with the additional overhead of maintaining a list of
-    violated constraints and finding helpful variables.
+    solution have been found.
 
-    :param size: number of 0-1 variables
+    :param varialbes: list of unique 0-1 variables
     :param is_feasible: function to test if a combination is feasible
     :param compare: function to compare two solutions, compare(s1, s2) returns
            True if s1 is a better solution than s2, and False otherwise
     :returns: optimal solution represented as a set of variables with value 1
     """
 
-    optimal_solution = set(range(size))  # start by choosing all atomics
+    optimal_solution = set(variables)  # start by choosing all atomics
     solution = set()
 
     def solve(idx):
-        if idx >= size:
+        if idx >= len(variables):
             return
         if not compare(solution, optimal_solution):
             return
-        solution.add(idx)
+        solution.add(variables[idx])
         if is_feasible(solution):
             if compare(solution, optimal_solution):
                 optimal_solution.clear()
                 optimal_solution.update(solution)
-            # No need to search further
-            # as adding more variable will only make the solution worse
+            # No need to search further as adding more variable will only make the solution worse
         else:
             solve(idx + 1)
-        solution.remove(idx)
+        solution.remove(variables[idx])
         solve(idx + 1)
 
     solve(0)
@@ -147,39 +141,34 @@ def find_optimal_atomics(monomials, argument_indices):
 
     :returns: list of atomic GEM expressions
     """
-    atomic_index = defaultdict(partial(next, itertools.count()))  # atomic -> int
-    connections = []
-    # add connections (list of sets, items in each tuple form a product)
+    atomics = OrderedDict()
     for monomial in monomials:
-        connections.append(set(map(lambda a: atomic_index[a], monomial.atomics)))
+        for atomic in monomial.atomics:
+            atomics[atomic] = 0
+    atomics = list(atomics.keys())
 
-    num_atomics = len(atomic_index)
-    if num_atomics <= 1:
-        return tuple(iterkeys(atomic_index))
+    if len(atomics) <= 1:
+        return tuple(atomics)
 
     def calculate_extent(solution):
-        extent = 0
-        for atomic, index in iteritems(atomic_index):
-            if index in solution:
-                extent += index_extent(atomic, argument_indices)
-        return extent
+        return sum(map(lambda atomic: index_extent(atomic, argument_indices), solution))
 
     def is_feasible(solution):
-        for conn in connections:
-            if not solution.intersection(conn):
+        # Solution is only feasible if it intersects with all monomials
+        # Potentially can improve this by keeping track of violated constraints
+        # and suggest the next atomic to try (instead of just returning True or False)
+        for monomial in monomials:
+            if not solution.intersection(monomial.atomics):
                 return False
         return True
 
     def compare(sol1, sol2):
-        if len(sol1) < len(sol2):
-            return True
-        elif len(sol1) > len(sol2):
-            return False
-        else:
+        if len(sol1) == len(sol2):
             return calculate_extent(sol1) > calculate_extent(sol2)
+        else:
+            return len(sol1) < len(sol2)
 
-    optimal_solution = solve_ip(num_atomics, is_feasible, compare)
-    return tuple(sorted([a for a, i in iteritems(atomic_index) if i in optimal_solution], key=atomic_index.get))
+    return tuple(sorted(solve_ip(atomics, is_feasible, compare)))
 
 
 def factorise_atomics(monomials, optimal_atomics, argument_indices):
--- a/tsfc/coffee_mode.py
+++ b/tsfc/coffee_mode.py
@@ -142,10 +142,12 @@ def find_optimal_atomics(monomials, argument_indices):
     :returns: list of atomic GEM expressions
     """
     atomics = OrderedDict()
+    idx = 0
     for monomial in monomials:
         for atomic in monomial.atomics:
-            atomics[atomic] = 0
-    atomics = list(atomics.keys())
+            if atomic not in atomics:
+                atomics[atomic] = idx
+                idx += 1
 
     if len(atomics) <= 1:
         return tuple(atomics)
@@ -165,7 +167,7 @@ def find_optimal_atomics(monomials, argument_indices):
         else:
             return len(sol1) < len(sol2)
 
-    return tuple(sorted(solve_ip(atomics, is_feasible, compare)))
+    return tuple(sorted(solve_ip(list(atomics.keys()), is_feasible, compare), key=atomics.get))
 
 
 def factorise_atomics(monomials, optimal_atomics, argument_indices):
--- a/tsfc/coffee_mode.py
+++ b/tsfc/coffee_mode.py
@@ -109,6 +109,48 @@ def monomial_sum_to_expression(monomial_sum):
     return make_sum(indexsums)
 
 
+def solve_ip(size, is_feasible, compare):
+    """Solve a 0-1 integer programming problem. The algorithm tries to set each
+    variable to 1 recursively, and stop the recursion early by comparing with
+    the optimal solution at entry. At worst case this is 2^N (as this is a
+    NP-hard problem), but recursions can be trimmed as soon as a reasonable
+    solutions have been found. One potential improvement is to keep track of
+    violated constraints at each step, and only try to set the variables which
+    could help these constraints in the next step. This will help find decent
+    solutions faster with additional overhead of maintaining a list of
+    violated constraints and finding helpful variables.
+
+    :param size: number of 0-1 variables
+    :param is_feasible: function to test if a combination is feasible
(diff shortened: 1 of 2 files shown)


tsfc/coffee_mode.py
```diff
@@ -121,34 +121,50 @@ def find_optimal_atomics(monomials, argument_indices):
     """
     atomic_index = defaultdict(partial(next, itertools.count()))  # atomic -> int
     connections = []
-    # add connections (list of tuples, items in each tuple form a product)
+    # add connections (list of sets, items in each tuple form a product)
     for monomial in monomials:
-        connections.append(tuple(map(lambda a: atomic_index[a], monomial.atomics)))
+        connections.append(set(map(lambda a: atomic_index[a], monomial.atomics)))
```
tj-sun: This should be okay as I only use the connections to intersect with the solution. (every monomial has to have at least one atomic in the solution)

tsfc/coffee_mode.py
```diff
@@ -121,34 +121,50 @@ def find_optimal_atomics(monomials, argument_indices):
...
+        if feasible:
+            if len(solution) < len(optimal_solution) or \
+                    (len(solution) == len(optimal_solution) and calculate_extent(solution) > calculate_extent(optimal_solution)):
+                optimal_solution.clear()
+                optimal_solution.update(solution)
+            # No need to search further
+        else:
+            solve_ip(idx + 1, solution, optimal_solution)
+        solution.remove(idx)
+        solve_ip(idx + 1, solution, optimal_solution)
```
tj-sun: Added a short description about the algorithm.

tsfc/coffee_mode.py
miklos1: OK

tsfc/coffee_mode.py
miklos1: Is this your own algorithm or have you got from someone else?

blechta:
This is great! The FFC regression tests (or the output in general) will not depend on `glpsol` version.

tsfc/coffee_mode.py
```diff
@@ -109,6 +108,49 @@ def monomial_sum_to_expression(monomial_sum):
     return make_sum(indexsums)
 
 
+def solve_ip(size, is_feasible, compare):
+    """Solve a 0-1 integer programming problem. The algorithm tries to set each
+    variable to 1 recursively, and stop the recursion early by comparing with
+    the optimal solution at entry. At worst case this is 2^N (as this is a
+    NP-hard problem), but recursions can be trimmed as soon as a reasonable
+    solutions have been found. One potential improvement is to keep track of
```
dorugeber: typo: "a reasonable solutions"

tsfc/coffee_mode.py
```diff
@@ -109,6 +108,49 @@ def monomial_sum_to_expression(monomial_sum):
...
+    violated constraints at each step, and only try to set the variables which
+    could help these constraints in the next step. This will help find decent
+    solutions faster with additional overhead of maintaining a list of
```
dorugeber: with _the_ additional overhead

tsfc/coffee_mode.py
dorugeber: this part still needs to be addressed

tsfc/coffee_mode.py
tj-sun: Amended.

tsfc/coffee_mode.py
tj-sun: Amended.

tsfc/coffee_mode.py
```diff
@@ -109,6 +108,49 @@ def monomial_sum_to_expression(monomial_sum):
...
+def solve_ip(size, is_feasible, compare):
+    """Solve a 0-1 integer programming problem. The algorithm tries to set each
+    variable to 1 recursively, and stop the recursion early by comparing with
+    the optimal solution at entry. At worst case this is 2^N (as this is a
+    NP-hard problem), but recursions can be trimmed as soon as a reasonable
+    solution have been found. One potential improvement is to keep track of
+    violated constraints at each step, and only try to set the variables which
+    could help these constraints in the next step. This will help find decent
+    solutions faster with the additional overhead of maintaining a list of
+    violated constraints and finding helpful variables.
```
miklos1: The purpose of documentation is to describe what one needs to know to figure out whether some function solves his particular problem and how to use it. As with everything else, it is supposed to be minimal but sufficient. Since the purpose of documentation is not to duplicate the code in English prose, thorough description of implementation details should be avoided in the documentation. If a brief description of the implementation is necessary to help the understanding of the code, it should happen in comments rather than in the documentation.

From "One potential improvement...": Ideation about potential improvements definitely does not belong to the documentation. Make it a comment if you find it necessary to be in the source code.

tsfc/coffee_mode.py
```diff
@@ -121,36 +163,37 @@ def find_optimal_atomics(monomials, argument_indices):
...
+                return False
+        return True
+
+    def compare(sol1, sol2):
+        if len(sol1) < len(sol2):
+            return True
+        elif len(sol1) > len(sol2):
+            return False
+        else:
+            return calculate_extent(sol1) > calculate_extent(sol2)
```
wence-: AKA:

```python
if len(sol1) == len(sol2):
    return calculate_extent(sol1) > calculate_extent(sol2)
else:
    return len(sol1) < len(sol2)
```

tsfc/coffee_mode.py
```diff
@@ -121,36 +163,37 @@ def find_optimal_atomics(monomials, argument_indices):
...
+        for atomic, index in iteritems(atomic_index):
+            if index in solution:
+                extent += index_extent(atomic, argument_indices)
+        return extent
+
+    def is_feasible(solution):
+        for conn in connections:
+            if not solution.intersection(conn):
+                return False
+        return True
```
wence-: So a solution is only feasible if it has a non-empty intersection with the atomics of all the monomials.  If you follow my proposed restructuring above, then this would become clearer because you would write this as:
```python
def is_feasible(solution):
    for monomial in monomials:
         if not solution.intersection(monomial.atomics):
              return False
    return True
```

Which is, I think, `all(solution.intersection(monomial.atomics) for monomial in monomials)`?

tsfc/coffee_mode.py
```diff
@@ -121,36 +163,37 @@ def find_optimal_atomics(monomials, argument_indices):
...
-        raise RuntimeError("Something bad happened during ILP")
-
-    def optimal(atomic):
-        return ilp_var[atomic_index[atomic]].value() == 1
+    def calculate_extent(solution):
+        extent = 0
+        for atomic, index in iteritems(atomic_index):
+            if index in solution:
+                extent += index_extent(atomic, argument_indices)
+        return extent
```
wence-: I note that this code seems unnecessarily complicated.  In this sense:

Previously, when you were passing data off to an external IP solver, you had to set up the problem with data structures it understood (i.e. integer variables, etc...).  But now you control all the data, so there's no need to mimic the data structures that pulp required.  In particular, `solution` is a set of integers that are mapped (via `atomic_index`) to the atomic factors we're trying to pull apart.  Would it not be much simpler if `solution` was just the set of current proposed atomics instead?

Then this could becomes:

```python
def calculate_extent(solution):
    return sum(map(lambda atomic: index_extent(atomic, argument_indices), solution))
```
?

tsfc/coffee_mode.py
```diff
@@ -121,36 +163,37 @@ def find_optimal_atomics(monomials, argument_indices):
...
+        if len(sol1) < len(sol2):
+            return True
+        elif len(sol1) > len(sol2):
+            return False
+        else:
+            return calculate_extent(sol1) > calculate_extent(sol2)
 
-    return tuple(sorted(filter(optimal, atomic_index), key=atomic_index.get))
+    optimal_solution = solve_ip(num_atomics, is_feasible, compare)
+    return tuple(sorted([a for a, i in iteritems(atomic_index) if i in optimal_solution], key=atomic_index.get))
```
wence-: And these two lines become:

`return tuple(sorted(solve_ip(num_atomics, is_feasible, compare)))`

wence-:
Since @tj-sun is away for much of the rest of the summer, I might have a go at cleaning this up next week as part of py3 (in part to remove our compiled dependency list).

tj-sun:
Somehow missed this earlier... Will have a go over the weekend.

tsfc/coffee_mode.py
tj-sun: Amended. Does look much clearer now.

tsfc/coffee_mode.py
tj-sun: Amended.

tsfc/coffee_mode.py
```diff
@@ -94,42 +93,37 @@ def monomial_sum_to_expression(monomial_sum):
...
     NP-hard problem), but recursions can be trimmed as soon as a reasonable
-    solution have been found. One potential improvement is to keep track of
-    violated constraints at each step, and only try to set the variables which
-    could help these constraints in the next step. This will help find decent
-    solutions faster with the additional overhead of maintaining a list of
-    violated constraints and finding helpful variables.
+    solution have been found.
 
-    :param size: number of 0-1 variables
+    :param varialbes: list of unique 0-1 variables
```
miklos1: Typo.

tsfc/coffee_mode.py
tj-sun: Amended.

tsfc/coffee_mode.py
tj-sun: Amended.

tsfc/coffee_mode.py
tj-sun: Removed the second part, added comments to `is_feasilbe()`

tsfc/coffee_mode.py
```diff
@@ -105,38 +141,33 @@ def find_optimal_atomics(monomials, argument_indices):
...
+        for atomic in monomial.atomics:
+            if atomic not in atomics:
+                atomics[atomic] = idx
+                idx += 1
+
+    if len(atomics) <= 1:
+        return tuple(atomics)
+
+    def calculate_extent(solution):
+        return sum(map(lambda atomic: index_extent(atomic, argument_indices), solution))
```
miklos1: "Extent" does not have much meaning for solutions, it is more like the cost, or, more precisely, part of the cost.

tsfc/coffee_mode.py
```diff
@@ -95,6 +93,44 @@ def monomial_sum_to_expression(monomial_sum):
...
+    """Solve a 0-1 integer programming problem. The algorithm tries to set each
+    variable to 1 recursively, and stop the recursion early by comparing with
+    the optimal solution at entry. At worst case this is 2^N (as this is a
+    NP-hard problem), but recursions can be trimmed as soon as a reasonable
+    solution have been found.
+
+    :param varialbes: list of unique 0-1 variables
+    :param is_feasible: function to test if a combination is feasible
+    :param compare: function to compare two solutions, compare(s1, s2) returns
+           True if s1 is a better solution than s2, and False otherwise
```
miklos1: You call it compare, but returning a logical value, it is a *strictly less* operator. I still think you should go the Python way, and have a key function instead.

tsfc/coffee_mode.py
```diff
@@ -105,38 +141,33 @@ def find_optimal_atomics(monomials, argument_indices):
...
-    ilp_prob.solve()
-    if ilp_prob.status != 1:
-        raise RuntimeError("Something bad happened during ILP")
-
-    def optimal(atomic):
-        return ilp_var[atomic_index[atomic]].value() == 1
+        for atomic in monomial.atomics:
+            if atomic not in atomics:
+                atomics[atomic] = idx
+                idx += 1
```
miklos1: Why was it a good idea to spell out `defaultdict(partial(next, itertools.count()))`? You don't even really need this mapping anymore. Just take:
```python
atomics = tuple(OrderedDict.fromkeys(*[monomial.atomics for monomial in monomials]))
```

tsfc/coffee_mode.py
```diff
@@ -105,38 +141,33 @@ def find_optimal_atomics(monomials, argument_indices):
...
+
+    if len(atomics) <= 1:
+        return tuple(atomics)
```
miklos1: Is this branch really needed?

tsfc/coffee_mode.py
```diff
@@ -105,38 +141,33 @@ def find_optimal_atomics(monomials, argument_indices):
...
+        # Solution is only feasible if it intersects with all monomials
+        # Potentially can improve this by keeping track of violated constraints
+        # and suggest the next atomic to try (instead of just returning True or False)
+        return all(solution.intersection(monomial.atomics) for monomial in monomials)
+
+    def compare(sol1, sol2):
+        if len(sol1) == len(sol2):
+            return calculate_extent(sol1) > calculate_extent(sol2)
+        else:
+            return len(sol1) < len(sol2)
```
miklos1: Please go for key function instead!

tsfc/coffee_mode.py
```diff
@@ -105,38 +141,33 @@ def find_optimal_atomics(monomials, argument_indices):
...
 
-    return tuple(sorted(filter(optimal, atomic_index), key=atomic_index.get))
+    return tuple(sorted(solve_ip(list(atomics.keys()), is_feasible, compare), key=atomics.get))
```
miklos1:
```python
optimal_atomics = solve_ip(atomics, is_feasible, compare)
return tuple(atomic for atomic in atomics if atomic in optimal_atomics)
```
Last line takes care of consistent ordering.

wence-:
OK, I've cleaned up those points.  How do I test that this didn't regress the flop counts?

wence-:
OK, this produces same flop counts as before my changes, so will merge.

PR #128, opened 2017-07-17
Check estimated quadrature degree against function degrees

miklos1:
When estimated quadrature degree is used, issue a warning if the estimated degree is more than tenfold greater than any argument, coordinate, or coefficient degree.

Replacement for [this check](https://github.com/firedrakeproject/fiat/blob/master/FIAT/quadrature_schemes.py#L104), which causes firedrakeproject/fiat#9.

### DIFF
--- a/tests/test_estimated_degree.py
+++ b/tests/test_estimated_degree.py
@@ -1,0 +1,35 @@
+from __future__ import absolute_import, print_function, division
+import logging
+
+import pytest
+
+import ufl
+from tsfc import compile_form
+from tsfc.logging import logger
+
+
+class MockHandler(logging.Handler):
+    def emit(self, record):
+        raise RuntimeError()
+
+
+def test_estimated_degree():
+    cell = ufl.tetrahedron
+    mesh = ufl.Mesh(ufl.VectorElement('P', cell, 1))
+    V = ufl.FunctionSpace(mesh, ufl.FiniteElement('P', cell, 1))
+    f = ufl.Coefficient(V)
+    u = ufl.TrialFunction(V)
+    v = ufl.TestFunction(V)
+    a = u * v * ufl.tanh(ufl.sqrt(ufl.sinh(f) / ufl.sin(f**f))) * ufl.dx
+
+    handler = MockHandler()
+    logger.addHandler(handler)
+    with pytest.raises(RuntimeError):
+        compile_form(a)
+    logger.removeHandler(handler)
+
+
+if __name__ == "__main__":
+    import os
+    import sys
+    pytest.main(args=[os.path.abspath(__file__)] + sys.argv[1:])
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -8,6 +8,8 @@ import time
 from functools import reduce
 from itertools import chain
 
+from numpy import asarray
+
 import ufl
 from ufl.algorithms import extract_arguments, extract_coefficients
 from ufl.algorithms.analysis import has_type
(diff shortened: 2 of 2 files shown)


tsfc/driver.py
```diff
@@ -145,6 +145,18 @@ def compile_integral(integral_data, form_data, prefix, parameters,
         quadrature_degree = params.get("quadrature_degree",
                                        params["estimated_polynomial_degree"])
         try:
+            quadrature_degree = params["quadrature_degree"]
+        except KeyError:
+            quadrature_degree = params["estimated_polynomial_degree"]
+            functions = list(arguments) + [coordinates] + list(integral_data.integral_coefficients)
+            function_degrees = [f.ufl_function_space().ufl_element().degree() for f in functions]
+            if all(quadrature_degree > 10 * deg for deg in function_degrees):
```
wence-: Does this work for tuple degrees?  e.g. `(4, 4) > 10 * (2,2) => True`

miklos1: It does not seem to fail the Firedrake test suite for me, but that's probably because in Python 2 `(4, 4) > 10 * (2,2) => (4, 4) > (2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2) => True`.

miklos1:
@wence-: Do you have a sample form triggering this that we can add as a test case?

wence-:
How about:

```python
from firedrake import *
mesh = UnitCubeMesh(1, 1, 1)

V = FunctionSpace(mesh, "CG", 1)

u = TrialFunction(V)

v = TestFunction(V)

f = Function(V)

a = u*v*tanh(sqrt(sinh(f) / sin(f**f)))*dx

assemble(a)
=> tsfc:WARNING Estimated quadrature degree 13 more than tenfold greater than any argument/coefficient degree (max 1)
```

miklos1:
Thanks, now I just need to find a nice way to test for logging events.

wence-:
```
In [10]: class FooHandler(logging.Handler):
    ...:     def emit(self, record):
    ...:         raise RuntimeError()
    ...:     

In [11]: handler = FooHandler(level=logging.WARNING)

In [12]: tsfc.logging.logger.addHandler(handler)
```

miklos1:
I changed it to raising an exception.

miklos1:
Raising an exception fails `test_ip_viscosity.py` in Firedrake because 11 vs 1.

wence-:
Does that test pass if you manually specify degree?

miklos1:
> Does that test pass if you manually specify degree?

Yes, though it isn't very obvious how to do that, since there are many integrals, and they are parametrised in family/degree.

miklos1:
Okay, what shall I do with this? Go back to logging and implement the testing for logging?

wence-:
That would be fine.

miklos1:
This updated now.

tests/test_estimated_degree.py
```diff
@@ -0,0 +1,40 @@
...
+class MockHandler(logging.Handler):
+    def emit(self, record):
+        raise RuntimeError()
+
+
+@contextlib.contextmanager
+def push_handler(logger, handler):
+    logger.addHandler(handler)
+    yield
+    logger.removeHandler(handler)
```
wence-: Careful, I think this doesn't do what you want:

```python
import contextlib
foo = 0
@contextlib.contextmanager
def bar():
    global foo
    foo += 1
    yield
    foo -= 1

try:
    with bar():
        raise RuntimeError
except:
     pass

print foo
=> 1
```

miklos1: What? Why?

miklos1: This is useless, what's the point of context managers if they don't actually manager context in presence of exceptions?

wence-: Lolz, inorite!

blechta: What about
```
try:
    yield
finally:
    logger.removeHandler(handler)
```
?

miklos1: What is the point of a context manager if I manually have to write out try-finally clauses? It's just stupid. They could have, and should have done it in `contextlib.contextmanager`.

blechta: I don't think you're right. Author of a context manager decides what is right thing to do, here `finally: logger.removeHandler(handler)`, ans its client can just seamlessly use it, not seeing the try-finally clause. There are many different scenarios which would have sense in other cases and it is well documented, see https://docs.python.org/3/library/contextlib.html. `contextlib.contextmanager` is a general tool which requires the dev being explicit without hidden magic. Other wrappers in the package have some implicit error handling.

blechta: I don't think you're right. Author of a context manager decides what is right thing to do, here `finally: logger.removeHandler(handler)`, ans its client can just seamlessly use it, not seeing the try-finally clause. There are many different scenarios which would have sense in other cases and it is well documented, see https://docs.python.org/3/library/contextlib.html. `contextlib.contextmanager` is a general tool which requires the dev being explicit without hidden magic. Other wrappers in the package have some implicit error handling.

PR #129, opened 2017-07-17
Interpreter fixes

miklos1:


gem/interpreter.py
```diff
@@ -274,18 +294,23 @@ def _(e, self):
...
-    assert all(ops[0].fids == o.fids for o in ops)
-    return Result(numpy.asarray([o.arr for o in ops]).reshape(e.shape),
-                  ops[0].fids)
+    tmp = Result.empty(*ops)
+    arrs = []
+    for o in ops:
+        arr = numpy.empty(tmp.fshape)
+        arr[:] = o.broadcast(tmp.fids)
+        arrs.append(arr)
+    arrs = numpy.moveaxis(numpy.asarray(arrs), 0, -1).reshape(tmp.fshape + e.shape)
```
wence-: This is because list tensors can have entries with different free indices?

gem/interpreter.py
```diff
@@ -29,6 +29,27 @@ def __init__(self, arr, fids=None):
...
+        axes = tuple(self.fids.index(fi) for fi in fids if fi in self.fids)
+        assert len(axes) == len(self.fids)
+        # Add shape
+        axes += tuple(range(len(self.fids), self.arr.ndim))
+        # Move axes, insert extra axes
+        arr = numpy.transpose(self.arr, axes)
+        for i, fi in enumerate(fids):
+            if fi not in self.fids:
+                arr = numpy.expand_dims(arr, axis=i)
+        return arr
```
miklos1: `numpy.transpose` in the generic sense is a reordering of axes/dimensions, so yes, we need to do that, because free indices might appear in different order. `numpy.reshape`: sort of, but not quite. No axes are split or merged, but length-one dimensions are inserted, which will expand to the correct length through broadcasting.

gem/interpreter.py
```diff
@@ -274,18 +294,23 @@ def _(e, self):
 def _(e, self):
     """Index sums reduce over the given axis."""
     val = self(e.children[0])
-    idx = val.fids.index(e.index)
-    return Result(val.arr.sum(axis=idx),
-                  val.fids[:idx] + val.fids[idx+1:])
+    idx = tuple(map(val.fids.index, e.multiindex))
+    rfids = tuple(fi for fi in val.fids if fi not in e.multiindex)
```
miklos1: Yes, the reason this needs updating is that `gem.IndexSum` has changed.

gem/interpreter.py
miklos1: Yes, in particular, one of its entries might be zero (which loses free indices in GEM, but not in UFL), while others may have free indices.

Issue #89, opened 2016-12-12
Optimise away Kronecker deltas for arguments


miklos1:
It is a question whether this is best done in TSFC or in COFFEE.

miklos1:
Given @tj-sun's work, this is going to take place in TSFC eventually.

PR #130, opened 2017-07-18
Some convergence between 'coffee' and 'spectral' modes

miklos1:
Spectral mode:
- Adopts tighter argument factorisation from `'coffee'` mode.
  Not necessary for sum factorisation, but useful for delta elimination on arguments.
- Not needed to call `remove_componenttensors` afterwards, as no new `ComponentTensor` nodes are introduced.

Coffee mode:
- No longer lowers `Delta` nodes early.
- Eliminate unnecessary `remove_componenttensors` call at the beginning. `collect_monomials` takes care of this itself.

### DIFF
--- a/tsfc/coffee_mode.py
+++ b/tsfc/coffee_mode.py
@@ -6,19 +6,19 @@ from functools import partial
 from six import iteritems, iterkeys
 from six.moves import filter
 from collections import defaultdict
-from gem.optimise import (replace_division, make_sum, make_product,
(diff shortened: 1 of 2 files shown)


miklos1:
I can confirm that `'coffee'` mode (tested this patch, but probably does not matter) does not break if unrolling of short `IndexSum`s is turned off. (This is not the case for the old COFFEE package.)

miklos1:
Related to #89.

PR #131, opened 2017-07-20
Preliminary additions and fixes for supporting enriched FInAT element

miklos1:
- Adds new GEM node: `Concatenate`
- Fixes an accidental axis reordering issue with `gem.view` and `gem.reshape`, which apply to run-time kernel arguments.
- Teaches `gem.select_expression` (which factorises element expressions for facet integrals) about `ComponentTensor`s (which will appear inside `Concatenate` nodes).

Related to FInAT/FInAT#19.

### DIFF
--- a/gem/interpreter.py
+++ b/gem/interpreter.py
@@ -115,20 +115,20 @@ def _evaluate(expression, self):
     raise ValueError("Unhandled node type %s" % type(expression))
 
 
-@_evaluate.register(gem.Zero)  # noqa: F811
-def _(e, self):
+@_evaluate.register(gem.Zero)
+def _evaluate_zero(e, self):
     """Zeros produce an array of zeros."""
     return Result(numpy.zeros(e.shape, dtype=float))
 
 
-@_evaluate.register(gem.Constant)  # noqa: F811
-def _(e, self):
+@_evaluate.register(gem.Constant)
+def _evaluate_constant(e, self):
     """Constants return their array."""
     return Result(e.array)
 
 
-@_evaluate.register(gem.Variable)  # noqa: F811
-def _(e, self):
+@_evaluate.register(gem.Variable)
+def _evaluate_variable(e, self):
     """Look up variables in the provided bindings."""
     try:
         val = self.bindings[e]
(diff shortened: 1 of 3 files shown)


gem/interpreter.py
```diff
@@ -313,6 +314,25 @@ def _(e, self):
     return Result(arrs, tmp.fids)
 
 
+@_evaluate.register(gem.Concatenate)
+def _evaluate_concatenate(e, self):
+    """Concatenate nodes flatten and concatenate shapes."""
+    ops = [self(o) for o in e.children]
+    fids = tuple(OrderedDict.fromkeys(itertools.chain(*(o.fids for o in ops))))
```
wence-: This is `tuple(set(fids))` but order-preserving.

gem/optimise.py
```diff
@@ -187,6 +187,12 @@ def child(expression):
     if types <= {Literal, Zero, Failure}:
         return partial_indexed(ListTensor(expressions), (index,))
 
+    if types <= {ComponentTensor, Zero}:
```
wence-: So this function is called on `exprs` which have had component tensors removed (L228-230).  So how can a `ComponentTensor` show up here?

wence-: Oh, they won't show up right now, but will later?

gem/optimise.py
```diff
@@ -187,6 +187,12 @@ def child(expression):
...
+        shape, = set(e.shape for e in expressions)
```
wence-: This is like `expressions[0].shape` but safe against "bad" input? (`select_expressions` asserts that all shapes are the same, but `_select_expressions (called recursively) does not?)

gem/optimise.py
miklos1: Yes, because we cannot remove `ComponentTensor` directly inside `Concatenate` nodes. Similarly, we cannot remove `ComponentTensor` nodes at the root of the expression.

gem/interpreter.py
miklos1: Yes.

gem/optimise.py
miklos1: Yes, it has an implicit but checked assumption that all shapes are the same.

PR #132, opened 2017-07-21
Support FInAT enriched element

miklos1:
Still being tested for breakages and regressions, but code review should start.

### DIFF
--- a/gem/gem.py
+++ b/gem/gem.py
@@ -560,8 +560,8 @@ class FlexiblyIndexed(Scalar):
     def index_ordering(self):
         """Running indices in the order of indexing in this node."""
         return tuple(index
-                     for offset, idxs in self.dim2idxs
-                     for index, stride in idxs
+                     for _, idxs in self.dim2idxs
+                     for index, _ in idxs
                      if isinstance(index, Index))
 
 
--- a/gem/gem.py
+++ b/gem/gem.py
@@ -500,6 +500,10 @@ class Indexed(Scalar):
 
         return self
 
+    def index_ordering(self):
+        """Running indices in the order of indexing in this node."""
+        return tuple(i for i in self.multiindex if isinstance(i, Index))
+
 
 class FlexiblyIndexed(Scalar):
     """Flexible indexing of :py:class:`Variable`s to implement views and
@@ -552,6 +556,13 @@ class FlexiblyIndexed(Scalar):
         self.children = (variable,)
         self.dim2idxs = tuple(dim2idxs_)
         self.free_indices = unique(free_indices)
+
+    def index_ordering(self):
+        """Running indices in the order of indexing in this node."""
+        return tuple(index
+                     for offset, idxs in self.dim2idxs
+                     for index, stride in idxs
+                     if isinstance(index, Index))
 
 
 class ComponentTensor(Node):
--- a/gem/impero_utils.py
+++ b/gem/impero_utils.py
@@ -15,7 +15,6 @@ import itertools
 from singledispatch import singledispatch
 
 from gem.node import traversal, collect_refcount
-from gem.utils import OrderedSet
 from gem import gem, impero as imp, optimise, scheduling
 
 
@@ -61,17 +60,17 @@ def compile_gem(assignments, prefix_ordering, remove_zeros=False):
     expressions = [expression for variable, expression in assignments]
 
     # Collect indices in a deterministic order
-    indices = OrderedSet()
+    indices = collections.OrderedDict()
     for node in traversal(expressions):
         if isinstance(node, gem.Indexed):
             for index in node.multiindex:
                 if isinstance(index, gem.Index):
-                    indices.add(index)
+                    indices.setdefault(index)
         elif isinstance(node, gem.FlexiblyIndexed):
             for offset, idxs in node.dim2idxs:
                 for index, stride in idxs:
                     if isinstance(index, gem.Index):
-                        indices.add(index)
+                        indices.setdefault(index)
 
     # Build ordered index map
     index_ordering = make_prefix_ordering(indices, prefix_ordering)
--- a/gem/impero_utils.py
+++ b/gem/impero_utils.py
@@ -10,7 +10,7 @@ from __future__ import absolute_import, print_function, division
 from six.moves import filter
 
 import collections
-import itertools
+from itertools import chain, groupby
 
 from singledispatch import singledispatch
 
@@ -60,17 +60,11 @@ def compile_gem(assignments, prefix_ordering, remove_zeros=False):
     expressions = [expression for variable, expression in assignments]
 
     # Collect indices in a deterministic order
-    indices = collections.OrderedDict()
-    for node in traversal(expressions):
-        if isinstance(node, gem.Indexed):
-            for index in node.multiindex:
-                if isinstance(index, gem.Index):
-                    indices.setdefault(index)
-        elif isinstance(node, gem.FlexiblyIndexed):
-            for offset, idxs in node.dim2idxs:
-                for index, stride in idxs:
-                    if isinstance(index, gem.Index):
-                        indices.setdefault(index)
+    indices = list(collections.OrderedDict.fromkeys(chain.from_iterable(
+        node.index_ordering()
+        for node in traversal(expressions)
+        if isinstance(node, (gem.Indexed, gem.FlexiblyIndexed))
+    )))
 
     # Build ordered index map
     index_ordering = make_prefix_ordering(indices, prefix_ordering)
@@ -174,7 +168,7 @@ def make_loop_tree(ops, get_indices, level=0):
     """
     keyfunc = lambda op: op.loop_shape(get_indices)[level:level+1]
     statements = []
-    for first_index, op_group in itertools.groupby(ops, keyfunc):
+    for first_index, op_group in groupby(ops, keyfunc):
         if first_index:
             inner_block = make_loop_tree(op_group, get_indices, level+1)
             statements.append(imp.For(first_index[0], inner_block))
--- a/gem/unconcatenate.py
+++ b/gem/unconcatenate.py
@@ -173,7 +173,7 @@ def _unconcatenate(cache, pairs):
 
 
 def unconcatenate(pairs, cache=None):
-    """Splits a list of (variable reference, expression) pairs along
+    """Splits a list of (indexed variable, expression) pairs along
     :py:class:`Concatenate` nodes embedded in the expressions.
 
     :param pairs: list of (indexed variable, expression) pairs
--- a/gem/unconcatenate.py
+++ b/gem/unconcatenate.py
@@ -1,0 +1,192 @@
+"""Utility functions for decomposing Concatenate nodes."""
+
+from __future__ import absolute_import, print_function, division
+from six.moves import range, zip
+
+from itertools import chain
+
+import numpy
+
+from gem.node import Memoizer, reuse_if_untouched
+from gem.gem import (ComponentTensor, Concatenate, FlexiblyIndexed,
+                     Index, Indexed, Node, reshape, view)
+from gem.optimise import remove_componenttensors
+
+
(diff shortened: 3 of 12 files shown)


gem/gem.py
```diff
@@ -553,6 +557,13 @@ def __init__(self, variable, dim2idxs):
         self.dim2idxs = tuple(dim2idxs_)
         self.free_indices = unique(free_indices)
 
+    def index_ordering(self):
+        """Running indices in the order of indexing in this node."""
+        return tuple(index
+                     for offset, idxs in self.dim2idxs
+                     for index, stride in idxs
```
wence-: Could write `for _, ids in self.dim2idxs for index, _ in idxs` so it's clear that those variables aren't being used?

gem/impero_utils.py
```diff
@@ -61,17 +60,11 @@ def nonzero(assignment):
...
-        elif isinstance(node, gem.FlexiblyIndexed):
-            for offset, idxs in node.dim2idxs:
-                for index, stride in idxs:
-                    if isinstance(index, gem.Index):
-                        indices.add(index)
+    indices = list(collections.OrderedDict.fromkeys(chain.from_iterable(
+        node.index_ordering()
+        for node in traversal(expressions)
+        if isinstance(node, (gem.Indexed, gem.FlexiblyIndexed))
+    )))
```
wence-: OK, simplifying existing code.  You collect unique indices in the order they appear in the dag traversal.

gem/unconcatenate.py
```diff
@@ -0,0 +1,192 @@
...
+        node = lifo.pop()
+        if not free_indices.intersection(node.free_indices):
+            continue
+
+        if isinstance(node, Indexed):
+            child, = node.children
+            if isinstance(child, Concatenate):
+                i, = node.multiindex
+                assert i in free_indices
+                if (index or i) == i:
```
wence-: OK, so if this is the first `Concatenate`, collect it, otherwise only collect it if its free indices match.  So this doesn't produce all `Concatenate` nodes, just a subset with matching free indices (but not the list of such subsets).

gem/unconcatenate.py
```diff
@@ -0,0 +1,192 @@
...
+
+def split_variable(variable_ref, index, multiindices):
+    """Splits a flexibly indexed variable along a concatenation index.
+
+    :param variable_ref: flexibly indexed variable to split
+    :param index: :py:class:`Concatenate` index to split along
+    :param multiindices: one multiindex for each split variable
+
+    :returns: generator of split indexed variables
+    """
```
wence-: OK, so let's see if I understand what this is doing.  `variable_ref` is some kind of indexed object.  First I add shape to it in the axis order `index`, then all the other indices.

Now I want to pull apart the first dimension into sub-indexed bits, corresponding to the different parts of the Concatenate node I'm splitting?  So I index into a reshaped view of the component tensor.  But really I'm only slicing the first dimension, it's just you have to write this code because more general "numpy-like" slicing isn't implemented on component tensors?

gem/unconcatenate.py
```diff
@@ -0,0 +1,192 @@
...
+    mapper = Memoizer(_replace_node)
+    mapper.mapping = mapping
+    mapper.cut = cut or (lambda node: False)
+    return mapper(expression)
+
+
+def _unconcatenate(cache, pairs):
+    # Tail-call recursive core of unconcatenate.
+    # Assumes that input has already been sanitised.
+    concat_group = find_group([e for v, e in pairs])
```
wence-: OK, so look for a `Concatenate` node.  This is a bit like `groupby`?

miklos1:
With TSFC vanilla and COFFEE -O2 (previous default configuration) two Firedrake test cases break with COFFEE: https://gist.github.com/miklos1/04e2c6832e0d53e45a4c9c9bbee72f11 and
```
$ py.test -v -x tests/extrusion/test_enrichment_1_feec.py::test_feec[horiz_complex5-vert_complex5]
...
E           ConvergenceError: Nonlinear solve failed to converge after 0 nonlinear iterations.
E           Reason:
E              Inner linear solve failed to converge after 0 iterations with reason: unknown reason (petsc4py enum incomplete?), try with -snes_converged_reason and -ksp_converged_reason

firedrake/solving_utils.py:233: ConvergenceError
---------------------------------- Captured stderr call -----------------------------------
```

Not sure if we care about these.

gem/unconcatenate.py
```diff
@@ -0,0 +1,192 @@
...
+    if concat_group is None:
+        return pairs
```
wence-: None found, so we're at the root of the recursion.

gem/unconcatenate.py
```diff
@@ -0,0 +1,192 @@
...
+
+    # Run again, there may be other Concatenate groups
+    return _unconcatenate(cache, split_pairs)
+
+
+def unconcatenate(pairs, cache=None):
+    """Splits a list of (variable reference, expression) pairs along
+    :py:class:`Concatenate` nodes embedded in the expressions.
+
+    :param pairs: list of (indexed variable, expression) pairs
```
wence-: Which is it? indexed variable, or variable reference?

gem/unconcatenate.py
```diff
@@ -0,0 +1,192 @@
...
+    :param cache: index splitting cache :py:class:`dict` (optional)
+
+    :returns: list of (indexed variable, expression) pairs
+    """
```
wence-: So the idea is that if I have an expression that contains a `Concatenate(a, b)` then I get back two expressions, one that has `a` and the other `b`?

gem/unconcatenate.py
```diff
@@ -0,0 +1,192 @@
...
+        return index not in node.free_indices
+
+    # Build Concatenate node replacement mappings
+    mappings = [{} for i in range(len(multiindices))]
+    for concat_ref in concat_group:
+        concat_expr, = concat_ref.children
+        for i in range(len(multiindices)):
+            sub_ref = Indexed(concat_expr.children[i], multiindices[i])
+            sub_ref, = remove_componenttensors((sub_ref,))
+            mappings[i][concat_ref] = sub_ref
```
wence-: OK, so we're going to replace the `Concatenate` node with the appropriate sub component.

gem/unconcatenate.py
```diff
@@ -0,0 +1,192 @@
...
+
+    # Finally, split assignment pairs
+    split_pairs = []
+    for var, expr in pairs:
+        if index not in var.free_indices:
+            split_pairs.append((var, expr))
```
wence-: So you expect not to get (or care about) an expression that has `IndexSum(Indexed(Concatenate(...), i), i)`

tsfc/coffee_mode.py
```diff
@@ -61,7 +60,7 @@ def optimise_expressions(expressions, argument_indices):
             return expressions
 
     # Apply argument factorisation unconditionally
-    classifier = partial(classify, set(argument_indices))
+    classifier = partial(spectral.classify, set(argument_indices))
     monomial_sums = collect_monomials(expressions, classifier)
     return [optimise_monomial_sum(ms, argument_indices) for ms in monomial_sums]
```
wence-: OK, so this is just merging some spectral and coffee technology.

tsfc/driver.py
```diff
@@ -205,25 +210,43 @@ def compile_integral(integral_data, form_data, prefix, parameters,
...
+            for multiindex, suffix in zip(index_cache[index],
+                                          string.ascii_lowercase):
+                name_multiindex(multiindex, name + suffix)
+
+    def name_multiindex(multiindex, name):
+        if len(multiindex) == 1:
+            name_index(multiindex[0], name)
+        else:
+            for i, index in enumerate(multiindex):
+                name_index(index, name + str(i))
```
wence-: For consistency, should we either add an underscore here or remove it from the quadrature index?

tsfc/driver.py
```diff
@@ -205,25 +210,43 @@ def compile_integral(integral_data, form_data, prefix, parameters,
...
+
+    for multiindex, name in zip(argument_multiindices, ['j', 'k']):
+        name_multiindex(multiindex, name)
+
     if len(quadrature_indices) == 1:
         index_names.append((quadrature_indices[0], 'ip'))
     else:
         for i, quadrature_index in enumerate(quadrature_indices):
             index_names.append((quadrature_index, 'ip_%d' % i))
```
wence-: I think I would prefer removing it here.

tsfc/vanilla.py
```diff
@@ -29,19 +30,20 @@ def predicate(index):
...
+    :arg index_cache: cache :py:class:`dict` for :py:func:`unconcatenate`
 
     :returns: series of (return variable, GEM expression root) pairs
     """
-    for variable, reps in var_reps:
-        expressions = reps  # representations are expressions
-        yield (variable, reduce(Sum, expressions))
+    return unconcatenate([(variable, reduce(Sum, reps))
+                          for variable, reps in var_reps],
+                         cache=index_cache)
```
wence-: OK, so actually unconcat the representation when generating code.

tsfc/tensor.py
```diff
@@ -83,7 +83,7 @@ def classify(quadrature_indices, expression):
     return result
 
 
-def flatten(var_reps):
+def flatten(var_reps, index_cache):
     for variable, reps in var_reps:
         expressions = reps
         for expression in expressions:
```
wence-: No unconcat handling in tensor because it won't work, or because it's not implemented yet?

gem/gem.py
miklos1: OK, will do.

gem/unconcatenate.py
miklos1: Yes. I think when we get to NCF elements, we will have `Concatenate` nodes inside `Concatenate` nodes, so we have to unpick the outer ones first before the inner ones are consistently indexed.

gem/unconcatenate.py
miklos1: Yes.

gem/unconcatenate.py
miklos1: Not sure what you mean. It does precisely what you described in `find_group`.

gem/unconcatenate.py
miklos1: It does happen during coefficient evaluation, but that also contains within the `IndexSum` a corresponding kernel argument that needs to be split. So it is always pairs of flexibly indexed variable and expression containing the `Concatenate` node.

gem/unconcatenate.py
miklos1: Indexed variable.

gem/unconcatenate.py
miklos1: Basically, yes.

tsfc/driver.py
miklos1: OK.

tsfc/driver.py
miklos1: OK.

tsfc/tensor.py
miklos1: I should probably do something more here not to break `EnrichedElement`s in the tensor mode. This was just added because `flatten` not accepting the extra argument fails test cases in TSFC.

miklos1:
Nothing breaks on default configuration, but the Firedrake test suite takes about 2-3% more time, based on one measurement with and one without this patch.

miklos1:
I am wondering if it could make sense not to do this for low-degree elements. It should be easy to get back the unstructured version, I can make a pass that looks for `Concatenate` nodes, pulls them through the interpreter, casts the result back to GEM, and replaces the original node in the expression with that.

miklos1:
Spectral mode still works as well.

miklos1:
> I think the unconcatenate code could maybe do with an expository example?

@wence-: What form should be used to do this? Test cases, docstrings, comments, or some Sphinx-style documentation?

wence-:
Maybe just some comments?

Issue #135, opened 2017-07-27
Some gusto tests now take forever to optimise kernels


wence-:
The ILP solver does the naive `2**N` thing.  We need to cut off somewhere.  For example, one kernel has 24 variables, so we're checking `2**24` options.

miklos1:
So I guess this is a regression from #126.

wence-:
I guess, yes. We could just truncate the search?

miklos1:
You mean to settle for a suboptimal solution in less time?

dham:
What did the previous solution do in this case?

miklos1:
> What did the previous solution do in this case?

miklos1:
> What did the previous solution do in this case?

You mean `glpsol`? Something smart, I guess...

wence-:
Turns out the `hdivcurl` branch helps quite a lot.  But you can try out a "problematic" form here:

```python
from __future__ import absolute_import, print_function, division
from firedrake import *
from firedrake.formmanipulation import split_form
from tsfc import compile_form

Vc = VectorElement(TensorProductElement(FiniteElement("DQ", quadrilateral, 1),
                                        FiniteElement("P", interval, 1)),
                   degree=1)

mesh = PeriodicUnitSquareMesh(10, 1, quadrilateral=True)
mesh = ExtrudedMesh(mesh, layers=1)

Vp = FunctionSpace(mesh, TensorProductElement(FiniteElement("DQ", quadrilateral, 1),
                                              FiniteElement("DP", interval, 1)))

Vu = FunctionSpace(mesh,
                   EnrichedElement(HDivElement(TensorProductElement(FiniteElement("RTCF", quadrilateral, 2),
                                                                    FiniteElement("DP", interval, 1))),
                                   HDivElement(TensorProductElement(FiniteElement("DQ", quadrilateral, 1),
                                                                    FiniteElement("P", interval, 2)))))

W = FunctionSpace(mesh, MixedElement(Vu.ufl_element(), Vp.ufl_element()))

w, phi = TestFunctions(W)
u, p = TrialFunctions(W)

# L = Constant(mesh)

Ap = (inner(w, u) + div(w)*div(u) + phi*p)*dx(metadata={"mode": "coffee"})

forms = split_form(Ap)

A00 = forms[0].form
A01 = forms[1].form
A10 = forms[2].form
A11 = forms[3].form
k, = compile_form(A00)
```

wence-:
[...]

PR #136, opened 2017-07-28
Cutoff ilp

wence-:
Does two things:

1. Adds cut off to ILP so we don't do too much.  Set arbitrarily at 16, so we don't do more than 2^16 things.

2. Attacks the major constant factor in the ILP program.

Running the test from #135, we have:

```sh
# With 1. (But basically "before", because there are 16 factors)
$ python many_monos.py 
COFFEE:PERF_OK COFFEE finished in 0.00246692 seconds (flops: 96 -> 96)
tsfc:INFO compute_form_data finished in 0.173406 seconds.
tsfc:INFO compile_integral finished in 12.517 seconds.
tsfc:INFO TSFC finished in 12.6912 seconds.

# With 2 (lipstick on the pig)
$ python many_monos.py 
COFFEE:PERF_OK COFFEE finished in 0.00254011 seconds (flops: 96 -> 96)
tsfc:INFO compute_form_data finished in 0.163301 seconds.
tsfc:INFO compile_integral finished in 2.17711 seconds.
tsfc:INFO TSFC finished in 2.3412 seconds.
```

### DIFF
--- a/tsfc/coffee_mode.py
+++ b/tsfc/coffee_mode.py
@@ -12,6 +12,7 @@ from gem.node import traversal
 from gem.gem import IndexSum, Failure, Sum, one
 from gem.utils import groupby
 
+from tsfc.logging import logger
 import tsfc.spectral as spectral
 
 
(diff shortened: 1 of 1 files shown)


miklos1:
If this is urgent, feel free to merge, because a poor solution in sensible time is better than waiting forever for the optimal solution. However, what I don't like about this cutoff, is that only the first 16 atomics are ever considered; if say, there is no feasible solution using only them, then you are stuck with the pessimal (but always feasible) solution.

PR #137, opened 2017-07-28
Rewrite ILP algorithm employed in coffee mode

### LINKED ISSUE #135: Some gusto tests now take forever to optimise kernels

miklos1:
New version is believed to be much faster, and sensibly interruptible, however, this has not been thoroughly tested yet.

Fixes #135, closes #136.

### DIFF
--- a/tsfc/coffee_mode.py
+++ b/tsfc/coffee_mode.py
@@ -1,5 +1,5 @@
 from __future__ import absolute_import, print_function, division
-from six.moves import map, zip
+from six.moves import map, range, zip
 
 import numpy
 import itertools
@@ -11,6 +11,8 @@ from gem.unconcatenate import unconcatenate
 from gem.node import traversal
 from gem.gem import IndexSum, Failure, Sum, one
 from gem.utils import groupby
+
+from tsfc.logging import logger
 
 import tsfc.spectral as spectral
 
@@ -64,17 +66,6 @@ def optimise_expressions(expressions, argument_indices):
     return [optimise_monomial_sum(ms, argument_indices) for ms in monomial_sums]
 
 
-def index_extent(factor, argument_indices):
-    """Compute the product of the extents of argument indices of a GEM expression
-
-    :arg factor: GEM expression
-    :arg argument_indices: set of argument indices
-
-    :returns: product of extents of argument indices
-    """
-    return numpy.product([i.extent for i in set(factor.free_indices).intersection(argument_indices)])
-
-
 def monomial_sum_to_expression(monomial_sum):
     """Convert a monomial sum to a GEM expression.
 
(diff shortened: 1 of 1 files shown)


tsfc/coffee_mode.py
```diff
@@ -93,69 +84,63 @@ def monomial_sum_to_expression(monomial_sum):
...
-    return tuple(atomic for atomic in atomics if atomic in optimal_atomics)
+    optimal_solution = set(atomics)  # pessimal but feasible solution
+    solution = set()
+
+    max_it = 1 << 12
+    it = iter(range(max_it))
+
+    def solve(idx):
+        while idx < len(monomials) and solution.intersection(monomials[idx].atomics):
+            idx += 1
```
wence-: OK, find the first monomial that doesn't intersect with the current solution.

tsfc/coffee_mode.py
```diff
@@ -93,69 +84,63 @@ def monomial_sum_to_expression(monomial_sum):
...
+
+        if idx < len(monomials):
+            if len(solution) < len(optimal_solution):
```
wence-: If we haven't intersected with all monomials less and the current solution is shorter than the optimal, then let's try.

tsfc/coffee_mode.py
```diff
@@ -93,69 +84,63 @@ def monomial_sum_to_expression(monomial_sum):
...
+                for atomic in monomials[idx].atomics:
+                    solution.add(atomic)
+                    solve(idx + 1)
+                    solution.remove(atomic)
```
wence-: Try each atomic in the current monomial in turn.

tsfc/coffee_mode.py
```diff
@@ -93,69 +84,63 @@ def monomial_sum_to_expression(monomial_sum):
...
+        else:
+            if cost(solution) < cost(optimal_solution):
+                optimal_solution.clear()
+                optimal_solution.update(solution)
```
wence-: We reached the end of the list of monomials so we have a feasible solution.  If it is cheaper than the current one, then accept.

tsfc/coffee_mode.py
```diff
@@ -93,69 +84,63 @@ def monomial_sum_to_expression(monomial_sum):
...
+            next(it)
```
wence-: Keep track of the number of attempts.

miklos1:
@wence-: do you want to test it with the Gusto test suite?

wence-:
OK.

wence-:
Seems good.

PR #138, opened 2017-07-28
Register Luporini 2016 citation in coffee mode

wence-:


wence-:
Also adds the tsfc preprint (unconditionally).

PR #139, opened 2017-08-02
Use FInAT mixed element and update UFC kernel interface

miklos1:
These changes mostly affect FEniCS, though there are some common refactorings, such as `create_element` is now accessed through the kernel interface, because Firedrake and FEniCS need different DoF ordering when translating UFL `VectorElement`/`TensorElement`. Similarly, the common `tsfc.finatinterface.create_element` gets an optional argument to choose between these two translations.

List of FEniCS-specific changes:
- `MixedElement` now comes from FInAT/FInAT#37, so mixed arguments don't lose structure. Essentially form splitting is applied in the form compiler using the existing element splitting infrastructure (#132).
- Coefficient splitting now longer applied for UFC, because FInAT mixed elements allow unhindered access to the underlying structure of the subelements, so the kernel interface is simplified.
- Vector/tensor DoF transposition trick is no longer applied for arguments and coefficients, however, it is now applied for coordinates, because in FEniCS coordinates XYZ XYZ XYZ XYZ ordering, while ordinary coefficients have XXXX YYYY ZZZZ.

### DIFF
--- a/gem/gem.py
+++ b/gem/gem.py
@@ -699,7 +699,7 @@ class Concatenate(Node):
 
     def __new__(cls, *children):
         if all(isinstance(child, Zero) for child in children):
-            size = sum(numpy.prod(child.shape, dtype=int) for child in children)
+            size = int(sum(numpy.prod(child.shape, dtype=int) for child in children))
             return Zero((size,))
 
         self = super(Concatenate, cls).__new__(cls)
@@ -708,7 +708,7 @@ class Concatenate(Node):
 
     @property
     def shape(self):
-        return (sum(numpy.prod(child.shape, dtype=int) for child in self.children),)
+        return (int(sum(numpy.prod(child.shape, dtype=int) for child in self.children)),)
 
 
 class Delta(Scalar, Terminal):
--- a/gem/gem.py
+++ b/gem/gem.py
@@ -697,8 +697,14 @@ class Concatenate(Node):
     """
     __slots__ = ('children',)
 
-    def __init__(self, *children):
+    def __new__(cls, *children):
+        if all(isinstance(child, Zero) for child in children):
+            size = sum(numpy.prod(child.shape, dtype=int) for child in children)
+            return Zero((size,))
+
+        self = super(Concatenate, cls).__new__(cls)
         self.children = children
+        return self
 
     @property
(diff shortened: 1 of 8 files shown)


tsfc/finatinterface.py
```diff
@@ -81,7 +81,7 @@ def fiat_compat(element):
 
 
 @singledispatch
-def convert(element):
+def convert(element, vector_transpose=False):
```
miklos1: Another poor name, not sure how shall I call this.

gem/gem.py
```diff
@@ -697,8 +697,14 @@ class Concatenate(Node):
     """
     __slots__ = ('children',)
 
-    def __init__(self, *children):
+    def __new__(cls, *children):
+        if all(isinstance(child, Zero) for child in children):
+            size = sum(numpy.prod(child.shape, dtype=int) for child in children)
```
wence-: This results in an `np.int64`, not sure if you care.

tsfc/ufl_utils.py
```diff
@@ -202,6 +202,9 @@ def modified_terminal(self, o):
 def split_coefficients(expression, split):
     """Split mixed coefficients, so mixed elements need not be
     implemented."""
+    if split is None:
```
wence-: docstring for `split` argument?

tsfc/finatinterface.py
wence-: shape_innermost?

tsfc/kernel_interface/ufc.py
```diff
@@ -248,8 +231,8 @@ def prepare_coordinates(coefficient, name, interior_facet=False):
                                qualifiers=["const"])]
         variable0 = gem.Variable(name+"_0", (size,))
         variable1 = gem.Variable(name+"_1", (size,))
-        expression = (gem.reshape(variable0, shape),
-                      gem.reshape(variable1, shape))
+        expression = (transpose(gem.reshape(variable0, transposed_shape)),
+                      transpose(gem.reshape(variable1, transposed_shape)))
```
wence-: OK, so previously you treated all coefficients specially except the coordinates.  Now you treat just the coordinates specially.

Issue #140, opened 2017-08-02
quadrature warning breaks with mix of tuple and integer degrees


stephankramer:
Still not entirely sure when this happens, but to reproduce check out the python3-fixes branch of thetis and run the test/bottomFriction/test_bottom_friction.py test. For that case `function_degrees` (line 161 of tsfc/driver.py) contains:

      [(1, 1), (1, 1), (1, 1), 0, 0, (1, 1), (1, 1), (1, 1), (1, 1)]

which gives an error evaluating the max() in line 167:

    TypeError: unorderable types: int() > tuple()

This only fails in python3 as python2 seems happy to compare integers with tuples (integers are always smaller than any tuple it seems).

miklos1:
Please test with branch `fix-issue-140`.

stephankramer:
With that fix it no longer crashes but the warning is still not correct. It gives a big backtrace for an exception handled by the logger because `quadrature_degree` and `max_degree(function_degree`) are tuples which don't match the format specifier `%d`:

    TypeError: %d format: a number is required, not tuple
    Call stack:
     ... \<massive-call-stack\> ...
    Message: 'Estimated quadrature degree %d more than tenfold greater than any argument/coefficient degree (max %d)'
    Arguments: ((11, 11), (1, 1))

miklos1:
Thanks, patch updated.

stephankramer:
Thanks, that fixes it

PR #141, opened 2017-08-02
Fix issue #140

miklos1:


tsfc/driver.py
```diff
@@ -164,7 +165,7 @@ def compile_integral(integral_data, form_data, prefix, parameters,
                 logger.warning("Estimated quadrature degree %d more "
                                "than tenfold greater than any "
                                "argument/coefficient degree (max %d)",
```
wence-: Need to use %s not %d to be safe for tuples.

Issue #112, opened 2017-05-03
Temporaries graph to drive argument factorisation


wence-:

connorjward:
Closing as there is insufficient information here to work from now that members have left the project.

wence-:
FWIW, this is the mechanism described in Fabio's thesis.

connorjward:
Ah that's very helpful thank you! I will port this over to Firedrake then so we don't lose this (and do the same for https://github.com/firedrakeproject/tsfc/issues/115).

PR #142, opened 2017-08-02
Delta cancellation and amalgamate spectral and coffee modes

### LINKED ISSUE #89: Optimise away Kronecker deltas for arguments

### LINKED ISSUE #116: Incorporate sum-factorisation into coffee mode optimisation pass
The `spectral` optimisation mode performs unconditional argument factorisation followed by sum factorisation.  We should incorporate sum factorisation technology into the coffee pass.  This involves (amongst other things) appropriate modification of the cost model.

miklos1:
Main news:
- Delta cancellation arrives in spectral mode, both for vector/tensor elements, and for GL/GLL type elements with the lucky quadrature rule. Resolves #89.
- Spectral mode now incorporates @tj-sun's the simplified COFFEE algorithm (#98), and becomes the default mode. (Resolves #116?)
- Coffee mode remains as it is for @tj-sun to finish porting the complete COFFEE algorithm with graph rewriting (#112).

Some details:
- A minor change to the COFFEE algorithm (see commit c67721a), which otherwise moves to GEM. It is now shared between the coffee mode and the spectral mode.
- To avoid a performance hit, `gem.optimise.delta_elimination` now injects `ComponentTensor` nodes whose removal is now the caller's responsibility.
- Spectral mode is mostly rewritten.
- Added test cases for underintegration.

### DIFF
--- a/gem/coffee.py
+++ b/gem/coffee.py
@@ -1,0 +1,202 @@
+"""This module contains an implementation of the COFFEE optimisation
+algorithm operating on a GEM representation.
+
+This file is NOT for code generation as a COFFEE AST.
+"""
+
+from __future__ import absolute_import, print_function, division
+from six.moves import map, range
+
+from collections import OrderedDict
+import itertools
+import logging
+
+import numpy
+
+from gem.gem import IndexSum, one
+from gem.optimise import make_sum, make_product
+from gem.refactorise import Monomial
+from gem.utils import groupby
+
+
+try:
+    from firedrake import Citations
+    Citations().register("Luporini2016")
+except ImportError:
+    pass
+
+
+__all__ = ['optimise_monomial_sum']
+
+
+def monomial_sum_to_expression(monomial_sum):
+    """Convert a monomial sum to a GEM expression.
+
+    :arg monomial_sum: an iterable of :class:`Monomial`s
+
+    :returns: GEM expression
+    """
+    indexsums = []  # The result is summation of indexsums
+    # Group monomials according to their sum indices
+    groups = groupby(monomial_sum, key=lambda m: frozenset(m.sum_indices))
+    # Create IndexSum's from each monomial group
+    for _, monomials in groups:
+        sum_indices = monomials[0].sum_indices
+        products = [make_product(monomial.atomics + (monomial.rest,)) for monomial in monomials]
+        indexsums.append(IndexSum(make_sum(products), sum_indices))
+    return make_sum(indexsums)
+
+
+def index_extent(factor, argument_indices):
+    """Compute the product of the extents of argument indices of a GEM expression
+
+    :arg factor: GEM expression
+    :arg argument_indices: set of argument indices
+
+    :returns: product of extents of argument indices
+    """
+    return numpy.prod([i.extent for i in factor.free_indices if i in argument_indices])
+
+
+def find_optimal_atomics(monomials, argument_indices):
+    """Find optimal atomic common subexpressions, which produce least number of
+    terms in the resultant IndexSum when factorised.
+
+    :arg monomials: A list of :class:`Monomial`s, all of which should have
+                    the same sum indices
+    :arg argument_indices: tuple of argument indices
+
+    :returns: list of atomic GEM expressions
+    """
+    atomics = tuple(OrderedDict.fromkeys(itertools.chain(*(monomial.atomics for monomial in monomials))))
+
+    def cost(solution):
+        extent = sum(map(lambda atomic: index_extent(atomic, argument_indices), solution))
+        # Prefer shorter solutions, but larger extents
+        return (len(solution), -extent)
+
+    optimal_solution = set(atomics)  # pessimal but feasible solution
+    solution = set()
+
+    max_it = 1 << 12
+    it = iter(range(max_it))
+
+    def solve(idx):
+        while idx < len(monomials) and solution.intersection(monomials[idx].atomics):
+            idx += 1
+
+        if idx < len(monomials):
+            if len(solution) < len(optimal_solution):
+                for atomic in monomials[idx].atomics:
+                    solution.add(atomic)
+                    solve(idx + 1)
+                    solution.remove(atomic)
+        else:
+            if cost(solution) < cost(optimal_solution):
+                optimal_solution.clear()
+                optimal_solution.update(solution)
+            next(it)
+
+    try:
+        solve(0)
+    except StopIteration:
+        logger = logging.getLogger('tsfc')
+        logger.warning("Solution to ILP problem may not be optimal: search "
+                       "interrupted after examining %d solutions.", max_it)
+
+    return tuple(atomic for atomic in atomics if atomic in optimal_solution)
+
+
+def factorise_atomics(monomials, optimal_atomics, argument_indices):
+    """Group and factorise monomials using a list of atomics as common
+    subexpressions. Create new monomials for each group and optimise them recursively.
+
+    :arg monomials: an iterable of :class:`Monomial`s, all of which should have
+                    the same sum indices
+    :arg optimal_atomics: list of tuples of atomics to be used as common subexpression
+    :arg argument_indices: tuple of argument indices
+
+    :returns: an iterable of :class:`Monomials`s after factorisation
+    """
+    if not optimal_atomics or len(monomials) <= 1:
+        return monomials
+
+    # Group monomials with respect to each optimal atomic
+    def group_key(monomial):
+        for oa in optimal_atomics:
+            if oa in monomial.atomics:
+                return oa
+        assert False, "Expect at least one optimal atomic per monomial."
+    factor_group = groupby(monomials, key=group_key)
+
+    # We should not drop monomials
+    assert sum(len(ms) for _, ms in factor_group) == len(monomials)
+
+    sum_indices = next(iter(monomials)).sum_indices
+    new_monomials = []
+    for oa, monomials in factor_group:
+        # Create new MonomialSum for the factorised out terms
+        sub_monomials = []
+        for monomial in monomials:
+            atomics = list(monomial.atomics)
+            atomics.remove(oa)  # remove common factor
+            sub_monomials.append(Monomial((), tuple(atomics), monomial.rest))
+        # Continue to factorise the remaining expression
+        sub_monomials = optimise_monomials(sub_monomials, argument_indices)
+        if len(sub_monomials) == 1:
+            # Factorised part is a product, we add back the common atomics then
+            # add to new MonomialSum directly rather than forming a product node
+            # Retaining the monomial structure enables applying associativity
+            # when forming GEM nodes later.
+            sub_monomial, = sub_monomials
+            new_monomials.append(
+                Monomial(sum_indices, (oa,) + sub_monomial.atomics, sub_monomial.rest))
+        else:
+            # Factorised part is a summation, we need to create a new GEM node
+            # and multiply with the common factor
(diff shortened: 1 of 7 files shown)


gem/coffee.py
```diff
@@ -0,0 +1,202 @@
...
+from gem.optimise import make_sum, make_product
+from gem.refactorise import Monomial
+from gem.utils import groupby
+
+
+try:
+    from firedrake import Citations
+    Citations().register("Luporini2016")
+except ImportError:
+    pass
```
miklos1: So this module is only imported from `tsfc.coffee` and `tsfc.spectral`, which are in turn only imported on demand, when the particular mode is used.

gem/coffee.py
```diff
@@ -0,0 +1,202 @@
...
+    :returns: an iterable of factorised :class:`Monomials`s
+    """
+    assert len(set(frozenset(m.sum_indices) for m in monomials)) <= 1,\
+        "All monomials required to have same sum indices for factorisation"
+
+    result = [m for m in monomials if not m.atomics]  # skipped monomials
+    active_monomials = [m for m in monomials if m.atomics]
+    optimal_atomics = find_optimal_atomics(active_monomials, linear_indices)
+    result += factorise_atomics(active_monomials, optimal_atomics, linear_indices)
+    return result
```
wence-: This is just code-movement.

gem/optimise.py
```diff
@@ -242,6 +242,15 @@ def delta_elimination(sum_indices, factors):
     """
     sum_indices = list(sum_indices)  # copy for modification
 
+    def substitute(expression, from_, to_):
+        if from_ not in expression.free_indices:
+            return expression
+        elif isinstance(expression, Delta):
+            mapper = MemoizerArg(filtered_replace_indices)
+            return mapper(expression, ((from_, to_),))
```
wence-: OK, so the delta must have `from_` as a free index, so replace with `to_`, and at the same time, do a little simplification.

tsfc/spectral.py
```diff
@@ -44,14 +32,86 @@ def Integrals(expressions, quadrature_multiindex, argument_multiindices, paramet
 
     :returns: list of integral representations
     """
+    # Rewrite: a / b => a * (1 / b)
+    expressions = replace_division(expressions)
```
wence-: This continues to feel like cargo-culting to me.  Maybe @FabioLuporini can comment, but is it really the case that compilers do a better job when given:

```c

A = B * (1/C);
```
?

tsfc/spectral.py
```diff
@@ -44,14 +32,86 @@ def Integrals(expressions, quadrature_multiindex, argument_multiindices, paramet
...
     if max_extent:
         def predicate(index):
             return index.extent <= max_extent
         expressions = unroll_indexsum(expressions, predicate=predicate)
-    # Integral representation: just a GEM expression
-    return replace_division([index_sum(e, quadrature_multiindex) for e in expressions])
+
+    expressions = [index_sum(e, quadrature_multiindex) for e in expressions]
+    argument_indices = tuple(chain(*argument_multiindices))
+    return [Integral(e, quadrature_multiindex, argument_indices) for e in expressions]
```
wence-: OK, so now the representation is no longer just `gem`.

tsfc/spectral.py
```diff
@@ -44,14 +32,86 @@ def Integrals(expressions, quadrature_multiindex, argument_multiindices, paramet
...
+        # Argument factorise expressions
+        classifier = partial(classify, set(free_indices))
+        monomial_sums = collect_monomials(expressions, classifier)
+        # For each monomial, apply delta cancellation and insert
+        # result into delta_simplified.
+        for variable, monomial_sum in zip(variables, monomial_sums):
+            for monomial in monomial_sum:
+                var, s, a, r = delta_elimination(variable, *monomial)
+                narrow_variables.setdefault(var)
+                delta_simplified[var].add(s, a, r)
```
wence-: Because there is no `OrderedDefaultDict`.

tsfc/spectral.py
```diff
@@ -68,28 +128,58 @@ def classify(argument_indices, expression):
...
+    """IndexSum-Delta cancellation for monomials."""
+    factors = list(args) + [variable, rest]  # construct factors
 
-    def group_key(assignment):
-        variable, expression = assignment
-        return variable.free_indices
+    def prune(factors):
+        result = remove_componenttensors(factors[:-1])
+        result.append(factors[-1])
+        return result
```
wence-: Is it obvious why you don't want to remove component tensors from the last factor?

tsfc/spectral.py
```diff
@@ -68,28 +128,58 @@ def classify(argument_indices, expression):
...
+            # share free indices with the tail ordering.
+            #
+            # Based on this, split the monomial sum, then recursively
+            # optimise each sub monomial sum with the first tail index
+            # removed.
+            tail_indices = tuple(i for i in sum_indices if i in tail_ordering)
+            tail_atomics = tuple(a for a in atomics
+                                 if set(tail_indices) & set(a.free_indices))
+            head_indices = tuple(i for i in sum_indices if i not in tail_ordering)
+            head_atomics = tuple(a for a in atomics if a not in tail_atomics)
```
wence-: Again because `OrderedSet` is not a thing.

tsfc/spectral.py
```diff
@@ -68,28 +128,58 @@ def classify(argument_indices, expression):
...
+            key = (head_indices, head_atomics)
+            key_ordering.setdefault(key)
+            sub_monosums[key].add(tail_indices, tail_atomics, rest)
+        sub_monosums = [(k, sub_monosums[k]) for k in key_ordering]
+
+        monomial_sum = MonomialSum()
+        for (sum_indices, atomics), monosum in sub_monosums:
+            new_rest = sum_factorise(variable, tail_ordering[1:], monosum)
+            monomial_sum.add(sum_indices, atomics, new_rest)
```
wence-: OK, so this is more "loop invariant code motion".

tsfc/spectral.py
FabioLuporini: As far as I remember, yes when 'b' is a constant

FabioLuporini: the ``b`` I'm referring to above is the one in the code, in ``a / b``

wence-: Ah, I see.

wence-: OK, I see.  A little bit of experimentation, using GCC 7.1:

```c
#define VEC 4

void kernel(double * __restrict__ A, double *  C, double *  B, double D)
{
    for (int j = 0; j < 128; j++) {
        for (int k = 0; k < VEC; k++) {
            A[j*VEC + k] += C[j*VEC + k]*B[j*VEC + k] / D;
        }
    }
}
```

If I compile with `-O3 -march=broadwell` I get:

```asm
kernel(double*, double*, double*, double):
  vbroadcastsd ymm0, xmm0
  xor eax, eax
.L2:
  vmovupd ymm1, YMMWORD PTR [rsi+rax]
  vmulpd ymm1, ymm1, YMMWORD PTR [rdx+rax]
  vdivpd ymm1, ymm1, ymm0
  vaddpd ymm1, ymm1, YMMWORD PTR [rdi+rax]
  vmovupd YMMWORD PTR [rdi+rax], ymm1
  add rax, 32
  cmp rax, 4096
  jne .L2
  vzeroupper
  ret
```

So B is broadcast across the vector lanes out of the loop, but then the division happens in the loop.

If I add `-ffast-math` (meaning that the optimisations don't have to maintain IEEE compliance, which, amongst other things allows the compiler to make the substitution `a/b => a*(1/b)` then I get:

```asm
kernel(double*, double*, double*, double):
  vmovsd xmm1, QWORD PTR .LC0[rip]
  xor eax, eax
  vdivsd xmm1, xmm1, xmm0
  vbroadcastsd ymm1, xmm1
.L2:
  vmovupd ymm0, YMMWORD PTR [rsi+rax]
  vmulpd ymm0, ymm0, YMMWORD PTR [rdx+rax]
  vfmadd213pd ymm0, ymm1, YMMWORD PTR [rdi+rax]
  vmovupd YMMWORD PTR [rdi+rax], ymm0
  add rax, 32
  cmp rax, 4096
  jne .L2
  vzeroupper
  ret
.LC0:
  .long 0
  .long 1072693248
```

So we perform the division out of the loop, and inside it's turned into a multiplication.

If I manually replace ` / D => * (1/D)` then I get

```asm
kernel(double*, double*, double*, double):
  vmovsd xmm1, QWORD PTR .LC0[rip]
  xor eax, eax
  vdivsd xmm1, xmm1, xmm0
  vbroadcastsd ymm1, xmm1
.L2:
  vmovupd ymm0, YMMWORD PTR [rsi+rax]
  vmulpd ymm0, ymm0, YMMWORD PTR [rdx+rax]
  vfmadd213pd ymm0, ymm1, YMMWORD PTR [rdi+rax]
  vmovupd YMMWORD PTR [rdi+rax], ymm0
  add rax, 32
  cmp rax, 4096
  jne .L2
  vzeroupper
  ret
.LC0:
  .long 0
  .long 1072693248
```

Which is as if I'd said `-ffast-math`.

We currently don't put `-ffast-math` in the compiler flags, which I think is a reasonable position.

gem/coffee.py
miklos1: This very end has actually changed (c67721a), but otherwise yet.

gem/optimise.py
miklos1: Yes, because if we allow to wrap `Delta` nodes with `Indexed(ComponentTensor(..., ...), ...)` then `isinstance(f, Delta)` will fail in the next round of the fixed point iteration looking for delta cancellation opportunities. See the test case.

tsfc/spectral.py
miklos1: As far as I remember @tj-sun porting the old COFFEE code, the logic was something like this: apply this rewriting if `b` is a number or a `Symbol`. A `Symbol` could either mean a constant or a variable reference. Generally the denominator was assigned to a temporary if it was used more than once. That is, instead of dividing `n` times with the same number, the substitution makes 1 division and `n` multiplication. Most common use case is the calculation of the inverse Jacobian, where the determinant is the denominator.

tsfc/spectral.py
miklos1: During argument factorisation, the "atomics" are subexpressions of FInAT-provided tabulations, which means they are small so this index substitution is cheap. Moreover, `MonomialSum` collapses monomials with the same set of sum indices and atomics. Layers of `Indexed(ComponentTensor(..., ...), ...)` would atomics that are actually the same to look different, and thus would hinder the shortening of the `MonomialSum`.

The `rest` part, which is the last factor if you look at the ordering above, is the opposite of what is described above. It can be arbitrarily big, making `remove_componenttensors` costly, and there are not benefits to early removal of those `ComponentTensor`s.

Perhaps I should write some if this into a comment?

tsfc/spectral.py
wence-: FWIW, clang won't generate FMAs unless you say either `-ffp-contract=fast` or `-ffast-math`.

tsfc/spectral.py
wence-: Sounds like a good idea.

wence-:
> Spectral mode now incorporates @tj-sun's the simplified COFFEE algorithm (#98), and becomes the default mode. (Resolves #116?)

I think so, given the what is current "COFFEE" in tsfc.

PR #143, opened 2017-08-03
Handle Failure nodes in the interpreter

wence-:
Map to an appropriately shaped array of NaNs.  Necessary if anyone
ever asks for entity_support_dofs on a trace element.


wence-:
FInaT/FInaT#38 broke Firedrake because we now ask for the full entity_support_dofs which results in Failure nodes needing to be handled by the interpreter for Trace spaces.

PR #147, opened 2017-08-10
Revise FInAT interface

miklos1:
This pull request must be merged at the same time with FInAT/FInAT#39 which wraps a more complete set of FIAT elements. At the same FInAT interface will not have a generic fallback on the FIAT interface, such fallback only happens explicitly for `RestrictedElement`s and should happen for `NodalEnrichedElement`s (currently not translated) as well.

At the same time, we also revise UFL -> FInAT element conversion in anticipation of the changes required to use with runtime tabulated elements in Themis. This patch does not need dynamic scoping.

### DIFF
--- a/tsfc/finatinterface.py
+++ b/tsfc/finatinterface.py
@@ -22,6 +22,7 @@
 # along with FFC. If not, see <http://www.gnu.org/licenses/>.
 
 from __future__ import absolute_import, print_function, division
+from six import iteritems
 
 from singledispatch import singledispatch
 import weakref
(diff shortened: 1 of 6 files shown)


tsfc/finatinterface.py
```diff
@@ -134,87 +128,119 @@ def convert_finiteelement(element, shape_innermost=True):
...
-    scalar_element = create_element(element.sub_elements()[0], shape_innermost)
-    return finat.TensorFiniteElement(scalar_element,
-                                     element.reference_value_shape(),
-                                     transpose=not shape_innermost)
+def convert_tensorelement(element, **kwargs):
+    scalar_elem, deps = _create_element(element.sub_elements()[0], **kwargs)
+    shape = element.reference_value_shape()
+    shape_innermost = kwargs["shape_innermost"]
+    return (finat.TensorFiniteElement(scalar_elem, shape, not shape_innermost),
+            deps | {"shape_innermost"})
```
wence-: OK, so `deps` records which `kwargs` affected element conversion, and then the cache key records `name, value` pairs.

PR #148, opened 2017-08-11
Less CI testing with Python 2.7

miklos1:
Since Python 2.7 is no longer the primary target for TSFC, a single job in the build matrix should be enough for testing it.

@wence-, @tkarna: In light of #17, I wonder if we should still test with old numpy, but do so with Python 3 instead.


miklos1:
@wence-: The compat code has been removed a long time ago, because it is no longer used.

wence-:
Oh, OK, in which case even fewer issues...

PR #151, opened 2017-09-07
Make UFL -> GEM aware of physical geometry

### LINKED ISSUE #150: Make coordinates special during UFL to GEM translation
Currently, the UFL to GEM translation implemented in `tsfc.fem.compile_ufl` has no clue about coordinates and physical geometry. This is achieved through a preliminary preprocessing step that replaces instances `SpatialCoordinate(domain)` with `ReferenceValue(Coefficient(...))`.

There are several signs that this approach is not satisfactory:
- Last week the implementation of `CellOrigin` came up again, and this issue is the reason we cannot do that easily.
- I realised that the situation of `CellVolume` and `FacetArea` are actually quite analogous to `CellOrigin`. However, they are currently implemented, by spoon-feeding `compile_ufl` thunks that generate GEM for cell volume and facet area on demand, and these thunks are provided the coordinates.
- The plan for implementing @rckirby's https://arxiv.org/abs/1706.09017 requires coordinates (physical geometry) to be available when asking FInAT for tabulating basis functions.

Idea:
- The kernel interface expected at the UFL to GEM translation (`tsfc.kernel_interface.__init__.py`) to be extended to implement a mapping from a UFL domain to a UFL coordinate `Coefficient`.
This API in principle supports multi-domain forms.
- The Firedrake kernel interface could be Firedrake-specific without fear, so `SpatialCoordinate(mesh)` could be replaced with `mesh.coordinates`.
- Eliminate that preprocessing step that replaces `SpatialCoordinate`s. Instead, replace them on the spot with different UFL using the API defined above, and then translate that replaced UFL to GEM.
- This setup should make it easy to implement `CellOrigin`. Similarly, `CellVolume` and `FacetArea` should move to `tsfc.fem` from `tsfc.driver`, slightly simplifying the different `compile_integral`-like codes in the broader Firedrake codebase.
- It should become relatively easy to implement the geometric mapping API required by FInAT for @rckirby's special elements.

miklos1:
Resolves #150. Adds support for UFL node `CellOrigin`.

Requires FInAT/FInAT#41 and some Firedrake changes.

### DIFF
--- a/tsfc/fem.py
+++ b/tsfc/fem.py
@@ -18,7 +18,8 @@ from ufl.classes import (Argument, CellCoordinate, CellEdgeVectors,
                          CellVolume, Coefficient, FacetArea,
                          FacetCoordinate, GeometricQuantity,
                          QuadratureWeight, ReferenceCellVolume,
-                         ReferenceFacetVolume, ReferenceNormal)
+                         ReferenceFacetVolume, ReferenceNormal,
+                         SpatialCoordinate)
 
 from FIAT.reference_element import make_affine_mapping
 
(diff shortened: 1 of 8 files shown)


tsfc/fem.py
```diff
@@ -153,12 +164,17 @@ def basis_evaluation(self, finat_element, local_derivatives, entity_id):
...
 
 class Translator(MultiFunction, ModifiedTerminalMixin, ufl2gem.Mixin):
-    """Contains all the context necessary to translate UFL into GEM."""
+    """Multifunction for translating UFL -> GEM.  Incorporates ufl2gem.Mixin, and
+    dispatches on terminal type when reaching modified terminals."""
 
     def __init__(self, context):
+        # MultiFunction.__init__ does not call further __init__
+        # methods, but ufl2gem.Mixin must be initialised.
+        # (ModifiedTerminalMixin requires no initialisation.)
```
wence-: I guess this is working around `MultiFunction.__init__` not calling super?

miklos1: Well, yes, we have to do it the non-`super` way, calling it for each base class manually.

PR #153, opened 2017-10-12
Complex

dham:
This is a rebased version of Nic's complex branch. This probably doesn't want to land until we're ready to land ufl and firedrake  too.


miklos1:
3 lines changed per commit is too low. Can we have a shorter history?

gem/gem.py
```diff
@@ -108,7 +108,7 @@ class Constant(Terminal):
 
     Convention:
      - array: numpy array of values
-     - value: float value (scalars only)
+     - value: complex value (scalars only)
```
miklos1: Is it always complex?

gem/gem.py
```diff
@@ -161,7 +161,7 @@ def __new__(cls, array):
             return super(Literal, cls).__new__(cls)
 
     def __init__(self, array):
-        self.array = asarray(array, dtype=float)
+        self.array = asarray(array, dtype=float) if asarray(array).dtype is numpy.dtype('int') else asarray(array)
```
miklos1: What is the reason for this?

gem/gem.py
```diff
@@ -175,7 +175,7 @@ def get_hash(self):
 
     @property
     def value(self):
-        return float(self.array)
+        return self.array
```
miklos1: This is wrong. `value` should give back a scalar, not an n-dimensional array.

gem/gem.py
```diff
@@ -289,12 +289,18 @@ class MathFunction(Scalar):
...
     __front__ = ('name',)
 
-    def __init__(self, name, *args):
+    def __new__(cls, name, *args):
         assert isinstance(name, str)
         assert all(arg.shape == () for arg in args)
 
+        if name in {'conj', 'real', 'imag'}:
+            if isinstance(args[0], Zero):
+                return args[0]
```
miklos1: No magic numbers, please.

gem/interpreter.py
```diff
@@ -172,6 +172,28 @@ def _evaluate_operator(e, self):
     return result
 
 
+@_evaluate.register(gem.MathFunction)  # noqa: F811
```
miklos1: `noqa` outdated

gem/interpreter.py
```diff
@@ -172,6 +172,28 @@ def _evaluate_operator(e, self):
...
+            result[idx] = [o[o.filter(idx,result.fids)].imag for o in ops]
+    elif e.name is 'conj':
+        for idx in numpy.ndindex(result.tshape):
+            result[idx] = [o[o.filter(idx,result.fids)].conjugate() for o in ops]
+    elif e.name is 'real':
+        for idx in numpy.ndindex(result.tshape):
+            result[idx] = [o[o.filter(idx,result.fids)].real for o in ops]
+    else:
+        for idx in numpy.ndindex(result.tshape):
+            result[idx] = op(*(o[o.filter(idx, result.fids)] for o in ops))
```
miklos1: So this is not performance critical? It seems to be done the slow way rather than the fast way (see `_evaluate_operator`).

miklos1: Also it's ugly that the loop is duplicated several times with very tiny changes to each.

tsfc/driver.py
```diff
@@ -80,6 +84,9 @@ def compile_integral(integral_data, form_data, prefix, parameters,
         _.update(parameters)
         parameters = _
 
+    # Set the scalar type
+    set_scalar_type(parameters["scalar_type"])
```
miklos1: This is horrible...

tsfc/driver.py
```diff
@@ -238,7 +245,8 @@ def name_multiindex(multiindex, name):
         name_multiindex(multiindex, name)
 
     # Construct kernel
-    body = generate_coffee(impero_c, index_names, parameters["precision"], expressions, split_argument_indices)
+    body = generate_coffee(impero_c, index_names, parameters["precision"], 
+                            parameters["scalar_type"], expressions, split_argument_indices)
```
miklos1: flake8

tsfc/driver.py
```diff
@@ -266,8 +274,14 @@ def compile_expression_at_points(expression, points, coordinates, parameters=Non
     if extract_arguments(expression):
         return ValueError("Cannot interpolate UFL expression with Arguments!")
 
+    # Set the scalar type
+    set_scalar_type(parameters["scalar_type"])
```
miklos1: Still horrible...

tsfc/driver.py
```diff
@@ -46,8 +46,12 @@ def compile_form(form, prefix="form", parameters=None):
     cpu_time = time.time()
 
     assert isinstance(form, Form)
-
-    fd = ufl_utils.compute_form_data(form)
+    
+    # determine if we're in complex mode; coffee mode does not support complex
+    complx = parameters and "scalar_type" in parameters and parameters["scalar_type"] == 'double complex'
```
miklos1: ugly variable name

tsfc/driver.py
```diff
@@ -305,12 +319,12 @@ def compile_expression_at_points(expression, points, coordinates, parameters=Non
     return_shape = (len(points),) + value_shape
     return_indices = point_set.indices + tensor_indices
     return_var = gem.Variable('A', return_shape)
-    return_arg = ast.Decl(SCALAR_TYPE, ast.Symbol('A', rank=return_shape))
+    return_arg = ast.Decl(scalar_type(), ast.Symbol('A', rank=return_shape))
```
miklos1: Why is `scalar_type()` any better than `SCALAR_TYPE`?

miklos1: x4 for `tsfc/kernel_interface/firedrake.py`

miklos1: x5 for `tsfc/kernel_interface/ufc.py`

tsfc/ufl2gem.py
```diff
@@ -46,6 +46,27 @@ def product(self, o, *ops):
...
+        else:
+            return MathFunction('imag', expr)
+
+    def conj(self, o, expr):
+        if o.ufl_shape:
+            indices = tuple(Index() for i in range(len(o.ufl_shape)))
+            return ComponentTensor(MathFunction('conj', Indexed(expr, indices)), indices)
+        else:
+            return MathFunction('conj', expr)
+
```
miklos1: :+1:

tsfc/ufl_utils.py
```diff
@@ -79,19 +83,27 @@ def one_times(measure):
...
+def preprocess_expression(expression, complex_mode=False):
     """Imitates the compute_form_data processing pipeline.
 
     Useful, for example, to preprocess non-scalar expressions, which
     are not and cannot be forms.
     """
+    if complex_mode:
+        expression = do_comparison_check(expression)
+    if not complex_mode:
+        expression = remove_complex_nodes(expression)
```
miklos1: This is odd way of writing `else`

tsfc/ufl_utils.py
```diff
@@ -79,19 +83,27 @@ def one_times(measure):
...
     expression = apply_derivatives(expression)
     expression = apply_function_pullbacks(expression)
     expression = apply_geometry_lowering(expression, preserve_geometry_types)
     expression = apply_derivatives(expression)
     expression = apply_geometry_lowering(expression, preserve_geometry_types)
     expression = apply_derivatives(expression)
+    if not complex_mode:
+        expression = remove_complex_nodes(expression)
+    if complex_mode:
+        expression = optimise_complex_nodes(expression)
```
miklos1: This is an odd way of writing `else`

tsfc/ufl_utils.py
```diff
@@ -79,19 +83,27 @@ def one_times(measure):
...
+        expression = remove_complex_nodes(expression)
     expression = apply_algebra_lowering(expression)
     expression = apply_derivatives(expression)
     expression = apply_function_pullbacks(expression)
     expression = apply_geometry_lowering(expression, preserve_geometry_types)
     expression = apply_derivatives(expression)
     expression = apply_geometry_lowering(expression, preserve_geometry_types)
     expression = apply_derivatives(expression)
+    if not complex_mode:
+        expression = remove_complex_nodes(expression)
```
miklos1: Why are we removing complex nodes twice?

gem/gem.py
wence-: Explanation: `conj`/`real`/`imag` have a single operand (presumably), so write:

```python
arg, = args
if isinstance(arg, Zero):
    return arg
```

tsfc/ufl_utils.py
dham: I think it's because intermediate steps can add new complex nodes.

miklos1: What about doing just the second removal then?

gem/gem.py
dham: It looks a bit odd, but I think Nic is trying to cast integers to floats without casting complex to float.

tsfc/ufl_utils.py
dham: I think Nic was concerned that complex nodes might disable simplifications that might otherwise occur.

gem/gem.py
wence-:
```
try:
   array = asarray(array, dtype=float)
except TypeError:
   array = asarray(array, dtype=complex)
```

gem/interpreter.py
```diff
@@ -172,6 +172,28 @@ def _evaluate_operator(e, self):
     return result
 
 
+@_evaluate.register(gem.MathFunction)  # noqa: F811
+def _(e, self):
```
wence-: avoid the need for noqa by writing `def _evaluate_mathfunction(e, self)`

gem/interpreter.py
```diff
@@ -172,6 +172,28 @@ def _evaluate_operator(e, self):
...
+        for idx in numpy.ndindex(result.tshape):
+            result[idx] = [o[o.filter(idx,result.fids)].real for o in ops]
+    else:
+        for idx in numpy.ndindex(result.tshape):
+            result[idx] = op(*(o[o.filter(idx, result.fids)] for o in ops))
+    return result
+
+
 @_evaluate.register(gem.MathFunction)
 def _evaluate_mathfunction(e, self):
```
wence-: There are now two handlers for `gem.MathFunction`...

gem/interpreter.py
wence-: AKA:
```
ops = [self(o) for o in e.children]
result = Result.empty(*ops)
op = {"abs": numpy.ndarray.abs,
                "log": numpy.ndarray.log,
                "imag": operator.attrgetter("imag"),
                "real": operator.attrgetter("real"),
                "conf": numpy.ndarray.conjugate}[e.name]
fids = result.fids
result.arr = op(*[o.broadcast(fids) for o in ops])
return result
```

I think.

wence-: Maybe `numpy.concatenate([op(o.broadcast(fids)) for o in ops])`

tsfc/coffee.py
```diff
@@ -257,7 +260,18 @@ def _expression_mathfunction(expr, parameters):
         'cyl_bessel_i': 'boost::math::cyl_bessel_i',
         'cyl_bessel_k': 'boost::math::cyl_bessel_k',
     }
-    name = name_map.get(expr.name, expr.name)
+    complex_name_map = {
+        'ln': 'clog',
+        'conj': 'conj'
+        # TODO: Are there different complex Bessel Functions?
```
wence-: The standard bessel functions are defined over the field of complex numbers.  But boost for one does not actually implement them.

tsfc/coffee.py
```diff
@@ -320,11 +334,20 @@ def _expression_scalar(expr, parameters):
     if isnan(expr.value):
         return coffee.Symbol("NAN")
     else:
-        v = expr.value
-        r = round(v, 1)
-        if r and abs(v - r) < parameters.epsilon:
-            v = r  # round to nonzero
-        return coffee.Symbol(("%%.%dg" % parameters.precision) % v)
+        vr = expr.value.real
+        rr = round(1)
```
miklos1: This looks wrong... should probably be `rr = round(vr, 1)`.

tsfc/coffee.py
```diff
@@ -320,11 +334,20 @@ def _expression_scalar(expr, parameters):
...
+        if rr and abs(vr - rr) < parameters.epsilon:
+            vr = rr  # round to nonzero
+
+        vi = expr.value.imag  # also checks if v is purely real
+        if vi == 0.0:
+            return coffee.Symbol(("%%.%dg" % parameters.precision) % vr)
+        ri = round(1)
```
miklos1: Ditto

tsfc/coffee.py
```diff
@@ -320,11 +334,20 @@ def _expression_scalar(expr, parameters):
...
+
+        if ri and abs(vi - ri) < parameters.epsilon:
+            vi = ri
+        return coffee.Symbol(("%%.%dg" % parameters.precision) % vr + " + " +
+                             ("%%.%dg" % parameters.precision) % vi + " * I")
```
miklos1: Looks slightly dangerous... COFFEE probably does not surround symbols with parenthesis, which can lead to a precedence issue.

tsfc/driver.py
```diff
@@ -43,7 +43,11 @@ def compile_form(form, prefix="form", parameters=None):
 
     assert isinstance(form, Form)
 
-    fd = ufl_utils.compute_form_data(form)
+    # determine if we're in complex mode; coffee mode does not support complex
+    complx = parameters and "scalar_type" in parameters and 'complex' in parameters["scalar_type"]
+    if complx:
+        parameters["mode"] = 'vanilla'
```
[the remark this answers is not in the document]
miklos1: I suspect the current code for refactorisation cannot apply the distributivity rule over linear operators such as `conj`, `real`, or `imag`.

tsfc/parameters.py
```diff
@@ -20,10 +14,18 @@
     # that makes compilation time much shorter.
     "unroll_indexsum": 3,
 
+    "scalar_type": "double",
     # Precision of float printing (number of digits)
-    "precision": numpy.finfo(NUMPY_TYPE).precision,
+    "precision": numpy.finfo(numpy.dtype("double")).precision,
+
+    "scalar_type": "double"
```
miklos1: scalar_type listed twice

gem/interpreter.py
```diff
@@ -176,8 +176,18 @@ def _evaluate_mathfunction(e, self):
     names = {"abs": abs,
              "log": math.log}
```
miklos1: We probably need a separate code path for `clog`

gem/interpreter.py
```diff
@@ -176,8 +176,18 @@ def _evaluate_mathfunction(e, self):
...
     op = names[e.name]
-    for idx in numpy.ndindex(result.tshape):
-        result[idx] = op(*(o[o.filter(idx, result.fids)] for o in ops))
+    if e.name is 'imag':
```
miklos1: How could this work? If `e.name` is `'imag'` then line 178 (`op = names[e.name]`) should throw a `KeyError` and execution should not have reached this point...

gem/interpreter.py
```diff
@@ -173,8 +173,13 @@ def _evaluate_operator(e, self):
 def _evaluate_mathfunction(e, self):
     ops = [self(o) for o in e.children]
     result = Result.empty(*ops)
-    names = {"abs": abs,
-             "log": math.log}
+    names = {
+        "abs": abs,
+        "log": math.log,
```
miklos1: We probably need something more for `cmath.log`.

requirements-git.txt
```diff
@@ -1,4 +1,4 @@
-git+https://github.com/coneoproject/COFFEE#egg=COFFEE
-git+https://github.com/firedrakeproject/ufl.git#egg=ufl
+git+https://github.com/coneoproject/COFFEE@complex#egg=COFFEE
+git+https://github.com/firedrakeproject/ufl.git@complex#egg=ufl
```
wence-: This should presumably be dropped now.

tsfc/driver.py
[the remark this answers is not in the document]
wence-: Warning warning.  `parameters` is destructively modified here, which could confuse the caller.

tsfc/driver.py
```diff
@@ -43,7 +43,11 @@ def compile_form(form, prefix="form", parameters=None):
 
     assert isinstance(form, Form)
 
-    fd = ufl_utils.compute_form_data(form)
+    # determine if we're in complex mode; coffee mode does not support complex
+    complx = parameters and "scalar_type" in parameters and 'complex' in parameters["scalar_type"]
```
wence-: Perhaps move the setting up of parameters from `compile_integral` to here, and then use:

```
complx = "complex" in parameters["scalar_type"]
```

Aside, here we're just checking `"complex"`, elsewhere we check `"double complex"`.

tsfc/driver.py
```diff
@@ -266,11 +273,14 @@ def compile_expression_at_points(expression, points, coordinates, parameters=Non
     if extract_arguments(expression):
         return ValueError("Cannot interpolate UFL expression with Arguments!")
 
+    # determine if we're in complex mode
+    complx = parameters["scalar_type"] == 'double complex'
```
wence-: This is different logic to the one implemented in `compile_form`, which just checks for `"complex"`.

tsfc/fem.py
```diff
@@ -76,8 +76,6 @@ def integration_dim(self):
 
     entity_ids = [0]
 
-    precision = PARAMETERS["precision"]
```
wence-: OK, so we don't provide a default precision any more.

tsfc/coffee.py
miklos1: The Boost functions are templated, but their implementation uses comparisons (e.g. `x / v < 0.25`) which does not compile when the templates are instantiated with `std::complex<double>`.

miklos1:
I think @blechta is right, and it would be slightly better if TSFC received the numpy dtype class as scalar type, and then converted it to C typename string.

Currently we get exactly that from PETSc, then with the help of PyOP2 map, we convert that `firedrake.utils` to a C typename string, which we pass to TSFC, and then TSFC converts is back to a numpy dtype class, because that is needed as well.

gem/interpreter.py
miklos1: No longer seems to apply.

tsfc/driver.py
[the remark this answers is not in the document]
miklos1: Fixed.

miklos1:
I pushed a few changes. It is almost there. I added 4922cf0a19a264d0a85c566fb8c29ec3b6927225 for compatibility with firedrake master, so that this could be merged without waiting for the firedrake branch. That commit can, of course, be dropped or reverted when the firedrake branch is ready.

miklos1:
I think this is ready to be merged, irrespective of the corresponding Firedrake changes, although it would be nice if someone could have a final look at this.

Although there are a number of reported, known, and unresolved issues (bugs) with the present complex numbers support, these bugs only affect new functionality; as far as I can tell, there are no
regressions for real numbers.  And since the code is of good quality, I no longer find it justified to sustain the divergence between complex functionality and other development branches.

PR #155, opened 2017-11-10
Register appropriate cites based on optimisation mode

wence-:
I think these are the correct set.

### DIFF
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -370,7 +370,7 @@ def pick_mode(mode):
         cites = {"vanilla": ("Homolya2017", ),
                  "coffee": ("Luporini2016", "Homolya2017", ),
                  "spectral": ("Luporini2016", "Homolya2017", "Homolya2017a"),
-                 "tensor": ("Homolya2017", )}
+                 "tensor": ("Kirby2006", "Homolya2017", )}
         for c in cites[mode]:
             Citations().register(c)
     except ImportError:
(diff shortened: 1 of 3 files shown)


miklos1:
I wonder if this is better than the previous approach, where `Luporini2016` was registered if the COFFEE factorisation algorithm was called (which is called both from `"coffee"` and `"spectral"`), and the TSFC paper was registered always.

miklos1:
Otherwise I think this is correct, and it's good that the FInAT citation is added.

tsfc/driver.py
```diff
@@ -365,6 +365,16 @@ def lower_integral_type(fiat_cell, integral_type):
 
 def pick_mode(mode):
     "Return one of the specialized optimisation modules from a mode string."
+    try:
+        from firedrake_citations import Citations
+        cites = {"vanilla": ("Homolya2017", ),
+                 "coffee": ("Luporini2016", "Homolya2017", ),
+                 "spectral": ("Luporini2016", "Homolya2017", "Homolya2017a"),
+                 "tensor": ("Homolya2017", )}
```
miklos1: I would say [doi:10.1145/1163641.1163644](https://doi.org/10.1145/1163641.1163644) deserves a citation in tensor mode. Though I don't think anyone uses or knows about tensor mode.

wence-: Agreed, added.

wence-:
I've moved the citation registration to only happen in pick_mode, I agree that's better than the previous scattered approach.

PR #156, opened 2017-12-06
value packing

tj-sun:
TSFC packs value instead of pointers.


miklos1:
Would have preferred a rebase instead of the merge, but fine.

PR #158, opened 2018-02-14
Support averaging operators (fixes #107)

### LINKED ISSUE #107: Support for CellAvg and FacetAvg
Moving firedrakeproject/firedrake#1020 here, to make it easier to find TSFC related TODO items.

wence-:
Done by just integrating the averaged expression over the relevant
entity.

Needs a matching UFL merge.

### DIFF
--- a/tsfc/fem.py
+++ b/tsfc/fem.py
@@ -37,7 +37,7 @@ from tsfc.modified_terminals import (analyse_modified_terminal,
                                      construct_modified_terminal)
 from tsfc.parameters import NUMPY_TYPE, PARAMETERS
 from tsfc.ufl_utils import (ModifiedTerminalMixin, PickRestriction,
-                            one_times, simplify_abs,
+                            entity_avg, one_times, simplify_abs,
                             preprocess_expression)
 
 
@@ -174,7 +174,6 @@ class Translator(MultiFunction, ModifiedTerminalMixin, ufl2gem.Mixin):
         # Need context during translation!
         self.context = context
 
-    # XXXAvg is just (\sum_q w_q expr_q) / reference_XXX_volume
     # We just use the provided quadrature rule to
     # perform the integration.
     # Can't put these in the ufl2gem mixin, since they (unlike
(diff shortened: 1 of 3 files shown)


wence-:
That's here: https://bitbucket.org/fenics-project/ufl/pull-requests/89/cellavg-facetavg-are-no-longer-terminal/diff

wence-:
This does the wrong thing for `cell_avg(inner(f, f))*dx` where f is VFS right now. I think I'm just being dense.

wence-:
Think I fixed it, the XXX average is `(sum_q w_q expr_q) / (reference_entity_volume)`, not just `sum_q w_q expr_q`.

blechta:
For affine transform cell, yes. For bendy cells each infinitesimal reference cell volume contributes differently to physical volume...

wence-:
I don't think so.  Here's my reasoning.  For a function `f`, the mean over some compact domain U is `1/Vol(U) \int f dU`.  Which, with quadrature, is `(\sum_q w_q f_q) / (\sum_q w_q)`.

Now consider integrating `cell_avg(f)*dx`.

This is `\int_\Omega \bar{f} dx = \sum_e \int_e \bar{f} dx_e`.

Now I pull back to reference space:

`\sum_e \int_e \bar{f} dx_e = \sum_e \int_E \bar{\tilde{f}} \det J dX`

and `\bar{\tilde{f}} = \int_E \tilde{f} dX / Vol(dX)` (because we're already in reference space).

I think the only thing I assumed here was that averaging commutes will pullback.

Does this make sense?

wence-:
This is still totally borked.  Promotes bugs in tsfc optimisation passes (possibly I'm doing something wrong).  Or doesn't work at all.

blechta:
I don't think that the assumption is correct. We have

```
\bar{f} = \frac{ \int_e f dx }{ \int_e dx } = \frac{ \int_E \hat{f} |\det J| dX }{ \int_E |\det J| dX }
```

This is clearly not equal to

```
\frac{ \int_E \hat{f} dX }{ \int_E dX }
```

unless `\det J` is constant in `E`.

wence-:
Thanks.  I've come to that conclusion too.  The latest commits do this "right", but still have issues elsewise.  So I'm holding off any merging for now.

miklos1:
I like that you could implement this in such a small diff.

PR #159, opened 2018-02-28
generates loopy kernels

tj-sun:
This PR adds the capability to generate loopy kernels instead of coffee kernels.

I duplicated a lot of code in `kernel_interface/firedrake_loopy.py` because I think it will replace `kernel_interface/firedrake.py` eventually, when loopy supports all operations we need in firedrake, so probably not worthwhile to build a common parent class for the two interfaces.

The default behaviour of calling `compile_form()` is still generating a coffee kernel, so it shouldn't affect customers who uses tsfc other than firedrake.

### DIFF
--- a/requirements-git.txt
+++ b/requirements-git.txt
@@ -2,3 +2,4 @@ git+https://github.com/coneoproject/COFFEE#egg=COFFEE
 git+https://github.com/firedrakeproject/ufl.git#egg=ufl
 git+https://github.com/firedrakeproject/fiat.git#egg=fiat
 git+https://github.com/FInAT/FInAT.git#egg=finat
+git+https://github.com/inducer/loopy.git#egg=loopy
--- a/requirements-git.txt
+++ b/requirements-git.txt
@@ -2,4 +2,4 @@ git+https://github.com/coneoproject/COFFEE#egg=COFFEE
 git+https://github.com/firedrakeproject/ufl.git#egg=ufl
 git+https://github.com/firedrakeproject/fiat.git#egg=fiat
 git+https://github.com/FInAT/FInAT.git#egg=finat
---recursive git+https://github.com/firedrakeproject/loopy.git#egg=loopy
+git+https://github.com/firedrakeproject/loopy.git@firedrake#egg=loopy
--- a/requirements-git.txt
+++ b/requirements-git.txt
@@ -2,4 +2,4 @@ git+https://github.com/coneoproject/COFFEE#egg=COFFEE
 git+https://github.com/firedrakeproject/ufl.git#egg=ufl
 git+https://github.com/firedrakeproject/fiat.git#egg=fiat
 git+https://github.com/FInAT/FInAT.git#egg=finat
-git+https://gitlab.tiker.net/inducer/loopy.git#egg=loopy
+--recursive git+https://github.com/firedrakeproject/loopy.git#egg=loopy
--- a/requirements-git.txt
+++ b/requirements-git.txt
@@ -2,4 +2,4 @@ git+https://github.com/coneoproject/COFFEE#egg=COFFEE
 git+https://github.com/firedrakeproject/ufl.git#egg=ufl
 git+https://github.com/firedrakeproject/fiat.git#egg=fiat
 git+https://github.com/FInAT/FInAT.git#egg=finat
-git+https://github.com/inducer/loopy.git#egg=loopy
+git+https://gitlab.tiker.net/inducer/loopy.git#egg=loopy
--- a/tsfc/coffee.py
+++ b/tsfc/coffee.py
@@ -23,18 +23,13 @@ class Bunch(object):
     pass
 
 
-def generate(impero_c, index_names, precision, scalar_type=None, roots=(), argument_indices=()):
+def generate(impero_c, index_names, precision, scalar_type=None):
     """Generates COFFEE code.
 
     :arg impero_c: ImperoC tuple with Impero AST and other data
     :arg index_names: pre-assigned index names
     :arg precision: floating-point precision for printing
     :arg scalar_type: type of scalars as C typename string
-    :arg roots: list of expression DAG roots for attaching
-        #pragma coffee expression
-    :arg argument_indices: argument indices for attaching
-        #pragma coffee linear loop
-        to the argument loops
     :returns: COFFEE function body
     """
     params = Bunch()
@@ -42,8 +37,6 @@ def generate(impero_c, index_names, precision, scalar_type=None, roots=(), argum
     params.indices = impero_c.indices
     params.precision = precision
     params.epsilon = 10.0 * eval("1e-%d" % precision)
-    params.roots = roots
-    params.argument_indices = argument_indices
     params.scalar_type = scalar_type or SCALAR_TYPE
 
     params.names = {}
@@ -88,12 +81,7 @@ def _ref_symbol(expr, parameters):
 
 
 def _root_pragma(expr, parameters):
-    """Decides whether to annonate the expression with
-    #pragma coffee expression"""
-    if expr in parameters.roots:
-        return "#pragma coffee expression"
-    else:
-        return None
+    return None
 
 
 @singledispatch
@@ -119,10 +107,6 @@ def statement_block(tree, parameters):
 
 @statement.register(imp.For)
 def statement_for(tree, parameters):
-    if tree.index in parameters.argument_indices:
-        pragma = "#pragma coffee linear loop"
-    else:
-        pragma = None
     extent = tree.index.extent
     assert extent
     i = _coffee_symbol(parameters.index_names[tree.index])
@@ -130,8 +114,7 @@ def statement_for(tree, parameters):
     return coffee.For(coffee.Decl("int", i, init=0),
                       coffee.Less(i, extent),
                       coffee.Incr(i, 1),
-                      statement(tree.children[0], parameters),
-                      pragma=pragma)
+                      statement(tree.children[0], parameters))
 
 
 @statement.register(imp.Initialise)
--- a/tsfc/coffee.py
+++ b/tsfc/coffee.py
@@ -80,10 +80,6 @@ def _ref_symbol(expr, parameters):
     return _coffee_symbol(parameters.names[expr], rank=tuple(rank))
 
 
-def _root_pragma(expr, parameters):
-    return None
-
-
 @singledispatch
 def statement(tree, parameters):
     """Translates an Impero (sub)tree into a COFFEE AST corresponding
@@ -127,26 +123,20 @@ def statement_initialise(leaf, parameters):
 
 @statement.register(imp.Accumulate)
 def statement_accumulate(leaf, parameters):
-    pragma = _root_pragma(leaf.indexsum, parameters)
     return coffee.Incr(_ref_symbol(leaf.indexsum, parameters),
-                       expression(leaf.indexsum.children[0], parameters),
-                       pragma=pragma)
+                       expression(leaf.indexsum.children[0], parameters))
 
 
 @statement.register(imp.Return)
 def statement_return(leaf, parameters):
-    pragma = _root_pragma(leaf.expression, parameters)
     return coffee.Incr(expression(leaf.variable, parameters),
-                       expression(leaf.expression, parameters),
-                       pragma=pragma)
+                       expression(leaf.expression, parameters))
 
 
 @statement.register(imp.ReturnAccumulate)
 def statement_returnaccumulate(leaf, parameters):
-    pragma = _root_pragma(leaf.indexsum, parameters)
     return coffee.Incr(expression(leaf.variable, parameters),
-                       expression(leaf.indexsum.children[0], parameters),
-                       pragma=pragma)
+                       expression(leaf.indexsum.children[0], parameters))
 
 
 @statement.register(imp.Evaluate)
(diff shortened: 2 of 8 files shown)


tsfc/loopy.py
```diff
@@ -0,0 +1,340 @@
...
+    if not domains:
+        domains = [isl.BasicSet("[] -> {[]}")]
+
+    # Create loopy kernel
+    knl = lp.make_kernel(domains, instructions, data, name=kernel_name, target=lp.CTarget(), seq_dependencies=True)
+
+    # Prevent loopy interchange by loopy
+    knl = lp.prioritize_loops(knl, ",".join(ctx.index_extent.keys()))
+
+    # Help loopy in scheduling by assigning priority to instructions
```
wence-: Why not prioritise before making the kernel?

tsfc/loopy.py
```diff
@@ -0,0 +1,340 @@
...
+        rank = []
+        for index in multiindex:
+            if isinstance(index, gem.Index):
+                rank.append(self.active_indices[index])
+            elif isinstance(index, gem.VariableIndex):
+                rank.append(expression(index.expression, self))
+            else:
+                assert isinstance(index, int)
+                rank.append(index)
+        return tuple(rank)
```
wence-: rank is definitely the wrong noun, why not "indices"?

tsfc/loopy.py
```diff
@@ -0,0 +1,340 @@
...
+
+    def pymbolic_variable(self, node):
+        try:
+            pym = self.gem_to_pymbolic[node]
+        except KeyError:
+            name = self.name_gen(node.name)
+            pym = p.Variable(name)
+            self.gem_to_pymbolic[node] = pym
+        if node in self.indices:
```
wence-: How is a `Variable` ever a `Subscript`?  This is an odd smell.

tj-sun: So if an gem node carries indices (as specified by `ImperoC.indices`), we return a `Subscript`? This is mimicing `tsfc.coffee._ref_symbol()`.

tsfc/loopy.py
tj-sun: Amended.

tsfc/loopy.py
tj-sun: Mainly that I don't know how many instructions I have in total, and it's nicer to have priorities go down from `len(instructions)` down to 0?

tsfc/kernel_interface/firedrake_loopy.py
```diff
@@ -0,0 +1,356 @@
+from __future__ import absolute_import, print_function, division
```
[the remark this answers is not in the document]
tj-sun: Amended.

tsfc/coffee.py
```diff
@@ -83,12 +76,7 @@ def _ref_symbol(expr, parameters):
 
 
 def _root_pragma(expr, parameters):
-    """Decides whether to annonate the expression with
-    #pragma coffee expression"""
-    if expr in parameters.roots:
-        return "#pragma coffee expression"
-    else:
-        return None
+    return None
```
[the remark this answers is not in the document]
tj-sun: Done.

tsfc/kernel_interface/firedrake_loopy.py
```diff
@@ -0,0 +1,354 @@
...
+
+    def get_loopy_arguments(self):
+        args = [self.local_tensor, self.coordinates_arg]
+        if self.kernel.oriented:
+            args.append(self.cell_orientations_loopy_arg)
+        args.extend(self.coefficient_args)
+        if self.kernel.integral_type in ["exterior_facet", "exterior_facet_vert"]:
+            args.append(lp.GlobalArg("facet", dtype=numpy.uint32, shape=(1,)))
+        elif self.kernel.integral_type in ["interior_facet", "interior_facet_vert"]:
+            args.append(lp.GlobalArg("facet", dtype=numpy.uint32, shape=(2,)))
```
wence-: Duplicate code (the same stuff appears in `construct_kernel`).

tsfc/loopy.py
wence-: OK.

tsfc/loopy.py
wence-: OK.

tsfc/loopy.py
```diff
@@ -0,0 +1,338 @@
...
+    """
+    raise AssertionError("cannot generate loopy from %s" % type(tree))
+
+
+@statement.register(imp.Block)
+def statement_block(tree, ctx):
+    statements = []
+    for child in tree.children:
+        statements.extend(statement(child, ctx))
+    return statements
```
wence-: AKA `return list(statement(child, ctx) for child in tree.children)`

wence-: Oh wait, you get a list of statements, so `list(chain(*statement(child, ctx) for child in tree.children))`

tsfc/loopy.py
```diff
@@ -0,0 +1,338 @@
...
+
+
+@_expression.register(gem.Failure)
+def _expression_failure(expr, parameters):
+    raise expr.exception
+
+
+@_expression.register(gem.Product)
+def _expression_product(expr, ctx):
+    return p.Product(tuple([expression(c, ctx) for c in expr.children]))
```
wence-: No need to cons a list, why not: `tuple(expression(c, ctx) for c in expr.children))`

tsfc/loopy.py
```diff
@@ -0,0 +1,338 @@
...
+
+
+@_expression.register(gem.Sum)
+def _expression_sum(expr, ctx):
+    return p.Sum(tuple([expression(c, ctx) for c in expr.children]))
```
wence-: Ditto.

tsfc/loopy.py
```diff
@@ -0,0 +1,338 @@
...
+
+
+@_expression.register(gem.Division)
+def _expression_division(expr, ctx):
+    return p.Quotient(*[expression(c, ctx) for c in expr.children])
```
wence-: Also here: `p.Quotient(*(...))` doesn't cons a list.

tsfc/loopy.py
```diff
@@ -0,0 +1,338 @@
...
+
+
+@_expression.register(gem.Power)
+def _expression_power(expr, ctx):
+    base, exponent = expr.children
+    return p.Power(expression(base, ctx), expression(exponent, ctx))
```
wence-: Why not write this one as `p.Power(*(expression(c, ctx) for c in expr.children))`?

tsfc/driver.py
```diff
@@ -266,7 +268,7 @@ def compile_expression_at_points(expression, points, coordinates, parameters=Non
     expression = ufl_utils.preprocess_expression(expression)
 
     # Initialise kernel builder
-    builder = firedrake_interface.ExpressionKernelBuilder()
+    builder = firedrake_interface_coffee.ExpressionKernelBuilder()
```
wence-: Could we migrate this to loopy too?

tsfc/loopy.py
tj-sun: Probably need to change this after we do dependencies correctly. Loopy doesn't seem to guarantee deterministic code generation though, which will be a problem.

tsfc/kernel_interface/firedrake_loopy.py
tj-sun: I've forgotten why I need this function in the first place, deleted.

tsfc/loopy.py
tj-sun: should it be
`list(chain.from_iterable(statement(child, ctx) for child in tree.children))` ?

tsfc/loopy.py
tj-sun: Oh yes!

tsfc/loopy.py
wence-: Either that or `list(chain(*(statement(...) ...)))` ?

tsfc/loopy.py
tj-sun: Sure.

tsfc/driver.py
tj-sun: Let me try.

tsfc/loopy.py
tj-sun: Yep that's better.

requirements-git.txt
```diff
@@ -2,3 +2,4 @@ git+https://github.com/coneoproject/COFFEE#egg=COFFEE
 git+https://github.com/firedrakeproject/ufl.git#egg=ufl
 git+https://github.com/firedrakeproject/fiat.git#egg=fiat
 git+https://github.com/FInAT/FInAT.git#egg=finat
+--recursive git+https://github.com/firedrakeproject/loopy.git#egg=loopy
```
miklos1: So this recursive here is just wrong... That option should be used when you want to pull another requirements file.

tsfc/loopy.py
```diff
@@ -0,0 +1,340 @@
...
+            return p.Variable("j0")(expression(arg, ctx))
+        elif nu == gem.one:
+            return p.Variable("j1")(expression(arg, ctx))
+    if name == 'yn':
+        nu, arg = expr.children
+        if nu == gem.Zero():
+            return p.Variable("y0")(expression(arg, ctx))
+        elif nu == gem.one:
+            return p.Variable("y1")(expression(arg, ctx))
+    return p.Variable(name)(*[expression(c, ctx) for c in expr.children])
```
miklos1: This function needs updating... see current `tsfc/coffee.py`.

tj-sun: Done -  but need testing on the complex branch.

tsfc/driver.py
```diff
@@ -84,7 +89,7 @@ def compile_integral(integral_data, form_data, prefix, parameters, interface):
         _.update(parameters)
         parameters = _
     if interface is None:
-        interface = firedrake_interface.KernelBuilder
+        interface = firedrake_interface_loopy
```
miklos1: Needs update, I guess...

miklos1:
OK, I find this mostly fine. The one thing I find most ad-hoc and ugly, are the random conditionals here and there about whether we are generating coffee or loopy... So the question is whether the presence of two code generation systems meant to be a short-term temporary thing, or a longer term thing.

- In the former case, I think it is not worth investing time to improve code quality in this respect, since the COFFEE code generation will go away soon anyway, and with that the problem of ad-hoc conditionals also goes away.
- In the latter case, however, a refactoring around which code generation pipeline is being used is strongly suggested.

tsfc/driver.py
tj-sun: corrected.

tj-sun:
> OK, I find this mostly fine. The one thing I find most ad-hoc and ugly, are the random conditionals here and there about whether we are generating coffee or loopy... So the question is whether the presence of two code generation systems meant to be a short-term temporary thing, or a longer term thing.

It is intended to be short term - functionalities such as Slate still need to go through COFFEE codegen before we extend loopy to incorporate more linear algebra intrinsics, which is not too hard to do but requires some careful planning.

Another thing is the `coffee=False` flag predates the `interface` parameter, so maybe now we can remove the `coffee` flag.

tsfc/driver.py
```diff
@@ -56,6 +57,10 @@ def compile_form(form, prefix="form", parameters=None, interface=None):
     logger.info(GREEN % "compute_form_data finished in %g seconds.", time.time() - cpu_time)
 
     kernels = []
+    if coffee:
```
wence-: This now ignores the incoming interface and should say:
```python
if interface is None:
   if coffee:
       ...
   else:
```

wence-: Although that then means that we handle `interface` is None in two places. So perhaps move this logic to `compile_integral`

tsfc/driver.py
```diff
@@ -84,7 +89,7 @@ def compile_integral(integral_data, form_data, prefix, parameters, interface):
         _.update(parameters)
         parameters = _
     if interface is None:
-        interface = firedrake_interface.KernelBuilder
+        interface = firedrake_interface_loopy.KernelBuilder
```
wence-: i.e. pass `coffee=True/False` into compile_integral and handle the relevant cases here.

tsfc/driver.py
```diff
@@ -291,7 +295,10 @@ def compile_expression_at_points(expression, points, coordinates, interface=None
 
     # Initialise kernel builder
     if interface is None:
-        interface = firedrake_interface.ExpressionKernelBuilder
+        if coffee:
+            interface = firedrake_interface_coffee.ExpressionKernelBuilder
+        else:
+            interface = firedrake_interface_loopy.ExpressionKernelBuilder
```
wence-: OK, this one is right.

tsfc/driver.py
dham: I don't really understand this code. Maybe you had better do it.

tsfc/kernel_interface/firedrake.py
```diff
@@ -171,20 +172,25 @@ def require_cell_orientations(self):
...
-    def construct_kernel(self, return_arg, body):
+    def construct_kernel(self, return_arg, impero_c, precision, index_names):
         """Constructs an :class:`ExpressionKernel`.
 
         :arg return_arg: COFFEE argument for the return value
         :arg body: function body (:class:`coffee.Block` node)
         :returns: :class:`ExpressionKernel` object
         """
+
+        from tsfc.coffee import generate as generate_coffee
```
wence-: Already imported at top level.

tsfc/driver.py
dham: this is the one I did by hand!

wence-:
Let's review #189 instead, the diff is identical, but without the WIP commits everywhere.

PR #161, opened 2018-05-03
Support NodalEnrichedElement

wence-:
@blechta, you had an old branch that did this, does this DTRT for you?


blechta:
Otherwise looks good.

wence-:
No, because although the name maps directly, we need to convert all the child elements too.

PR #162, opened 2018-05-03
FIAT and FInAT interface: add bindings for elements on hexahedra

wence-:
Resurrect some finished, but non-merged branches.

### DIFF
--- a/tsfc/fiatinterface.py
+++ b/tsfc/fiatinterface.py
@@ -56,6 +56,8 @@ supported_elements = {
     "Q": None,
     "RTCE": None,
     "RTCF": None,
+    "NCE": None,
+    "NCF": None,
 }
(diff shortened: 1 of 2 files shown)


miklos1:
Probably the reason I did not merge this, is that facet integrals were broken on unstructured hexahedra (only affects FEniCS since Firedrake only has semi-structured hexahedra). I described the issue in more detail in the future work section of my thesis.

wence-:
Good point.  Probably I should add a `NotImplementedError` in `compile_form` if doing facet integrals + hexahedron in "spectral" mode.

miklos1:
The problem occurs in other modes as well, because they originate from the tensor product element now being structured... and the conversion of elements UFL -> FInAT does not depend on mode.

wence-:
Ah, OK, thanks.

PR #165, opened 2018-05-18
Update in UFC kernel interface for the loopy work

blechta:


tsfc/kernel_interface/ufc.py
```diff
@@ -96,7 +96,23 @@ def set_coefficients(self, integral_data, form_data):
...
+
+        :arg name: function name
+        :arg impero_c: ImperoC tuple with Impero AST and other data
+        :arg precision: floating-point precision for printing
+        :arg index_names: pre-assigned index names
+        :returns: a COFFEE function definition object
+        """
+        from tsfc.coffee import generate as generate_coffee
+        body = generate_coffee(impero_c, index_names, precision)
+        return self._construct_kernel_from_body(name, body)
```
wence-: OK, so you just update to the new coffee interface.  I think going forward we would like to drop the coffee output backends and just make loopy.  But I don't know what the ffc-related issues are likely to be.

PR #167, opened 2018-06-27
Add support for new FInAT PhysicalGeometry callbacks

wence-:
This enables FInAT to correctly tabulate Bell, Argyris, Hermite, and Morley elements.

Only affine mappings on triangles for now.  Uses theory of Kirby 2018.

### DIFF
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -7,7 +7,7 @@ from itertools import chain
 
 from numpy import asarray
 
-import ufl
+# import ufl
 from ufl.algorithms import extract_arguments, extract_coefficients
 from ufl.algorithms.analysis import has_type
 from ufl.classes import Form, GeometricQuantity
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -141,7 +141,8 @@ def compile_integral(integral_data, form_data, prefix, parameters,
         mode = pick_mode(params["mode"])
         mode_irs.setdefault(mode, collections.OrderedDict())
 
-        integrand = ufl.replace(integral.integrand(), form_data.function_replace_map)
+        # integrand = ufl.replace(integral.integrand(), form_data.function_replace_map)
+        integrand = integral.integrand()
         integrand = ufl_utils.split_coefficients(integrand, builder.coefficient_split)
 
         # Check if the integral has a quad degree attached, otherwise use
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -105,6 +105,7 @@ def compile_integral(integral_data, form_data, prefix, parameters,
     return_variables = builder.set_arguments(arguments, argument_multiindices)
 
     builder.set_coordinates(mesh)
+    builder.set_cell_sizes(mesh)
 
     builder.set_coefficients(integral_data, form_data)
 
@@ -202,6 +203,9 @@ def compile_integral(integral_data, form_data, prefix, parameters,
     # Look for cell orientations in the IR
     if builder.needs_cell_orientations(expressions):
         builder.require_cell_orientations()
+
+    if builder.needs_cell_sizes(expressions):
+        builder.require_cell_sizes()
 
     # Construct ImperoC
     split_argument_indices = tuple(chain(*[var.index_ordering()
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -7,7 +7,7 @@ from itertools import chain
 
 from numpy import asarray
 
-# import ufl
+import ufl
 from ufl.algorithms import extract_arguments, extract_coefficients
 from ufl.algorithms.analysis import has_type
 from ufl.classes import Form, GeometricQuantity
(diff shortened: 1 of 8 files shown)


tsfc/driver.py
```diff
@@ -137,7 +138,8 @@ def compile_integral(integral_data, form_data, prefix, parameters,
         mode = pick_mode(params["mode"])
         mode_irs.setdefault(mode, collections.OrderedDict())
 
-        integrand = ufl.replace(integral.integrand(), form_data.function_replace_map)
+        # integrand = ufl.replace(integral.integrand(), form_data.function_replace_map)
+        integrand = integral.integrand()
```
miklos1: This was originally added because of some DOLFIN `Expression` issue, and this draft change undoes that. So it likely breaks some cases of using DOLFIN with TSFC, but I did not have time to debug how to make every case work.

tsfc/fem.py
```diff
@@ -108,6 +112,88 @@ def translator(self):
...
+    Required for elements whose basis transformation requires physical
+    geometry such as Argyris and Hermite.
+
+    :arg mt: The modified terminal whose element will be tabulated.
+    :arg interface: The interface carrying information (generally a
+        :class:`PointSetContext`).
+    """
+    def __init__(self, mt, interface):
+        super().__init__()
+        self.mt = mt
```
miklos1: Chuck in the whole modified terminal, so we access to restrictions (for interior facet integrals) etc.

tsfc/fem.py
```diff
@@ -108,6 +112,88 @@ def translator(self):
...
+            expr = NegativeRestricted(expr)
+        expr = preprocess_expression(expr)
+
+        config = {"point_set": PointSingleton(point)}
+        config.update(self.config)
+        context = PointSetContext(**config)
+        return map_expr_dag(context.translator, expr)
+
+    def reference_normals(self):
+        return gem.Literal(numpy.asarray([self.interface.fiat_cell.compute_normal(i) for i in range(3)]))
```
miklos1: Yes, FIAT `compute_normal` gives normals pointing either in or out, consistent with the Rognes et al. (2009) numbering.

tsfc/fem.py
```diff
@@ -482,6 +569,7 @@ def callback(entity_id):
 
 @translate.register(Coefficient)
 def translate_coefficient(terminal, mt, ctx):
+    # import ipdb; ipdb.set_trace()
```
miklos1: debug code

tsfc/kernel_interface/common.py
```diff
@@ -58,6 +58,11 @@ def cell_orientation(self, restriction):
                                                gem.Literal(1),
                                                gem.Literal(numpy.nan)))
 
+    def cell_size(self, restriction):
+        f = {None: (), '+': (0, ), '-': (1, )}[restriction]
+        # cell_sizes expression must have been set up by now.
+        return gem.partial_indexed(self._cell_sizes, f)
+
```
miklos1: `cell_size` is like `cell_orientations` that they are calculated once globally?

tsfc/kernel_interface/firedrake.py
```diff
@@ -87,6 +90,15 @@ def needs_cell_orientations(ir):
                 return True
         return False
 
+    @staticmethod
+    def needs_cell_sizes(ir):
+        """Does a multi-root GEM expression DAG references cell
+        orientations?"""
```
miklos1: Docstring copy-paste

tsfc/kernel_interface/firedrake.py
```diff
@@ -199,7 +220,7 @@ def set_coefficients(self, integral_data, form_data):
         # of reduced_coefficients the integral requires.
         for i in range(len(integral_data.enabled_coefficients)):
             if integral_data.enabled_coefficients[i]:
-                coefficient = form_data.function_replace_map[form_data.reduced_coefficients[i]]
+                coefficient = form_data.reduced_coefficients[i]
```
miklos1: This is related to the same DOLFIN issue mentioned above.

tsfc/kernel_interface/firedrake.py
```diff
@@ -221,6 +242,16 @@ def require_cell_orientations(self):
         """Set that the kernel requires cell orientations."""
         self.kernel.oriented = True
 
+    def require_cell_sizes(self):
+        """Set that the kernel requires cell sizes."""
+        self.kernel.needs_cell_sizes = True
+
+    def set_cell_sizes(self, domain):
+        f = Coefficient(FunctionSpace(domain, FiniteElement("P", domain.ufl_cell(), 1)))
```
miklos1: Continuous first-order?

tsfc/kernel_interface/common.py
wence-: Indeed.

tsfc/driver.py
wence-: I think I fixed the failing that this caused for some other context via UFL changes a while ago.  So I reverted these bits and am running the Firedrake test suite to see if everything still works.

wence-: Yeah, that seemed to work.

miklos1: That's good!

PR #169, opened 2018-07-04
gem: Add some minimal sugar

### LINKED ISSUE #168: Add overloaded operators to GEM
So that writing complicated GEM expressions might be more concise and easier to read and review. Relevant especially with changes like FInAT/FInAT#48 and #167.

To avoid excess, we may want to restrict ourselves initially to the basic operators `+`, `-`, `*`, `/`, and indexing.

wence-:
Support +, -, *, /, and __getitem__ magic methods.  Fixes #168.

### DIFF
--- a/tests/test_syntax_sugar.py
+++ b/tests/test_syntax_sugar.py
@@ -1,0 +1,26 @@
+import pytest
+from gem import gem
+
+
+def test_expressions():
+    x = gem.Variable("x", (3, 4))
+    y = gem.Variable("y", (4, ))
+    i = gem.Index()
+    j = gem.Index()
+
+    xij = x[i, j]
+    yj = y[j]
+
+    assert xij == gem.Indexed(x, (i, j))
+    assert yj == gem.Indexed(y, (j, ))
+
+    assert xij + yj == gem.Sum(xij, yj)
+    assert xij * yj == gem.Product(xij, yj)
+    assert xij - yj == gem.Sum(xij, gem.Product(gem.Literal(-1), yj))
+    assert xij / yj == gem.Division(xij, yj)
+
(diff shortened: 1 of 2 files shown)


tests/test_syntax_sugar.py
```diff
@@ -0,0 +1,26 @@
...
+    assert xij == gem.Indexed(x, (i, j))
+    assert yj == gem.Indexed(y, (j, ))
+
+    assert xij + yj == gem.Sum(xij, yj)
+    assert xij * yj == gem.Product(xij, yj)
+    assert xij - yj == gem.Sum(xij, gem.Product(gem.Literal(-1), yj))
+    assert xij / yj == gem.Division(xij, yj)
+
+    with pytest.raises(AssertionError):
+        xij + 1
```
miklos1: I think it may make sense to enable this instead of having to write `xij + Literal(1)`. For that, operators could just pipe their arguments through a function `as_gem` (like as `as_ufl`), which returns GEM expressions untouched, wraps scalar numbers (`numbers.Number`?) in a Literal, and fails for everything else.

And then one may also need the reverse operators (radd, rsub, rmul, etc.) so that `1 + xij` could work as well.

tests/test_syntax_sugar.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+from gem import gem
+
+
+def test_expressions():
+    x = gem.Variable("x", (3, 4))
+    y = gem.Variable("y", (4, ))
+    i = gem.Index()
+    j = gem.Index()
```
miklos1: One more UFL imitation that might make sense to adopt is `indices`, so the above two lines could be replaced with:

    i, j = gem.indices(2)

PR #175, opened 2018-10-24
Register Bernstein element

miklos1:
Requires the UFL and FIAT changes first, of course.


tsfc/finatinterface.py
```diff
@@ -104,6 +104,8 @@ def convert_finiteelement(element, **kwargs):
             raise ValueError("Quadrature scheme and degree must be specified!")
 
         return finat.QuadratureElement(cell, degree, scheme), set()
+    elif element.family() == "Bernstein":
+        return fiat_compat(element), set()
```
miklos1: The combination of this function and the mapping from element names to FInAT classes is getting a bit messy, and might do well with a bit of refactoring at some point.

PR #178, opened 2018-11-09
Perhaps fix translation of facet_avg

wence-:
Could be correct?


wence-:
This is basically copying the translation of FacetArea.

PR #179, opened 2018-12-03
Use correct measure of facet volume in facet_avg

wence-:


wence-:
I can provide a better commit message :). Tested in firedrakeproject/firedrake#1344

PR #184, opened 2019-02-27
Wence/builder interface

wence-:
This changes the interface to compile_form to pass an interface class, rather than a module. Simplifies code in PCPatch in Firedrake.

### DIFF
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -43,8 +43,6 @@ def compile_form(form, prefix="form", parameters=None, interface=None):
 
     assert isinstance(form, Form)
 
-    if interface is None:
-        interface = firedrake_interface
     # Determine whether in complex mode:
     # complex nodes would break the refactoriser.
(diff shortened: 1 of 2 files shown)


tsfc/driver.py
```diff
@@ -57,6 +55,9 @@ def compile_form(form, prefix="form", parameters=None, interface=None):
     fd = ufl_utils.compute_form_data(form, complex_mode=complex_mode)
     logger.info(GREEN % "compute_form_data finished in %g seconds.", time.time() - cpu_time)
 
+    if interface is None:
+        interface = firedrake_interface.KernelBuilder
```
miklos1: I think it would be slightly nicer to this in `compile_integral` instead (setting the default value of the optional argument to `None`), and let `compile_form` pass through `None` or whatever was specified by the caller.  That way, we only need to tell once (instead of twice) what the default kernel interface is.

wence-: Good point.

PR #185, opened 2019-02-27
Tighten classifier for delta elimination

### LINKED ISSUE #182: cannot generate COFFEE from <class 'gem.gem.Delta'>
[demo-beam3d_ufl.zip](https://github.com/firedrakeproject/tsfc/files/2895722/demo-beam3d_ufl.zip)
The attached file, when compiled with `ffc -l dolfin -r tsfc demo-beam3d_ufl.ufl` fails because of `AssertionError: cannot generate COFFEE from <class 'gem.gem.Delta'>`, while uflacs digests it without errors.

wence-:
If we have a Delta node inside a ListTensor, we can't consider an
Indexed node to be ATOMIC. Fixes #182.

Basically @miklos1's patch, but with a Memoizer. Seems to work for @mmorandi's example.

### DIFF
--- a/tests/test_tsfc_182.py
+++ b/tests/test_tsfc_182.py
@@ -1,0 +1,33 @@
+import pytest
+
+from ufl import (Coefficient, FiniteElement, MixedElement,
+                 TestFunction, VectorElement, dx, inner, tetrahedron)
+
+from tsfc import compile_form
+
+
+@pytest.mark.parametrize('mode', ['vanilla', 'coffee', 'spectral'])
+def test_delta_elimination(mode):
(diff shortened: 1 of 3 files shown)


miklos1:
I suggest just taking the minimal example from #182 as a test case. No need for specific assertions, just to make sure that the code generation does not fail. Preferably parametrizing the test case over optimisation modes. `vanilla`, `coffee`, and `spectral`.  If `tensor` happens to work as well, then that should be added, too, otherwise, don't bother.

miklos1:
Added test case.

PR #186, opened 2019-03-01
Safe refactorisation of Conditional nodes

### LINKED ISSUE #183: derivative of conditional
[beam.zip](https://github.com/firedrakeproject/tsfc/files/2901410/beam.zip)
Consider the attached ufl, a test case distilled from a more complex code. All the source of `NaN`s are protected by conditionals.
The generated code (beam.h, attached, obtained with ffc -l dolfin -r tsfc) has these lines:

`double  t12  = ((((((t9) * (t9))) + (((t10) * (t10))))) + (((t11) * (t11))));`
`double  t14  = ((1) / (t12));`
`double  t15  = ((sin(t12)) * (t14));`
`double  t23  = (t13 ? 1 : 0);`
`double  t25  = (((t13 ? 0 : 1)) * (t14));`
`double  t27  = ((((t23) * (t24))) + (((t25) * (((((-1) * (((t15) * (t24))))) + (((t26) * (t24))))))));`

With null `t9`, `t10` and `t11` (i.e. `u_phi`) `t15` will be equal to `NaN`. In `t27` tsfc tried to reproduce the conditional by multiplying `t24` by `t23`, and summing the result to the product between a second expression, function of `t15`, and `t25`. Clearly the reasoning is that if `t23==0` then `t25!=0` and vice versa. However, when `t23==1`, `t25==0` and `t15==NaN` then `t27==NaN` because `0*NaN=NaN`. Thus,  the form is computed incorrectly. It should be

`double  t27  = (t13 ? (t24) : (((t14) * (((((-1) * (((t15) * (t24))))) + (((t26) * (t24))))))));`

I think this was introduced in this [commit](https://github.com/firedrakeproject/tsfc/commit/92cc254e3872ff72e901b2b5ed8768144b11f547), in an attempt to fix iusse #119  .

miklos1:
Resolves #183.  Needs automatic testing, but I think that is better done through Firedrake, where the generated code can be executed so that verify the lack of `NaN`s in the result.

### DIFF
--- a/gem/gem.py
+++ b/gem/gem.py
@@ -408,7 +408,7 @@ class Conditional(Node):
 
     def __new__(cls, condition, then, else_):
         assert not condition.shape
-        assert then.shape == else_.shape
+        assert then.shape == else_.shape == ()
 
         # If both branches are the same, just return one of them.  In
         # particular, this will help constant-fold zeros.
--- a/gem/optimise.py
+++ b/gem/optimise.py
@@ -625,37 +625,3 @@ def aggressive_unroll(expression):
     expression, = unroll_indexsum((expression,), predicate=lambda index: True)
     expression, = remove_componenttensors((expression,))
     return expression
-
-
-@singledispatch
-def _expand_conditional(node, self):
-    raise AssertionError("cannot handle type %s" % type(node))
-
-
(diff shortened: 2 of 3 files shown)


miklos1:
Seems not to break Firedrake. Just need to prepare test cases now...

PR #187, opened 2019-03-11
Runtime tabulated v2

celdred:
This adds a Themis interface for TSFC that supports both assembly and interpolation kernels with runtime tabulation. The relevant changes are stubbed out in the Firedrake and UFC interfaces.

### DIFF
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -210,6 +210,9 @@ def compile_integral(integral_data, form_data, prefix, parameters, interface):
     expressions = impero_utils.preprocess_gem(expressions, **options)
     assignments = list(zip(return_variables, expressions))
 
+    # Register tabulations for runtime tabulated elements (used by Themis)
+    builder.register_tabulations(expressions)
+
     # Look for cell orientations in the IR
     if builder.needs_cell_orientations(expressions):
         builder.require_cell_orientations()
@@ -252,10 +255,10 @@ def compile_integral(integral_data, form_data, prefix, parameters, interface):
     body = generate_coffee(impero_c, index_names, parameters["precision"],
                            parameters["scalar_type"], expressions, split_argument_indices)
 
-    return builder.construct_kernel(kernel_name, body)
+    return builder.construct_kernel(kernel_name, body, quad_rule)
 
 
-def compile_expression_at_points(expression, points, coordinates, parameters=None):
+def compile_expression_at_points(expression, points, coordinates, interface=None, parameters=None):
     """Compiles a UFL expression to be evaluated at compile-time known
     reference points.  Useful for interpolating UFL expressions onto
     function spaces with only point evaluation nodes.
@@ -264,6 +267,7 @@ def compile_expression_at_points(expression, points, coordinates, parameters=Non
     :arg points: reference coordinates of the evaluation points
     :arg coordinates: the coordinate function
     :arg parameters: parameters object
+    :arg interface: backend module for the kernel interface
     """
     import coffee.base as ast
 
@@ -286,7 +290,10 @@ def compile_expression_at_points(expression, points, coordinates, parameters=Non
                                                  complex_mode=complex_mode)
 
     # Initialise kernel builder
-    builder = firedrake_interface.ExpressionKernelBuilder(parameters["scalar_type"])
+    if interface is None:
+        interface = firedrake_interface.ExpressionKernelBuilder
+
+    builder = interface(parameters["scalar_type"])
 
     # Replace coordinates (if any)
     domain = expression.ufl_domain()
@@ -317,6 +324,9 @@ def compile_expression_at_points(expression, points, coordinates, parameters=Non
     tensor_indices = tuple(gem.Index() for s in value_shape)
     if value_shape:
         ir = gem.Indexed(ir, tensor_indices)
+
+    # Register tabulations for runtime tabulated elements (used by Themis)
+    builder.register_tabulations([ir])
 
     # Build kernel body
     return_shape = (len(points),) + value_shape
--- a/tsfc/kernel_interface/firedrake.py
+++ b/tsfc/kernel_interface/firedrake.py
@@ -16,7 +16,7 @@ from tsfc.kernel_interface.common import KernelBuilderBase as _KernelBuilderBase
 
 
 # Expression kernel description type
-ExpressionKernel = namedtuple('ExpressionKernel', ['ast', 'oriented', 'needs_cell_sizes', 'coefficients'])
+ExpressionKernel = namedtuple('ExpressionKernel', ['ast', 'oriented', 'needs_cell_sizes', 'coefficients', 'tabulations'])
 
 
 def make_builder(*args, **kwargs):
@@ -25,7 +25,7 @@ def make_builder(*args, **kwargs):
 
 class Kernel(object):
     __slots__ = ("ast", "integral_type", "oriented", "subdomain_id",
-                 "domain_number", "needs_cell_sizes",
+                 "domain_number", "needs_cell_sizes", "tabulations", "quadrature_rule",
                  "coefficient_numbers", "__weakref__")
     """A compiled Kernel object.
 
@@ -38,10 +38,12 @@ class Kernel(object):
         original_form.ufl_domains() to get the correct domain).
     :kwarg coefficient_numbers: A list of which coefficients from the
         form the kernel needs.
+    :kwarg quadrature_rule: The finat quadrature rule used to generate this kernel
+    :kwarg tabulations: The runtime tabulations this kernel requires
     :kwarg needs_cell_sizes: Does the kernel require cell sizes.
     """
     def __init__(self, ast=None, integral_type=None, oriented=False,
-                 subdomain_id=None, domain_number=None,
+                 subdomain_id=None, domain_number=None, quadrature_rule=None,
                  coefficient_numbers=(),
                  needs_cell_sizes=False):
         # Defaults
@@ -155,6 +157,13 @@ class ExpressionKernelBuilder(KernelBuilderBase):
                 self.coefficients.append(coefficient)
                 self.kernel_args.append(self._coefficient(coefficient, "w_%d" % (i,)))
 
+    def register_tabulations(self, expressions):
+        tabulations = {}
+        for node in traversal(expressions):
+            if isinstance(node, gem.Variable) and node.name.startswith("rt_"):
+                tabulations[node.name] = node.shape
+        self.tabulations = tuple(sorted(tabulations.items()))
+
     def require_cell_orientations(self):
         """Set that the kernel requires cell orientations."""
         self.oriented = True
@@ -176,8 +185,12 @@ class ExpressionKernelBuilder(KernelBuilderBase):
             args.append(self.cell_sizes_arg)
         args.extend(self.kernel_args)
 
+        for name_, shape in self.tabulations:
+            args.append(coffee.Decl(self.scalar_type, coffee.Symbol(
+                name_, rank=shape), qualifiers=["const"]))
+
         kernel_code = super(ExpressionKernelBuilder, self).construct_kernel("expression_kernel", args, body)
-        return ExpressionKernel(kernel_code, self.oriented, self.cell_sizes, self.coefficients)
+        return ExpressionKernel(kernel_code, self.oriented, self.cell_sizes, self.coefficients, self.tabulations)
 
 
 class KernelBuilder(KernelBuilderBase):
@@ -265,6 +278,13 @@ class KernelBuilder(KernelBuilderBase):
                 self._coefficient(coefficient, "w_%d" % i))
         self.kernel.coefficient_numbers = tuple(coefficient_numbers)
 
+    def register_tabulations(self, expressions):
+        tabulations = {}
+        for node in traversal(expressions):
+            if isinstance(node, gem.Variable) and node.name.startswith("rt_"):
+                tabulations[node.name] = node.shape
+        self.kernel.tabulations = tuple(sorted(tabulations.items()))
+
     def require_cell_orientations(self):
         """Set that the kernel requires cell orientations."""
         self.kernel.oriented = True
@@ -273,7 +293,7 @@ class KernelBuilder(KernelBuilderBase):
         """Set that the kernel requires cell sizes."""
         self.kernel.needs_cell_sizes = True
 
-    def construct_kernel(self, name, body):
+    def construct_kernel(self, name, body, quadrature_rule):
         """Construct a fully built :class:`Kernel`.
 
         This function contains the logic for building the argument
@@ -281,6 +301,7 @@ class KernelBuilder(KernelBuilderBase):
 
         :arg name: function name
         :arg body: function body (:class:`coffee.Block` node)
+        :arg quadrature rule: quadrature rule (not used, stubbed out for Themis integration)
         :returns: :class:`Kernel` object
         """
         args = [self.local_tensor, self.coordinates_arg]
@@ -297,6 +318,12 @@ class KernelBuilder(KernelBuilderBase):
             args.append(coffee.Decl("unsigned int",
                                     coffee.Symbol("facet", rank=(2,)),
                                     qualifiers=["const"]))
+
+        for name_, shape in self.kernel.tabulations:
+            args.append(coffee.Decl(self.scalar_type, coffee.Symbol(
+                name_, rank=shape), qualifiers=["const"]))
+
+        self.kernel.quadrature_rule = quadrature_rule
 
         self.kernel.ast = KernelBuilderBase.construct_kernel(self, name, args, body)
         return self.kernel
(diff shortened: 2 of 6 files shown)


tsfc/kernel_interface/themis.py
```diff
@@ -0,0 +1,411 @@
...
+                    coefficients.append(coefficient)
+                # This is which coefficient in the original form the
+                # current coefficient is.
+                # Consider f*v*dx + g*v*ds, the full form contains two
+                # coefficients, but each integral only requires one.
+                coefficient_numbers.append(form_data.original_coefficient_positions[i])
+        for i, coefficient in enumerate(coefficients):
+            self.coefficient_args.append(
+                self._coefficient(coefficient, "w_%d" % i))
+        self.kernel.coefficient_numbers = tuple(coefficient_numbers)
```
wence-: Again, lots of this code is identical to the firedrake kernel builder.

tsfc/kernel_interface/themis.py
```diff
@@ -0,0 +1,411 @@
...
+                                    coffee.Symbol("facet", rank=(1,)),
+                                    qualifiers=["const"]))
+        elif self.kernel.integral_type in ["interior_facet", "interior_facet_vert"]:
+            args.append(coffee.Decl("unsigned int",
+                                    coffee.Symbol("facet", rank=(2,)),
+                                    qualifiers=["const"]))
+
+        for name_, shape in self.tabulations:
+            args.append(coffee.Decl(self.scalar_type, coffee.Symbol(
+                name_, rank=shape), qualifiers=["const"]))
```
wence-: This bit is different.

tsfc/kernel_interface/themis.py
```diff
@@ -0,0 +1,411 @@
...
+         expression - GEM expression referring to the Coefficient
+                      values
+    """
+    assert isinstance(interior_facet, bool)
+
+    if coefficient.ufl_element().family() == 'Real':
+        # Constant
+
+        # Here we use arrays rather than pointers to fit with how Themis calls kernels
+        funarg = coffee.Decl(scalar_type, coffee.Symbol(name, rank=(1,)), qualifiers=["const"])
```
wence-: Can you show the difference in the code you want from what you get? Because I'm not sure we will have this control of the argument list with loopy.

This makes:
```c
const double foo[1];
```
Rather than the current 
```c
const * restrict foo
```
What is wrong with the latter for you?

tsfc/kernel_interface/themis.py
```diff
@@ -0,0 +1,411 @@
...
+
+    finat_element = create_element(coefficient.ufl_element())
+    shape = finat_element.index_shape
+    size = numpy.prod(shape, dtype=int)
+
+    # Here we use arrays rather than pointers to fit with how Themis calls kernels
+    if not interior_facet:
+        funarg = coffee.Decl(scalar_type, coffee.Symbol(name, rank=(size,)), qualifiers=["const"])
+    else:
+        funarg = coffee.Decl(scalar_type, coffee.Symbol(name, rank=(2*size,)), qualifiers=["const"])
```
wence-: Similarly here.

tsfc/kernel_interface/themis.py
```diff
@@ -0,0 +1,411 @@
...
+                   for r, s in zip(restrictions, u_shape)]
+                  for restrictions in product((0, 1), repeat=len(arguments))]
+    else:
+        c_shape = tuple(u_shape)
+        slicez = [[slice(s) for s in u_shape]]
+
+    funarg = coffee.Decl(scalar_type, coffee.Symbol("A", rank=c_shape))
+    varexp = gem.Variable("A", c_shape)
+    expressions = [expression(gem.view(varexp, *slices)) for slices in slicez]
+    return funarg, prune(expressions)
```
wence-: This is identical to the firedrake.py version.

tsfc/driver.py
```diff
@@ -286,7 +290,10 @@ def compile_expression_at_points(expression, points, coordinates, parameters=Non
                                                  complex_mode=complex_mode)
 
     # Initialise kernel builder
-    builder = firedrake_interface.ExpressionKernelBuilder(parameters["scalar_type"])
+    if interface is None:
+        builder = firedrake_interface.ExpressionKernelBuilder(parameters["scalar_type"])
+    else:
+        builder = interface.ExpressionKernelBuilder(parameters["scalar_type"])
```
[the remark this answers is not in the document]
celdred: Fixed this in the new commit.

tsfc/kernel_interface/firedrake.py
```diff
@@ -273,14 +273,15 @@ def require_cell_sizes(self):
...
-    def construct_kernel(self, name, body):
+    def construct_kernel(self, name, body, quadrature_rule):
         """Construct a fully built :class:`Kernel`.
 
         This function contains the logic for building the argument
         list for assembly kernels.
 
         :arg name: function name
         :arg body: function body (:class:`coffee.Block` node)
+        :arg quadrature rule: quadrature rule (not used, stubbed out for Themis integration)
```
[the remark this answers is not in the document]
celdred: Fixed this in the new commit.

tsfc/kernel_interface/ufc.py
```diff
@@ -102,14 +102,15 @@ def set_coefficients(self, integral_data, form_data):
...
-    def construct_kernel(self, name, body):
+    def construct_kernel(self, name, body, quadrature_rule):
         """Construct a fully built kernel function.
 
         This function contains the logic for building the argument
         list for assembly kernels.
 
         :arg name: function name
         :arg body: function body (:class:`coffee.Block` node)
+        :arg quadrature rule: quadrature rule (not used, stubbed out for Themis integration)
```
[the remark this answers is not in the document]
celdred: Fixed this in the new commit.

tsfc/kernel_interface/themis.py
celdred: There is no longer a Themis interface, instead the Firedrake interface (with tabulations added) works for us.

tsfc/kernel_interface/themis.py
celdred: I have refactored Themis to use the same kernel arguments as Firedrake (plus tabulations), so the Themis interface is no longer needed.

celdred:
I have removed the Themis interface and instead added tabulation registration to the Firedrake interface (which does nothing unless there are runtime tabulated elements in the form, so it shouldn't break anything on the Firedrake end). This gets rid of the code duplication. I also changed the compile_at_points interface argument to take an ExpressionKernelBuilder module (as in compile_integral), rather than an interface file.

tsfc/kernel_interface/themis.py
```diff
@@ -0,0 +1,411 @@
...
+        if self.oriented:
+            args.append(cell_orientations_coffee_arg)
+        if self.cell_sizes:
+            args.append(self.cell_sizes_arg)
+        args.extend(self.kernel_args)
+
+        for name_, shape in self.tabulations:
+            args.append(coffee.Decl(self.scalar_type, coffee.Symbol(
+                name_, rank=shape), qualifiers=["const"]))
+# POSSIBLY REGISTER A QUAD RULE HERE?
```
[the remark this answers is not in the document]
celdred: It does not! But the Themis interface is gone :)

celdred:
Note: The new compile_at_points requires https://github.com/firedrakeproject/firedrake/commit/db641d6168b1d11699e9e2e1b8e41a89b9ff6a30 in Firedrake (ie ignore the new list of tabulations that is returned). Thoughts?

tsfc/kernel_interface/firedrake.py
```diff
@@ -265,6 +278,14 @@ def set_coefficients(self, integral_data, form_data):
                 self._coefficient(coefficient, "w_%d" % i))
         self.kernel.coefficient_numbers = tuple(coefficient_numbers)
 
+    def register_tabulations(self, expressions):
+        tabulations = {}
+        for node in traversal(expressions):
+            if isinstance(node, gem.Variable) and node.name.startswith("rt_"):
+                tabulations[node.name] = node.shape
+        self.tabulations = tuple(sorted(tabulations.items()))
+        self.kernel.tabulations = tuple(sorted(tabulations.items()))
```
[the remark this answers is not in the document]
celdred: Fixed!

tsfc/kernel_interface/ufc.py
```diff
@@ -102,16 +102,20 @@ def set_coefficients(self, integral_data, form_data):
...
         :returns: a COFFEE function definition object
         """
+
+        assert(quadrature_rule is None)
```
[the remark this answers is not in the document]
celdred: Done!

celdred:
I tried to squash the commits together, but clearly did not do the correct thing. How exactly do I accomplish this?

tsfc/driver.py
```diff
@@ -252,10 +255,10 @@ def name_multiindex(multiindex, name):
     body = generate_coffee(impero_c, index_names, parameters["precision"],
                            parameters["scalar_type"], expressions, split_argument_indices)
 
-    return builder.construct_kernel(kernel_name, body)
+    return builder.construct_kernel(kernel_name, body, quad_rule)
```
miklos1: This is kind of bugged: `quad_rule` comes from a loop over quadratures, but here outside the loop it only references the value in the last iteration. It works as long as there is only one quadrature in an integral.

tsfc/kernel_interface/ufc.py
```diff
@@ -102,16 +102,20 @@ def set_coefficients(self, integral_data, form_data):
...
         This function contains the logic for building the argument
         list for assembly kernels.
 
         :arg name: function name
         :arg body: function body (:class:`coffee.Block` node)
+        :arg quadrature rule: quadrature rule (not used, stubbed out for Themis integration)
         :returns: a COFFEE function definition object
         """
+
+        assert quadrature_rule is None
```
miklos1: This assertion just breaks UFC...

tsfc/driver.py
celdred: This is very true! It We only need quad_rule in order to figure out which points to tabulate at for runtime tabulation. A better way to do this would be to somehow attach this information to the tabulations themselves. I am not sure how to do this though...

wence-: So what is the kernel interface right now? You make a quad rule so that TSFC knows what size the tabulations are (and that's it?). What part of themis uses the quad rule that comes back in `construct_kernel`?

celdred: I think TSFC already knows the size of the tabulations (they are just GEM nodes?). Themis uses the quad_rule to determine which points to tabulate at (using the shift axis and restriction), since the tabulations don't actually know this....

wence-: Well, what happens is that the quad rule comes in (with some points). This point set is provided to the element for tabulation, which constructs the shape of the tabulation matrix like so:

```python
        shape = tuple(index.extent for index in ps.indices) + self.index_shape + self.value_shape
```

where `ps.indices` are the indices used to index the quadrature rule's point set.

celdred: Is there a way to associate these points with the tabulation (specifically the GEM node) within TSFC/FInAT?

wence-: Right now, FInAT tabulates like this:

```python
        for derivative in range(order + 1):
            for alpha in mis(dimension, derivative):
                name = str.format("rt_{}_{}_{}_{}_{}_{}",
                                  self.variant,
                                  self.degree,
                                  ''.join(map(str, alpha)),
                                  self.shift_axes,
                                  'c' if self.continuous else 'd',
                                  {None: "",
                                   '+': "p",
                                   '-': "m"}[self.restriction])
                result[alpha] = gem.partial_indexed(gem.Variable(name, shape), ps.indices)
```

You would, I suppose, need to introduce a new node in gem. Say `gem.Tabulation`? that behaves like a variable, but carries that information. Arguably, this might things cleaner because the encoding of the tabulation information now doesn't live in the string name (I don't know if you use that at all).

miklos1:
I feel like this is the occasion to make use of dynamical scoping
(see `gem.utils.DynamicallyScoped`).  We could declare a dynamically
scoped variable next to RuntimeTabulated, the FInAT element, and the
FInAT element could record the quadrature points as a mapping from the
element names.

The dynamic scoping would ensure that these quadrature points not stick
around forever in global variables, but live only within a certain
dynamic context.

From Themis, one should then call the form compiler from within a
dynamic-let block.  Once the form compiler returned, but still within
the dynamic-let block, one could inspect and retrieve the quadrature
points that were given to RuntimeTabulated.  Leaving the block, then
everything is automatically cleaned up.

celdred:
I prefer the idea of creating a new GEM node (gem.tabulation), that can carry around the information currently in the string AND the actual points that are tabulated at. I will have a go at coding that up once the loopy changes land. 

What would be the advantages of using dynamic scoping instead?

miklos1:
When there are multiple quadratures in the same integral, it is likely
that the same element configuration needs to be tabulated at different
sets of quadrature points, which means that each of those tabulations
need a unique name for the kernel argument.  The dynamic context makes
it easy to inspect what other tabulations have been requested on the
same element configuration, so the naming issue could be solved without
needing to invent an elaborate name encoding of quadrature rules.

celdred:
Yes this issue has recently been discovered, in the context of CellVolume for facet integrals. Right now each gem Variable node carries information about name (which has variant, derivorder, shiftaxis, degree, continuity and restriction currently in it) and shape, which as you noted is not sufficient to uniquely identify tabulations.

I was planning on creating a new gem node Tabulation, that is a subclass of variable, and holds shape, variant, derivorder, shiftaxis, degree, continuity, restriction and point set in it (with all of these in __front__ except point set). Is this sufficient to unique identify tabulations?

miklos1:
This is, of course, enough to uniquely identify tabulations, but being
`gem.Variable`s (i.e. kernel arguments), they still need to have a
_unique_ name in the generated C code.

Currently, the name includes everything except for the quadrature rule,
which means that different tabulations would have the same name when
multiple quadratures are used.  Clearly, something related to the
quadrature point set has to be added to be able to distinguish between
different tabulations of the same element.

With the dynamic context, one could just number the different
tabulations of the same element.  Without that context, you would need
to come up with a scheme to encode the quadrature rule in the name.

celdred:
Ah that makes sense, I'll take a look at gem.utils.DynamicallyScoped.

celdred:
Uggh it is even uglier than we thought- there are of course multiple copies of the same (identical variant, shiftaxis, degree, continuity and restriction) runtime tabulated elements that are being tabulated (sometimes at the same points and derivorder, sometimes different). Ideally each unique combination of variant, shiftaxis, degree, continuity and restriction plus derivorder/pts should generate the "same" gem Variable/kernel argument.

I looked at gem.utils.DynamicallyScoped, I don't understand what it is doing! I am concerned about maintainability if we go that approach...

What about this idea:
Runtime Tabulated is a singleton (one per set of variant, shiftaxis, degree, continuity and restriction; per invocation of compile_form/compile_at_points), and one dictionary and one counter for each deriv order (0, 1, 2). Every time a tabulation is requested it checks the dictionary for the pts, and if not found, increments the counter and stores the pts. Then it generates a Tabulation node (that is a subclass of Variable) that stores variant, derivorder, shiftaxis, degree, continuity, restriction and pts index, plus the pts themselves.

Thus each combination of variant, shiftaxis, degree, continuity, restriction, derivorder and pts generates the "same" gem Variable/kernel argument.

miklos1:
Dynamical scoping is, in its original sense, a way of resolving variable
references.  By now, lexical scoping has been established as the
standard approach, and for good, since lexical scoping is much more
sensible in the general case.

Nevertheless, there are still some particular use cases, when dynamical
scoping is handy.  However, it is little known, for most popular
languages nowadays lack support for dynamical scoping.  The most popular
ones, that I know of, and have this feature, are Perl and Common Lisp.

Python does not have this feature either, but
gem.utils.DynamicallyScoped provides a library implementation.  Sadly,
that class is not really documented yet.

variable = DynamicallyScoped()

defines a dynamically scoped variable with no global value.

variable = DynamicallyScoped(42)

defines a dynamically scoped variable with 42 as its "default" global
value.  The value of the `variable` should always be referenced as
`variable.value`.  Dynamic extents can be defined using the `.let(...)`
method with 'with' blocks.

An example can enlighten this much better:

```python
var = DynamicallyScoped(13)

def f():
    print(var.value)

f()  # prints 13

with var.let(42):
     # within this block, the value of var is 42

     f()  # prints 42

# leaving the dynamic let block, var reverts to its previous value

f()  # prints 13
```

The same example without default value:


```python
var = DynamicallyScoped()

def f():
    print(var.value)

f()  # error, var unset

with var.let(42):
     f()  # ok, prints 42

f()  # error, var unset
```

About the idea you described.  1) Not quite sure what you mean by
singleton.  2) "per invocation of compile_form/compile_at_points": how
do you ensure that the naming / numbering is _per invocation_?  This is
exactly where dynamic variables might help, because if you wrap your
call to compile_form with a dynamic-let block (`with var.let(value):`),
then you numbering is automatically reset when you leave the block.
3) The rest I agree with.

PR #188, opened 2019-03-26
Refactor kernel interface for collecting external data dependencies

miklos1:


wence-:
Yes, thanks!

miklos1:
Moreover it unbreaks UFC... (assuming that Michal's branch is used on the FFC side, which adopts the previous interface changes)

PR #189, opened 2019-04-03
Add kernel interface for loopy codegen

wence-:
Supersedes #159, but is bisectable (because I squashed everything).

### DIFF
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -2,6 +2,7 @@ import collections
 import operator
 import string
 import time
+import sys
 from functools import reduce
 from itertools import chain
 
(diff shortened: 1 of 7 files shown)


tsfc/driver.py
```diff
@@ -23,20 +24,24 @@
...
 from tsfc.fiatinterface import as_fiat_cell
 from tsfc.logging import logger
 from tsfc.parameters import default_parameters, is_complex
 
-import tsfc.kernel_interface.firedrake as firedrake_interface
+import tsfc.kernel_interface.firedrake as firedrake_interface_coffee
+import tsfc.kernel_interface.firedrake_loopy as firedrake_interface_loopy
 
+# To handle big forms. The various transformations might need a deeper stack
+sys.setrecursionlimit(3000)
```
miklos1: See [`coffee_init`](https://github.com/coneoproject/COFFEE/blob/master/coffee/__init__.py#L137). Original commit seems to be `d017e1ba69b809a45ed2eee560951f4c4e81f664`.

It seems that Fabio chose the easy way out of deep expressions back then, and as a bonus, all of Firedrake got to enjoy the possibly deeper call stacks.  Now that COFFEE is retiring, we noticed that other codes also rely on deep call stacks.

miklos1: Nevertheless, it might be a good idea to check whether those deep expressions are really unavoidable. There are a couple of `recude(Sum, ...)` and `reduce(Product, ...)` time expressions builders, if they many arguments, then there is an easy way to reduce expression depth.

wence-: The problem does not appear to be this. TSFC doesn't really blow the tree depth up. But rather the incoming UFL expression has tree depth of around 140.

`compile_ufl` then produces a slightly deeper tree, which is then (after optimisation) much shallower.

So the question is whether the deep expressions in UFL are unavoidable. They come from terms like `div(grad(div(grad(...)))` on H^2 elements, and appear after apply pullback. It's not obvious to me that this isn't just how deep the tree needs to be.

miklos1: Well then. Thanks for investigating this...

PR #190, opened 2019-04-29
Serendipity

cyruscycheng21:
Reconstruct tensor product cells as corresponding UFC quads for the DPC and Serendipity elements


dham:
The test fails are spurious - it's just branch matching. I have verified this locally.

PR #192, opened 2019-06-14
Interpolation operator

dham:


dham:
Note that this needs to land at the same time as https://github.com/firedrakeproject/firedrake/pull/1453

tsfc/driver.py
```diff
@@ -257,13 +257,15 @@ def name_multiindex(multiindex, name):
...
-def compile_expression_at_points(expression, points, coordinates, interface=None, parameters=None, coffee=True):
+def compile_expression_at_points(expression, points, to_element, coordinates, interface=None,
+                                 parameters=None, coffee=True):
     """Compiles a UFL expression to be evaluated at compile-time known
     reference points.  Useful for interpolating UFL expressions onto
     function spaces with only point evaluation nodes.
 
     :arg expression: UFL expression
     :arg points: reference coordinates of the evaluation points
+    :arg to_element: UFL element to interpolate onto
```
miklos1: This new argument seems to be ignored.

thomasgibson: I agree with @miklos1, I don't think this is needed at all. Permission to update and remove so we can move forward with https://github.com/firedrakeproject/firedrake/pull/1453?

PR #193, opened 2019-06-20
Unbreak some design mistakes from the serendipity project.

dham:


tsfc/fiatinterface.py
```diff
@@ -140,16 +140,7 @@ def convert_finiteelement(element, vector_is_mixed):
...
-        if element.cell().geometric_dimension() == 2:
-            element = element.reconstruct(cell=ufl.hypercube(2))
-        elif element.cell.geometric_dimension() == 3:
-            element = element.reconstruct(cell=ufl.hypercube(3))
-    elif element.family() == "S":
-        if element.cell().geometric_dimension() == 2:
-            element = element.reconstruct(cell=ufl.hypercube(2))
-        elif element.cell().geometric_dimension() == 3:
-            element = element.reconstruct(cell=ufl.hypercube(3))
+
```
wence-: This was a sin why?

dham: Because lying about the element in this way causes breakages all over the codebase. The vastly better answer we adopted was to change FIAT to enable the relevant elements to be created on either tensor product or flat cells.

wence-: Thanks!

PR #203, opened 2019-12-03
Translation of gem's ComponentTensors/Solve/Inverse into loops in loopy.

sv2518:
Slate has global operations on matrices, which ufl has not. In the two stage slate translation, we compile slate to gem to loopy. For the gem to loopy stage we need to allow ComponentTensors to be resolved later in the compilation process. In order to do so, we need to introduce how to compile this gem structure into a loopy structure. My suggestion here is to treat ComponentTensor like a loop.

### DIFF
--- a/gem/gem.py
+++ b/gem/gem.py
@@ -562,7 +562,7 @@ class FlexiblyIndexed(Scalar):
     def __init__(self, variable, dim2idxs):
         """Construct a flexibly indexed node.
 
-        :arg variable: a :py:class:`Variable`
+        :arg variable: a node that has a shape
         :arg dim2idxs: describes the mapping of indices
 
         For example, if ``variable`` is rank two, and ``dim2idxs`` is
@@ -574,7 +574,7 @@ class FlexiblyIndexed(Scalar):
             variable[1 + i*12 + j*4 + k][0]
 
         """
-        assert isinstance(variable, Variable)
+        assert variable.shape
         assert len(variable.shape) == len(dim2idxs)
 
         dim2idxs_ = []
--- a/gem/gem.py
+++ b/gem/gem.py
@@ -32,7 +32,8 @@ __all__ = ['Node', 'Identity', 'Literal', 'Zero', 'Failure',
            'Index', 'VariableIndex', 'Indexed', 'ComponentTensor',
            'IndexSum', 'ListTensor', 'Concatenate', 'Delta',
            'index_sum', 'partial_indexed', 'reshape', 'view',
-           'indices', 'as_gem']
+           'indices', 'as_gem', 'FlexiblyIndexed',
+           'Inverse', 'Solve']
 
 
 class NodeMeta(type):
@@ -780,6 +781,44 @@ class Delta(Scalar, Terminal):
         free_indices = tuple(index for index in (i, j) if isinstance(index, Index))
         self.free_indices = tuple(unique(free_indices))
         return self
+
+
+class Inverse(Node):
+    """The inverse of a square matrix."""
+    __slots__ = ('children', 'shape')
+
+    def __new__(cls, tensor):
+        assert len(tensor.shape) == 2
+        assert tensor.shape[0] == tensor.shape[1]
+
+        # Invert 1x1 matrix
+        if tensor.shape == (1, 1):
+            multiindex = (Index(), Index())
+            return ComponentTensor(Division(one, Indexed(tensor, multiindex)), multiindex)
+
+        self = super(Inverse, cls).__new__(cls)
+        self.children = (tensor,)
+        self.shape = tensor.shape
+
+        return self
+
+
+class Solve(Node):
+    """Solution of a square matrix equation with (potentially) multiple right hand sides.
+
+    Represents the X obtained by solving AX = B.
+    """
+    __slots__ = ('children', 'shape')
+
+    def __init__(self, A, B):
+        # Shape requirements
+        assert B.shape
+        assert len(A.shape) == 2
+        assert A.shape[0] == A.shape[1]
+        assert A.shape[0] == B.shape[0]
+
+        self.children = (A, B)
+        self.shape = A.shape[1:] + B.shape[1:]
 
 
 def unique(indices):
--- a/gem/optimise.py
+++ b/gem/optimise.py
@@ -10,7 +10,7 @@ import numpy
 from gem.utils import groupby
 from gem.node import (Memoizer, MemoizerArg, reuse_if_untouched,
                       reuse_if_untouched_arg, traversal)
-from gem.gem import (Node, Terminal, Failure, Identity, Literal, Zero,
+from gem.gem import (Node, Failure, Identity, Literal, Zero,
                      Product, Sum, Comparison, Conditional, Division,
                      Index, VariableIndex, Indexed, FlexiblyIndexed,
                      IndexSum, ComponentTensor, ListTensor, Delta,
@@ -128,7 +128,6 @@ def replace_indices_indexed(node, self, subst):
 @replace_indices.register(FlexiblyIndexed)
 def replace_indices_flexiblyindexed(node, self, subst):
     child, = node.children
-    assert isinstance(child, Terminal)
     assert not child.free_indices
 
     substitute = dict(subst)
--- a/tsfc/loopy.py
+++ b/tsfc/loopy.py
@@ -14,10 +14,13 @@ import islpy as isl
 import loopy as lp
 
 import pymbolic.primitives as p
+from loopy.symbolic import SubArrayRef
 
 from pytools import UniqueNameGenerator
 
 from tsfc.parameters import is_complex
+
+from contextlib import contextmanager
 
 
 class LoopyContext(object):
@@ -28,7 +31,7 @@ class LoopyContext(object):
         self.gem_to_pymbolic = {}  # gem node -> pymbolic variable
         self.name_gen = UniqueNameGenerator()
 
-    def pym_multiindex(self, multiindex):
+    def fetch_multiindex(self, multiindex):
         indices = []
         for index in multiindex:
             if isinstance(index, gem.Index):
@@ -40,25 +43,69 @@ class LoopyContext(object):
                 indices.append(index)
         return tuple(indices)
 
+    # Generate index from gem multiindex
+    def gem_to_pym_multiindex(self, multiindex):
+        indices = []
+        for index in multiindex:
+            assert index.extent
+            if not index.name:
+                name = self.name_gen(self.index_names[index])
+            else:
+                name = index.name
+            self.index_extent[name] = index.extent
+            indices.append(p.Variable(name))
+        return tuple(indices)
+
+    # Generate index from shape
+    def pymbolic_multiindex(self, shape):
+        indices = []
+        for extent in shape:
+            name = self.name_gen(self.index_names[extent])
+            self.index_extent[name] = extent
+            indices.append(p.Variable(name))
+        return tuple(indices)
+
+    # Generate pym variable from gem
+    def pymbolic_variable_and_destruct(self, node):
+        pym = self.pymbolic_variable(node)
+        if isinstance(pym, p.Subscript):
+            return pym.aggregate, pym.index_tuple
+        else:
+            return pym, ()
+
+    # Generate pym variable or subscript
     def pymbolic_variable(self, node):
+        pym = self._gem_to_pym_var(node)
+        if node in self.indices:
+            indices = self.fetch_multiindex(self.indices[node])
+            if indices:
+                return p.Subscript(pym, indices)
+        return pym
+
+    def _gem_to_pym_var(self, node):
         try:
             pym = self.gem_to_pymbolic[node]
         except KeyError:
             name = self.name_gen(node.name)
             pym = p.Variable(name)
             self.gem_to_pymbolic[node] = pym
-        if node in self.indices:
-            indices = self.pym_multiindex(self.indices[node])
-            if indices:
-                return p.Subscript(pym, indices)
-            else:
-                return pym
-        else:
-            return pym
+        return pym
 
     def active_inames(self):
         # Return all active indices
         return frozenset([i.name for i in self.active_indices.values()])
+
+
+@contextmanager
+def active_indices(mapping, ctx):
+    """Push active indices onto context.
+   :arg mapping: dict mapping gem indices to pymbolic index expressions
+   :arg ctx: code generation context.
+   :returns: new code generation context."""
+    ctx.active_indices.update(mapping)
+    yield ctx
+    for key in mapping:
+        ctx.active_indices.pop(key)
 
 
 def generate(impero_c, args, precision, scalar_type, kernel_name="loopy_kernel", index_names=[]):
@@ -141,13 +188,9 @@ def statement_for(tree, ctx):
     extent = tree.index.extent
     assert extent
     idx = ctx.name_gen(ctx.index_names[tree.index])
-    ctx.active_indices[tree.index] = p.Variable(idx)
     ctx.index_extent[idx] = extent
-
-    statements = statement(tree.children[0], ctx)
-
-    ctx.active_indices.pop(tree.index)
-    return statements
+    with active_indices({tree.index: p.Variable(idx)}, ctx) as ctx_active:
+        return statement(tree.children[0], ctx_active)
 
 
 @statement.register(imp.Initialise)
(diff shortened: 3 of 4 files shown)


tsfc/loopy.py
```diff
@@ -190,6 +190,18 @@ def statement_evaluate(leaf, ctx):
         return ops
     elif isinstance(expr, gem.Constant):
         return []
+    elif isinstance(expr, gem.ComponentTensor):
+        idx=tuple()
+        for c,index in enumerate(expr.multiindex):
```
dham: space after comma

gem/gem.py
```diff
@@ -782,6 +783,79 @@ def __new__(cls, i, j):
...
 
+# TODO: inverse should be different from comp tensor
+# in the sense that it does not take a local to a global view of a matrix
+# but instead eats a global view and spits out another global view
+# it does not transformation between free indices and shape so you would think it does not need a multiindex
+# but the multiindex is necessary to feed in the global view of the tensor
+# it can still have free_indices inherited by the  aggregate (?)
+class Inverse(Node):
+    __slots__ = ('children', 'multiindex', 'shape')
+    __back__ = ('multiindex',)
```
miklos1, line 794: `multiindex`? Isn't it more natural to conceive `Inverse` as a shape-to-shape operation?

gem/gem.py
```diff
@@ -782,6 +783,79 @@ def __new__(cls, i, j):
...
+            return Zero()
+
+        self = super(Determinant, cls).__new__(cls)
+        self.children = (aggregate,)
+
+        return self
+
+class Solve(Node):
+    __slots__ = ('children', 'multiindex', 'shape')
+    __back__ = ('multiindex',)
```
miklos1, line 840: Same here.

gem/gem.py
```diff
@@ -782,6 +783,79 @@ def __new__(cls, i, j):
         return self
 
 
+# TODO: inverse should be different from comp tensor
+# in the sense that it does not take a local to a global view of a matrix
+# but instead eats a global view and spits out another global view
+# it does not transformation between free indices and shape so you would think it does not need a multiindex
+# but the multiindex is necessary to feed in the global view of the tensor
+# it can still have free_indices inherited by the  aggregate (?)
```
miklos1, line 791: Can you explain what is meant by _local_ and _global_ view of a matrix?  I have no idea what this text means...

gem/gem.py
```diff
@@ -782,6 +783,79 @@ def __new__(cls, i, j):
...
+        # Set index extents from shape
+        for index, extent in zip(multiindex, aggregate.shape):
+            assert isinstance(index, Index)
+            index.set_extent(extent)
+
+        self = super(Inverse, cls).__new__(cls)
+        self.children = (aggregate,)
+        self.shape = aggregate.shape
+        self.multiindex = multiindex
+        self.free_indices = aggregate.free_indices
```
miklos1, line 808: So you inherit shape and inherit free indices (both are fine), but there is still this multiindex...

gem/gem.py
```diff
@@ -782,6 +783,79 @@ def __new__(cls, i, j):
...
+class Determinant(Scalar):
+    __slots__ = ('children', )
+
+    def __new__(cls, aggregate):
+        # det of scalar valued constant is aggregate itself
+        if isinstance(aggregate, Constant):
+            return aggregate
+
+        # tensor must be square
+        assert not aggregate.shape
```
miklos1, line 826: That means that `aggregate` must be a scalar...

gem/gem.py
```diff
@@ -782,6 +783,79 @@ def __new__(cls, i, j):
...
+        assert all(aggregate.shape) == aggregate.shape[0], "Tensor must be square"
```
miklos1, line 827: That is not how `all` works, you probably meant `set`. Hint: this asserts that if no dimensions are 0, then the first dimension must be 1, if any dimension is zero, then the first dimension must be zero... What?...

tsfc/loopy.py
```diff
@@ -190,6 +193,73 @@ def statement_evaluate(leaf, ctx):
...
+        for c, index in enumerate(expr.multiindex):
+            ctx.active_indices.pop(index)
+        return statements
+    elif isinstance(expr, gem.Inverse):
+        idx = tuple()
+        idx2 = tuple()
+        for c, index in enumerate(expr.multiindex):
+            ctx.active_indices[index] = p.Variable(index.name)
+            ctx.index_extent[index.name] = index.extent
+            idx = idx + (ctx.active_indices[index], )
```
miklos1, line 221: Ah, so you use that multiindex for code generation...

gem/gem.py
miklos1: Okay, I think I'm staring to get it... by *global* you probably mean shape and by *local* you mean free indices.

miklos1: So, yes, I agree that `ComponentTensor` is different from `Inverse`.

tsfc/loopy.py
sv2518: Yes, I use the indices (i0,i1) of the thing the Inverse acts on (t0) to feed it it into the inverse function call in loopy. The multiindices (i2,i3) of the gem Inverse node are for the left hand side (t1). So `t1[i2,i3]=inverse(t0[i0,i1])` would be the statement in loopy. When the loopy code is translated to, however, `t1[i2,i3]` gets dropped since the inverse is void and `t0` is in/out parameter. If I find a way to call the inverse without assigning it to an lhs in loopy, we could drop the multi index in the gem nodes.

gem/gem.py
sv2518: Yes, sorry for terrible phrasing.

tsfc/loopy.py
```diff
@@ -43,41 +43,52 @@ def pym_multiindex(self, multiindex):
...
-def active_indices(gem_indices, pym_indices, lambda_statement, ctx):
+def active_indices(gem_indices, pym_indices, ctx):
+    """ Context manager function to keep track of active indices.
+    Note: Does not work for recursion, hence cannot be used for statement_for.
+    """
     for gem_index, pym_index in zip(gem_indices, pym_indices):
         ctx.active_indices[gem_index] = pym_index
-    yield lambda_statement(ctx)
+    yield ctx
     for index in gem_indices:
```
wence-, line 92: It's not clear to me why this doesn't work with recursion (you in-place modify `ctx`) so I think:
```python
with active_indices(a, b, ctx):
   do_something(ctx)
   with active_indices(c, d, ctx):
      do_other(ctx)
```

Should see `a -> b` in `do_something` and `a ->b, c -> d` in `do_other`

sv2518: Ok

wence-: BTW, is it obvious why sometimes one has to do `pym_index = ctx.name_gen(idx)` and sometimes just `pym_index = p.Variable(idx.name)` ?

sv2518: No, this is related to something that Miklos critiqued on the tsslac PR. I name all indices at gem stage so I can just do the latter at the moment. I will change that here as soon as I postponed the index naming to this stage.

gem/gem.py
sv2518: Addressed in c2a473f24b50396278f27cd4ecd1a6d4bfef0219

tsfc/loopy.py
```diff
@@ -40,26 +43,67 @@ def pym_multiindex(self, multiindex):
...
+                extent = index.extent
+            else:
+                name = self.name_gen(self.index_names[index])
+                extent = index
+            self.index_extent[name] = extent
+            pym_idx += (p.Variable(name), )
+        return pym_idx
+
+
+@contextmanager
```
sv2518, line 97: This is only available from python3.7 and we test Tsfc with python3.5. Should I use something else for the context manager?

miklos1: `contextmanager` has been around at least since Python 2.7

sv2518: Oh what, okay, seems like I misread google. Not quite sure why the branch does not build then.

wence-:
```
pip._vendor.urllib3.exceptions.ProtocolError: ("Connection broken: ConnectionResetError(104, 'Connection reset by peer')", ConnectionResetError(104, 'Connection reset by peer'))
288The command "pip install -r requirements.txt" failed and exited with 2 during .``` You had a network failure on travis.

File: gem/gem.py [hunk 11]
```diff
@@ -782,6 +783,58 @@ def __new__(cls, i, j):
         return self
 
 
+class Inverse(Node):
+    """ Inverse is a shape to shape operation.
+
+        :arg aggregate: Tensor to be inversed
+        :arg multiindex: Indices for result of Inverse
```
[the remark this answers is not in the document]
miklos1: Furthermore, the convention for other GEM nodes seems to be that class docstring contains only the description of the class, and description of constructor arguments go to the docstring of the constructor.

gem/gem.py
```diff
@@ -782,6 +783,58 @@ def __new__(cls, i, j):
...
+
+        Inverse is different from ComponentTensor, because its children must have shape.
+        It does no transformation between free indices and shape, but the multiindex is still necessary.
+        The multiindices of the children are used for indexing them when passing to inverse function call,
+        the multiindex of the Inverse node itself is used to index the result of the function call.
```
miklos1, line 795: Break your lines at 80 characters unless that becomes really inconvenient.

miklos1, line 795: Still these outdated multiindex comments...

miklos1: It is going to be more future-proof to just clearly explain what `Inverse` is, rather than how it related to `ComponentTensor`.  (What does `Inverse` have to do with `ComponentTensor` at all?)

gem/gem.py
```diff
@@ -782,6 +783,58 @@ def __new__(cls, i, j):
...
+    """
+    __slots__ = ('children', 'shape')
+    __back__ = ()
+
+    def __new__(cls, aggregate):
+        assert aggregate.shape
```
miklos1, line 801: This check seems pretty weak, only checks that the node is non-scalar.  You would expect square matrices, no?  Then write instead:
```python
assert len(aggregate.shape) == 2
assert aggregate.shape[0] == aggregate.shape[1]
```

miklos1: Also, why is it called `aggregate`? Just call it `matrix` or something... (if that's what you expect)

gem/gem.py
```diff
@@ -782,6 +783,58 @@ def __new__(cls, i, j):
...
+
+        self = super(Inverse, cls).__new__(cls)
+        self.children = (aggregate,)
+        self.shape = aggregate.shape
+        self.free_indices = aggregate.free_indices
```
miklos1, line 806: I think this is redundant, because `NodeMeta` takes care of setting this same value.

gem/gem.py
```diff
@@ -782,6 +783,58 @@ def __new__(cls, i, j):
...
+
+        return self
```
miklos1, line 808: You might as well just use `__init__` instead of `__new__`, since you never actually return any other type than an `Inverse`.  That spares you two lines: super-new and return-self.

tsfc/loopy.py
```diff
@@ -40,26 +43,67 @@ def pym_multiindex(self, multiindex):
...
+                else:
+                    name = index.name
+                assert index.extent
+                extent = index.extent
+            else:
+                name = self.name_gen(self.index_names[index])
+                extent = index
+            self.index_extent[name] = extent
+            pym_idx += (p.Variable(name), )
+        return pym_idx
```
wence-, line 94: So this is to generate index names and extents for some expression. It looks like it is only ever called on shaped objects, but one of them `ComponentTensor` has bound its indices already, and the others don't have indices in the gem representation. So does it not make more sense to use `pym_multiindex` in the component tensor code path and just use the shape-based code path for the other cases?

tsfc/loopy.py
```diff
@@ -181,15 +224,44 @@ def statement_evaluate(leaf, ctx):
...
+    elif isinstance(expr, gem.Solve):
+        idx = ctx.pymbolic_indices(expr)
+        var = ctx.pymbolic_variable(expr)
+        lhs = (SubArrayRef(idx, p.Subscript(var, idx)),)
+
+        reads = ()
+        for child in expr.children:
+            idx_reads = ctx.pymbolic_indices(child)
+            var_reads = ctx.pymbolic_variable(child)
+            reads += (SubArrayRef(idx_reads, p.Subscript(var_reads, idx_reads)),)
```
wence-, line 261: For collecting in a list, I prefer making a list and appending. You can cast to tuple after the loop.

tsfc/loopy.py
sv2518: Pym_multiindex does not generate the pymbolic Variable for the index at the moment. Do you just want me to move the code for the ComponentTensor case to pym_multiindex?

tsfc/loopy.py
```diff
@@ -40,26 +43,67 @@ def pym_multiindex(self, multiindex):
...
 
-    def pymbolic_variable(self, node):
+    def gem_to_pym(self, node):
         try:
             pym = self.gem_to_pymbolic[node]
         except KeyError:
             name = self.name_gen(node.name)
             pym = p.Variable(name)
             self.gem_to_pymbolic[node] = pym
+        return pym
```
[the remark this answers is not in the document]
sv2518: I think this does not work because the defaultdict constructor doesn't want a factory that takes an argument.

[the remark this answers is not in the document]
wence-: Oh yeah, I always forget that, sorry.

tsfc/loopy.py
wence-: So I think if you do that, then it's clearer in the Solve and Inverse case that you're making new indices rather than turning gem indices into pymbolic ones?

tsfc/loopy.py
```diff
@@ -40,26 +43,67 @@ def pym_multiindex(self, multiindex):
...
+            pym_idx += (p.Variable(name), )
+        return pym_idx
+
+
+@contextmanager
+def active_indices(gem_indices, pym_indices, ctx):
+    """ Context manager function to keep track of active indices.
+    """
+    for gem_index, pym_index in zip(gem_indices, pym_indices):
+        ctx.active_indices[gem_index] = pym_index
```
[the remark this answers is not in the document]
sv2518: I am popping the indices after yielding the ctx, so that we do not need the `saved`.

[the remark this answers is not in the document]
wence-: I guess I was thinking that there's no opposite to `dict.update(odict)` that removes `odict` from `dict`?

gem/gem.py
```diff
@@ -653,13 +654,13 @@ def __new__(cls, summand, multiindex):
...
-                              (0,) * len(unroll))
-            multiindex = tuple(index for index in multiindex
-                               if index not in unroll)
+        # unroll = tuple(index for index in multiindex if index.extent <= 1)
+        # if unroll:
+        #     assert numpy.prod([index.extent for index in unroll]) == 1
+        #     summand = Indexed(ComponentTensor(summand, unroll),
+        #                       (0,) * len(unroll))
+        #     multiindex = tuple(index for index in multiindex
+        #                        if index not in unroll)
```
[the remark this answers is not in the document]
sv2518: This did not work for a reason I do not remember before I simplified the compiler a few weeks ago. I will take it back in.

gem/gem.py
```diff
@@ -782,6 +783,58 @@ def __new__(cls, i, j):
...
+        assert A.shape
+        assert B.shape
+
+        # Match shapes
+        assert A.shape[1] == B.shape[0]
+
+        self = super(Solve, cls).__new__(cls)
+        self.children = (A, B)
+        self.shape = B.shape
+        self.free_indices = B.free_indices
```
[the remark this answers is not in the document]
sv2518: Yes, this is not right indeed.
Your `A` should have two indices, no? And this is a contraction in the end, which is why we need to guarantee the shape matching you mentioned above, so we would also need to "loose" an index, right?
Maybe I should something like `self.free_indices = unique(set(A.free_indices) - set(B.free_indices))`?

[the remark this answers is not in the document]
wence-: Well, I thought we had decided that `Solve` operates on shape, so it represents `A^{-1} B` (and therefore the free indices of the output are the union of the free indices of the input). There's no contraction on the free indices of A and B, I think.

We have `A.shape = (N, N)` with some free indices, and `B.shape = (N, O, P, Q, ...)` with some free indices. So then the matching of shapes has to be (since inverting A flips the shape axes) `A.shape[0] == B.shape[0]` and so we end up with something of shape `(N, O, P, Q, ...)` but it still has free indices merging those of A and B.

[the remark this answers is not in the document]
sv2518: Ah right, yes.

tsfc/loopy.py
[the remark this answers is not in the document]
sv2518: I do pop in a one-line loop.

gem/optimise.py
```diff
@@ -128,7 +128,6 @@ def replace_indices_indexed(node, self, subst):
 @replace_indices.register(FlexiblyIndexed)
 def replace_indices_flexiblyindexed(node, self, subst):
     child, = node.children
-    assert isinstance(child, Terminal)
```
wence-, line 131: Need to remove the import of `Terminal` from this file now for flake8.

tsfc/loopy.py
```diff
@@ -40,27 +43,70 @@ def pym_multiindex(self, multiindex):
...
 
+@contextmanager
+def active_indices(mapping, ctx):
+    """Push active indices onto context.
+   :arg mapping: dict mapping gem indices to pymbolic index expressions
+   :arg ctx: code generation context.
+   :returns: new code generation context."""
+    ctx.active_indices.update(mapping)
+    yield ctx
+    [ctx.active_indices.pop(key) for key in mapping.keys()]
```
wence-, line 107: Sorry, I didn't explain well. I wasn't worried about the lines of code. I'm happy with popping, but can you do this as:
```python
for key in mapping:
    ctx.active_indices.pop(key)
```
Please?

Otherwise some later reader of the code might be confused why this side-effectful list construction is not returned anywhere.

tsfc/loopy.py
```diff
@@ -181,15 +223,43 @@ def statement_evaluate(leaf, ctx):
...
             ops.append(lp.Assignment(p.Subscript(var, index + multiindex), expression(value, ctx), within_inames=ctx.active_inames()))
         return ops
     elif isinstance(expr, gem.Constant):
         return []
+    elif isinstance(expr, gem.ComponentTensor):
+        idx = ctx.gem_to_pym_multiindex(expr.multiindex)
+        var, sub_idx = ctx.pymbolic_variable_and_destruct(expr)
+        lhs = p.Subscript(var, idx + sub_idx)
+        mapping = {i: j for i, j in zip(expr.multiindex, idx)}
+        with active_indices(mapping, ctx) as ctx_active:
```
wence-, line 237: AKA (doesn't really matter):
```python
with active_indices(dict(zip(expr.multiindex, idx)), ctx as ctx_active):
```

tsfc/loopy.py
sv2518: Yes sure I agree, I thought you would prefer a one-liner because you had removed my loops in your earlier suggestion.

wence-:
Thanks! I have two final requests (which are hopefully easy).

1. Can we squash this PR into logical commits (I can help with this)

2. Once done, can you kick off a firedrake builder that tests master against this branch to ensure we don't have any regressions?

gem/gem.py
```diff
@@ -782,6 +783,54 @@ def __new__(cls, i, j):
...
+        return self
+
+
+class Solve(Node):
+    """ Solve is a shape to shape operation similar to Inverse.
+
+        :arg A: Tensor
+        :arg B: Right-hand side
+        :arg multiindex: Indices for result of solve
+    """
```
miklos1, line 814: Can you move constructor arguments to the constructor docstring.  Plus we don't have here multiindex anymore.

gem/gem.py
```diff
@@ -782,6 +783,54 @@ def __new__(cls, i, j):
...
+    __slots__ = ('children', 'shape')
+    __back__ = ()
```
miklos1, line 816: Redundant.

gem/gem.py
```diff
@@ -782,6 +783,54 @@ def __new__(cls, i, j):
...
+
+    def __init__(self, A, B):
+
+        # Children must have shape
+        assert A.shape
+        assert B.shape
```
miklos1, line 822: This needs cleanup, `assert A.shape` is contained within `assert len(A.shape) == 2` etc.

gem/gem.py
```diff
@@ -782,6 +783,54 @@ def __new__(cls, i, j):
...
+
+        # Shape requirements
+        assert len(A.shape) == 2
+        assert A.shape[0] == A.shape[1]
+        assert A.shape[0] == B.shape[0]
+
+        self.children = (A, B)
+        self.shape = A.shape[1:] + B.shape[1:]
+        self.free_indices = A.free_indices + B.free_indices
```
miklos1, line 831: This is both unneeded and wrong. Wrong because shared free indices between A and B would be listed twice, and because free indices are not sorted to their canonical ordering.  It is not needed, because `NodeMeta` calculates automatically (and correctly).

wence-:
Jenkins pass on the Firedrake side: https://jenkins.ese.ic.ac.uk:1080/blue/organizations/jenkins/firedrake/detail/test-tsfc-against-master/2/pipeline

miklos1:
@sv2518: thanks, I'm happy with the new nodes now.  I have not looked too deeply on the code generation part, but I trust Lawrence's judgement on that.

Feel free to merge.

wence-:
Thanks!

PR #206, opened 2020-03-31
Pefarrell/compile expression dual evaluation

wence-:
General-purpose interpolation into spaces without derivative evaluation in their dual basis.


tsfc/driver.py
```diff
@@ -363,6 +363,118 @@ def compile_expression_at_points(expression, points, coordinates, interface=None
...
+    dual_expressions = []
+    for dual in dual_functionals:
+        # Extract associated quadrature rule representation
+        # Make a quad_rule
+
+        quad_points = [*dual.pt_dict.keys()]
+        # get it working for scalar functions first; how do we deal with components??
+        weights = [dual.pt_dict[pt][0][0] for pt in quad_points]
+        quad_rule = QuadratureRule(PointSet(quad_points), weights)
+
```
rckirby: I would assert that dual.deriv_dict is empty here until it is figured out how to address that.  (One can always push the burden back on FIAT to expand derivative-based functionals into linear combinations of point values and then have only a minor change here)

tsfc/driver.py
```diff
@@ -432,20 +432,51 @@ def compile_expression_dual_evaluation(expression, dual_functionals, coordinates
...
-        dual_expressions.append(expressions)
+    # We need to do different things based on the shape of the expression, unfortunately
+
+    # Case 1: scalars
+    if len(expression.ufl_shape) == 0:
+        for dual in dual_functionals:
+            quad_points = [*dual.pt_dict.keys()]
+            weights = [dual.pt_dict[pt][0][0] for pt in quad_points]
+            quad_rule = QuadratureRule(PointSet(quad_points), weights)
+
```
rckirby: See comment from first commit -- we need to assert that there is no deriv_dict or else this will fail silently.  Fixing this (and/or FIAT) to work with a deriv_dict is also a possibility.

tsfc/driver.py
```diff
@@ -519,7 +519,69 @@ def compile_expression_dual_evaluation(expression, to_element, coordinates, inte
...
+            for comp1 in range(expression.ufl_shape[0]):
+                for comp2 in range(expression.ufl_shape[1]):
+                    exp = expression[comp1, comp2]
+                    quad_points_comp = [q for q in quad_points if dual.pt_dict[q][0][1] == () or (comp1, comp2) in [w[1] for w in dual.pt_dict[q]]]
+                    weights_comp = []
+                    for q in quad_points_comp:
+                        weights_comp.extend([w[0] for w in dual.pt_dict[q] if dual.pt_dict[q][0][1] == () or (comp1, comp2) == w[1] ])
+                    if len(quad_points_comp) > 0:
+                        quad_rule = QuadratureRule(PointSet(quad_points_comp), weights_comp)
```
rckirby: Here and elsewhere:  It might be worth commenting that `QuadratureRule` is being used as a container rather than doing what one typically things of as integration?

wence-:
@rckirby I think you did not review the most recent version. I already handle those comments I think.

rckirby:
Hmmm...I clicked on the email link asking me to review.  I will look again.

tsfc/driver.py
```diff
@@ -269,117 +269,59 @@ def name_multiindex(multiindex, name):
...
+    :arg to_element: A FIAT FiniteElement for the target space
     :arg coordinates: the coordinate function
     :arg interface: backend module for the kernel interface
     :arg parameters: parameters object
     :arg coffee: compile coffee kernel instead of loopy kernel
     """
     import coffee.base as ast
     import loopy as lp
+    if any(len(dual.deriv_dict) != 0 for dual in to_element.dual_basis()):
+        raise NotImplementedError("Can only interpolate onto dual basis functionals without derivative evaluation, sorry!")
```
rckirby, line 324: Yes, I see you've addressed an earlier concern!

tsfc/driver.py
rckirby: Fixed in later comment I hadn't gotten too.

tsfc/driver.py
```diff
@@ -329,22 +370,68 @@ def compile_expression_at_points(expression, points, coordinates, interface=None
...
-    tensor_indices = tuple(gem.Index() for s in value_shape)
-    if value_shape:
-        ir = gem.Indexed(ir, tensor_indices)
+    kernel_cfg = dict(interface=builder,
+                      ufl_cell=coordinates.ufl_domain().ufl_cell(),
+                      precision=parameters["precision"],
+                      argument_multiindices=argument_multiindices,
+                      index_cache={})
+
+    if len(to_element.value_shape()) == 0:
```
rckirby, line 379: `len(to_element.value_shape()) == 0` refers to the `to_element` being scalar valued, which is different than all the nodes of the element being point-evaluation.  (Example, if someone wrote a Lagrange-like element but with integral moments over edges.

wence-: Would something like this be safer?
```suggestion
    only_pointeval = all(isinstance(dual, FIAT.functional.PointEvaluation) for dual in to_element.dual_basis())
    if only_pointeval:
```

rckirby:
Yes.

tsfc/driver.py
pefarrell: Should we commit this change?

rckirby: Yes.

wence-:
I fixed the missing import from my proposed change.

Additionally, we can't merge this until the matching Firedrake side is also done (because it removes a function that Firedrake needs).

In the interest of doing that, I've opened that merge at firedrakeproject/firedrake#1650.

tsfc/driver.py
```diff
@@ -267,21 +269,60 @@ def name_multiindex(multiindex, name):
     return builder.construct_kernel(kernel_name, impero_c, parameters["precision"], index_names, quad_rule)
 
 
-def compile_expression_at_points(expression, points, coordinates, interface=None,
-                                 parameters=None, coffee=True):
-    """Compiles a UFL expression to be evaluated at compile-time known
-    reference points.  Useful for interpolating UFL expressions onto
-    function spaces with only point evaluation nodes.
+def apply_mapping(expression, to_element):
```
dham, line 272: Please document this function call. "Apply mapping" is incredibly generic! Remember that there was a nontrivial discussion about pullbacks and push forwards when this code was written, so it's not totally obvious to a reader what this does.

tsfc/driver.py
```diff
@@ -267,21 +269,104 @@ def name_multiindex(multiindex, name):
...
```
miklos1, line 272: It wasn't immediately obvious for me that `expression` is UFL expression, and `to_element` is FIAT element...  However, `to_element` is only used to query the mapping, so I suggest the following slight change: pass in `mapping` string instead of FIAT element (so FIAT element -> mapping string code moves to the caller), then this function will depend on UFL only, and can be moved to `ufl_utils.py` where the other UFL processors are.

In `ufl_utils.py` you can also import Index and MultiIndex from `ufl.classes`, so no need to use fully-qualified names.

tsfc/driver.py
```diff
@@ -267,21 +269,104 @@ def name_multiindex(multiindex, name):
...
-    :arg points: reference coordinates of the evaluation points
+    :arg to_element: A FIAT FiniteElement for the target space
     :arg coordinates: the coordinate function
     :arg interface: backend module for the kernel interface
     :arg parameters: parameters object
     :arg coffee: compile coffee kernel instead of loopy kernel
     """
     import coffee.base as ast
     import loopy as lp
+    from FIAT.functional import PointEvaluation
```
miklos1, line 367: Please import at the top of the module.

I think the two imports above are here, because we import neither coffee nor loopy at module load time, so if someone only uses one of those two output formats, they don't need the other one installed as a dependency.

tsfc/driver.py
```diff
@@ -329,22 +415,67 @@ def compile_expression_at_points(expression, points, coordinates, interface=None
...
-    # Deal with non-scalar expressions
-    value_shape = ir.shape
-    tensor_indices = tuple(gem.Index() for s in value_shape)
-    if value_shape:
-        ir = gem.Indexed(ir, tensor_indices)
+    kernel_cfg = dict(interface=builder,
+                      ufl_cell=coordinates.ufl_domain().ufl_cell(),
+                      precision=parameters["precision"],
+                      argument_multiindices=argument_multiindices,
+                      index_cache={})
```
miklos1, line 422: An explicitly shared index cache between `compile_ufl` calls...

tsfc/driver.py
```diff
@@ -329,22 +415,67 @@ def compile_expression_at_points(expression, points, coordinates, interface=None
...
+
+    if all(isinstance(dual, PointEvaluation) for dual in to_element.dual_basis()):
```
miklos1, line 424: So the 'then' branch if basically equivalent to the old code path, and then the 'else' branch deals with the general case...

miklos1:
Aside from minor cosmetic suggestions, I think this is fine...

tsfc/driver.py
pefarrell: Good idea! Both done.

tsfc/driver.py
wence-: Yep.

PR #207, opened 2020-04-15
Vertex Only Mesh

ReubenHill:
Linked to https://github.com/firedrakeproject/firedrake/pull/1638,  https://github.com/FEniCS/fiat/pull/41 , and https://github.com/FEniCS/ufl/pull/30

### DIFF
--- a/tsfc/kernel_interface/firedrake.py
+++ b/tsfc/kernel_interface/firedrake.py
@@ -98,11 +98,18 @@ class KernelBuilderBase(_KernelBuilderBase):
         physically mapped elements (Argyris, Bell, etc...).  We need a
         measure of the mesh size around each vertex (hence this lives
         in P1).
+
+        Should the domain have topological dimension 0 this does
+        nothing.
         """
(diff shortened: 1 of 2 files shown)


tsfc/driver.py
```diff
@@ -328,7 +328,10 @@ def compile_expression_dual_evaluation(expression, to_element, coordinates, inte
     if domain:
         assert coordinates.ufl_domain() == domain
         builder.domain_coordinate[domain] = coordinates
-        builder.set_cell_sizes(domain)
+        if domain.ufl_cell().topological_dimension() != 0:
+            # Only P0 is a valid finite element if topological_dimension
+            # is 0 so set_cell_sizes is unnecessary
+            builder.set_cell_sizes(domain)
```
wence-, line 334: Why not do the appropriate thing in the builder? Keep the logic close to the implementation.

ReubenHill: done in 9a46507296f28504666bfee290c3068ebe1c9ec0

PR #209, opened 2020-04-25
View gem dag

narciuch:
Function to easily view a GEM DAG in formats such as .png

### DIFF
--- a/gem/view_gem_dag.py
+++ b/gem/view_gem_dag.py
@@ -1,0 +1,86 @@
+from enum import Enum
+from graphviz import Digraph
+from re import search
+
+class GraphType(Enum):
+    """Type of graph to create
+    PARENT_CHILD: Graph is composed of parent-child relationships only and each node is labelled with all information about the node (its type and all its slots including 'children').
+    DAG: DAG visualisation. Each node label is shortened for readability.
+    GRAPH: The bits of information about each node (i.e. its slots) are unique nodes in the graph. Note that the visualisation from the this case shows the expression graph as a graph rather than as the dag it truely is. This can increase the perceived size of the expression.
+    """
+    PARENT_CHILD = 1
+    DAG = 2
+    GRAPH = 3
+
+
+def view_gem_dag(expressions, filename='gem_dag', directory=None, graph_type=GraphType.GRAPH, format='png'):
+    """Create visual representation of the gem dag given by 'expressions'
+    
+    :arg expressions: a list of expressions
+    :arg filename: filename to save the graph under
+    :arg directory: directory to save the graph under
+    :arg graph_type: the GraphType to create
+    :arg format: output format of the graph. See https://www.graphviz.org/doc/info/output.html for options
+    """
+    print('\n')
+    print('=== In view_gem_dag ===')
+    g = Digraph(filename=filename, directory=directory, format=format)
+    stack = []
+
+    if graph_type == GraphType.PARENT_CHILD:
+        for expr_part in expressions:
+            stack.append(expr_part)
+            parent = 'root'
+            while stack:
+                expr = stack.pop()
+                g.edge(str(parent), str(expr))
+                stack.extend(expr.children)
+                parent = expr
+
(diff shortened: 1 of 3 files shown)


gem/view_gem_dag.py
```diff
@@ -0,0 +1,86 @@
+from enum import Enum
+from graphviz import Digraph
+from re import search
+
+class GraphType(Enum):
+    """Type of graph to create
+    PARENT_CHILD: Graph is composed of parent-child relationships only and each node is labelled with all information about the node (its type and all its slots including 'children').
```
miklos1, line 7: If I remember correctly, the correct Sphinx syntax for giving docstring for non-function members is a string in the line immediately following the assignment, but please look it up!

miklos1: Please break your lines.

Yes, we disable the flake8 rule which would force every line to be shorter than 80 characters, because sometimes that would make code really ugly, but for the most part we should keep the 80-character limit.

gem/view_gem_dag.py
```diff
@@ -0,0 +1,86 @@
...
+    DAG: DAG visualisation. Each node label is shortened for readability.
+    GRAPH: The bits of information about each node (i.e. its slots) are unique nodes in the graph. Note that the visualisation from the this case shows the expression graph as a graph rather than as the dag it truely is. This can increase the perceived size of the expression.
+    """
+    PARENT_CHILD = 1
+    DAG = 2
+    GRAPH = 3
+
+
+def view_gem_dag(expressions, filename='gem_dag', directory=None, graph_type=GraphType.GRAPH, format='png'):
```
miklos1, line 16: I see that you inherit this API of specifying filename and directory separately, nevertheless, it feels unusual.  Why does Graphviz do it this way?

gem/view_gem_dag.py
```diff
@@ -0,0 +1,86 @@
...
+    :arg format: output format of the graph. See https://www.graphviz.org/doc/info/output.html for options
+    """
+    print('\n')
+    print('=== In view_gem_dag ===')
+    g = Digraph(filename=filename, directory=directory, format=format)
+    stack = []
+
+    if graph_type == GraphType.PARENT_CHILD:
+        for expr_part in expressions:
+            stack.append(expr_part)
```
miklos1, line 32: AKA `stack = [expr_part]` since `stack` is always empty when we get to this line.

gem/view_gem_dag.py
```diff
@@ -0,0 +1,86 @@
...
+            parent = 'root'
+            while stack:
+                expr = stack.pop()
+                g.edge(str(parent), str(expr))
+                stack.extend(expr.children)
+                parent = expr
```
miklos1, line 38: I don't think this is correct.  The algorithm looks like depth-first search, but the assignment of parents - I think - is wrong.  If root has children A, B, etc., and A has further children, then B will get one of A's children as parent, not root.

gem/view_gem_dag.py
```diff
@@ -0,0 +1,86 @@
...
+    elif graph_type == GraphType.DAG:
+        for expr_part in expressions:
+            stack.append(expr_part)
+            parent_id = 'root' 
+            while stack:
+                expr = stack.pop()
+                expr_id = pretty_print_type(expr)
+                g.edge(parent_id, expr_id, label='parent of')
+                for slot in expr.__slots__:
+                    if slot != 'children':
```
miklos1, line 49: It's probably better to iterate through `__front__` and `__back__` instead...

gem/view_gem_dag.py
```diff
@@ -0,0 +1,86 @@
...
+    print('\n')
+    g.view()
+
+def pretty_print_type(expression):
+    expr = str(type(expression))
+    try:
+        class_name = search("(\w+?)'>", expr).group(1)
+    except AttributeError:
+        class_name = expr
+    return class_name
```
miklos1, line 86: Please finish your file with a newline.

Issue #208, opened 2020-04-24
Interpolating `QuadratureWeight` gives TSFC error


pefarrell:
I want to evaluate some expressions at quadrature points. I try:

```
from firedrake import *
from ufl.geometry import QuadratureWeight

mesh = UnitIntervalMesh(2)

Qse = FiniteElement("Quadrature", mesh.ufl_cell(), degree=2, quad_scheme="default")
Qs = FunctionSpace(mesh, Qse)

# Works
x = SpatialCoordinate(mesh)[0]
xq = interpolate(x, Qs)
print(xq.dat.data_ro)

# Fails
w = QuadratureWeight(mesh)
wq = interpolate(w, Qs)
print(wq.dat.data_ro)
```

The interpolation of x works, but the interpolation of the quadrature weight doesn't. It yields

```
Traceback (most recent call last):
  File "bug.py", line 16, in <module>
    wq = interpolate(w, Qs)
  File "/mnt/ntfs/data/install-scripts/firedrake/firedrake-dev-20200222-mpich/src/firedrake/firedrake/interpolation.py", line 36, in interpolate
    return Interpolator(expr, V, subset=subset, access=access).interpolate()
  File "/mnt/ntfs/data/install-scripts/firedrake/firedrake-dev-20200222-mpich/src/firedrake/firedrake/interpolation.py", line 58, in __init__
    self.callable, arguments = make_interpolator(expr, V, subset, access)
  File "/mnt/ntfs/data/install-scripts/firedrake/firedrake-dev-20200222-mpich/src/firedrake/firedrake/interpolation.py", line 163, in make_interpolator
    loops.extend(_interpolator(V, tensor, expr, subset, arguments, access))
  File "</mnt/ntfs/data/install-scripts/firedrake/firedrake-dev-20200222-mpich/lib/python3.6/site-packages/decorator.py:decorator-gen-22>", line 2, in _interpolator
  File "/mnt/ntfs/data/install-scripts/firedrake/firedrake-dev-20200222-mpich/src/firedrake/firedrake/utils.py", line 64, in wrapper
    return f(*args, **kwargs)
  File "/mnt/ntfs/data/install-scripts/firedrake/firedrake-dev-20200222-mpich/src/firedrake/firedrake/interpolation.py", line 205, in _interpolator
    ast, oriented, needs_cell_sizes, coefficients, _ = compile_expression_dual_evaluation(expr, to_element, coords, coffee=False)
[...]

PR #210, opened 2020-04-28
Fix tsfc bug #208.

pefarrell:
Enables `interpolate(QuadratureWeight(mesh), FunctionSpace(..., QuadratureElement(...)))`.

Depends on https://github.com/FEniCS/fiat/pull/43 .

### DIFF
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -366,6 +366,7 @@ def compile_expression_dual_evaluation(expression, to_element, coordinates, inte
         # Allow interpolation onto QuadratureElements to refer to the quadrature
         # rule they represent
         if isinstance(to_element, FIAT.QuadratureElement):
+            assert all(qpoints == to_element._points)
             quad_rule = QuadratureRule(point_set, to_element._weights)
             config["quadrature_rule"] = quad_rule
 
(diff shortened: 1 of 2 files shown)


tsfc/driver.py
```diff
@@ -361,6 +362,12 @@ def compile_expression_dual_evaluation(expression, to_element, coordinates, inte
         point_set = PointSet(qpoints)
         config = kernel_cfg.copy()
         config.update(point_set=point_set)
+
+        # Fix for interpolating QuadratureWeight into a QuadratureElement
```
wence-, line 366: I think this comment only makes sense now, and not later. Perhaps "Allow interpolation onto QuadratureElements to refer to the quadrature rule they represent"?

pefarrell: Fixed, thanks.

tsfc/driver.py
```diff
@@ -361,6 +362,13 @@ def compile_expression_dual_evaluation(expression, to_element, coordinates, inte
         point_set = PointSet(qpoints)
         config = kernel_cfg.copy()
         config.update(point_set=point_set)
+
+        # Allow interpolation onto QuadratureElements to refer to the quadrature
+        # rule they represent
+        if isinstance(to_element, FIAT.QuadratureElement):
+            quad_rule = QuadratureRule(point_set, to_element._weights)
```
wence-, line 369: I know it's trivially true, but can you `assert` that the `point_set` points we've collected (`qpoints`) match the `to_element._points`?

pefarrell: Done.

PR #211, opened 2020-04-30
travis: Test on 3.6--3.8

wence-:


miklos1:
Is support for Python 3.5 being dropped?

3.5 will certainly break if not tested, since `dict`s became "ordered" in 3.6...

wence-:
Yeah, Firedrake is about to be (or maybe accidentally is) >= 3.6

PR #212, opened 2020-05-01
wence/complex

wence-:
This is the result of the complex sprint, with commits cleaned up. Firedrake master passes with this branch (https://jenkins.ese.ic.ac.uk:1080/blue/organizations/jenkins/firedrake/detail/wence%2Ftest-complex-tsfc/2/pipeline)

It needs one mini-patch in Firedrake (to set the scalar_type in tsfc parameters correctly).

I think this is ready for review.

### DIFF
--- a/gem/refactorise.py
+++ b/gem/refactorise.py
@@ -8,7 +8,7 @@ from sys import intern
 
 from gem.node import Memoizer, traversal
 from gem.gem import (Node, Conditional, Zero, Product, Sum, Indexed,
-                     ListTensor, one)
+                     ListTensor, one, MathFunction)
 from gem.optimise import (remove_componenttensors, sum_factorise,
                           traverse_product, traverse_sum, unroll_indexsum,
                           make_rename_map, make_renamer)
@@ -169,7 +169,7 @@ def _collect_monomials(expression, self):
     sums = []
     for expr in compounds:
         summands = traverse_sum(expr, stop_at=stop_at)
-        if len(summands) <= 1 and not isinstance(expr, Conditional):
+        if len(summands) <= 1 and not isinstance(expr, (Conditional, MathFunction)):
             # Compound term is not an addition, avoid infinite
             # recursion and fail gracefully raising an exception.
             raise FactorisationError(expr)
@@ -209,6 +209,29 @@ def _collect_monomials(expression, self):
 
         result.add(sum_indices, atomics, rest)
     return result
+
+
+@_collect_monomials.register(MathFunction)
+def _collect_monomials_mathfunction(expression, self):
+    name = expression.name
+    if name in {"conj", "real", "imag"}:
+        # These are allowed to be applied to arguments, and hence must
+        # be dealt with specially. Just push the function onto each
+        # entry in the monomialsum of the child.
+        # NOTE: This presently assumes that the "atomics" part of a
+        # MonomialSum are real. This is true for the coffee, tensor,
+        # spectral modes: the atomics are indexed tabulation matrices
+        # (which are guaranteed real).
+        # If the classifier puts (potentially) complex expressions in
+        # atomics, then this code needs fixed.
+        child_ms, = map(self, expression.children)
+        result = MonomialSum()
+        for k, v in child_ms.monomials.items():
+            result.monomials[k] = MathFunction(name, v)
+        result.ordering = child_ms.ordering.copy()
+        return result
+    else:
+        return _collect_monomials.dispatch(MathFunction.mro()[1])(expression, self)
 
 
 @_collect_monomials.register(Conditional)
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -98,6 +98,10 @@ def compile_integral(integral_data, form_data, prefix, parameters, interface, co
             # Delayed import, loopy is a runtime dependency
             import tsfc.kernel_interface.firedrake_loopy as firedrake_interface_loopy
             interface = firedrake_interface_loopy.KernelBuilder
+    if coffee:
+        scalar_type = parameters["scalar_type_c"]
+    else:
+        scalar_type = parameters["scalar_type"]
 
     # Remove these here, they're handled below.
     if parameters.get("quadrature_degree") in ["auto", "default", None, -1, "-1"]:
@@ -123,7 +127,7 @@ def compile_integral(integral_data, form_data, prefix, parameters, interface, co
     domain_numbering = form_data.original_form.domain_numbering()
     builder = interface(integral_type, integral_data.subdomain_id,
                         domain_numbering[integral_data.domain],
-                        parameters["scalar_type"],
+                        scalar_type,
                         diagonal=diagonal)
     argument_multiindices = tuple(builder.create_element(arg.ufl_element()).get_indices()
                                   for arg in arguments)
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -50,14 +50,7 @@ def compile_form(form, prefix="form", parameters=None, interface=None, coffee=Tr
     assert isinstance(form, Form)
 
     # Determine whether in complex mode:
-    # complex nodes would break the refactoriser.
     complex_mode = parameters and is_complex(parameters.get("scalar_type"))
-    if complex_mode:
-        logger.warning("Disabling whole expression optimisations"
-                       " in GEM for supporting complex mode.")
-        parameters = parameters.copy()
-        parameters["mode"] = 'vanilla'
-
     fd = ufl_utils.compute_form_data(form, complex_mode=complex_mode)
     logger.info(GREEN % "compute_form_data finished in %g seconds.", time.time() - cpu_time)
 
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -155,7 +155,8 @@ def compile_integral(integral_data, form_data, prefix, parameters, interface, co
                       integration_dim=integration_dim,
                       entity_ids=entity_ids,
                       argument_multiindices=argument_multiindices,
-                      index_cache=index_cache)
+                      index_cache=index_cache,
+                      complex_mode=is_complex(parameters.get("scalar_type")))
 
     mode_irs = collections.OrderedDict()
     for integral in integral_data.integrals:
@@ -342,7 +343,8 @@ def compile_expression_dual_evaluation(expression, to_element, coordinates, inte
                       ufl_cell=coordinates.ufl_domain().ufl_cell(),
                       precision=parameters["precision"],
                       argument_multiindices=argument_multiindices,
-                      index_cache={})
+                      index_cache={},
+                      complex_mode=complex_mode)
 
     if all(isinstance(dual, PointEvaluation) for dual in to_element.dual_basis()):
         # This is an optimisation for point-evaluation nodes which
--- a/tsfc/loopy.py
+++ b/tsfc/loopy.py
@@ -2,8 +2,6 @@
 
 This is the final stage of code generation in TSFC."""
 
-from math import isnan
-
 import numpy
 from functools import singledispatch
 from collections import defaultdict, OrderedDict
@@ -399,10 +397,10 @@ def _expression_conditional(expr, ctx):
 def _expression_scalar(expr, parameters):
     assert not expr.shape
     v = expr.value
-    if isnan(v):
+    if numpy.isnan(v):
         return p.Variable("NAN")
-    r = round(v, 1)
-    if r and abs(v - r) < parameters.epsilon:
+    r = numpy.round(v, 1)
+    if r and numpy.abs(v - r) < parameters.epsilon:
         return r
     return v
 
--- a/tsfc/loopy.py
+++ b/tsfc/loopy.py
@@ -3,10 +3,11 @@
 This is the final stage of code generation in TSFC."""
 
 import numpy
-from functools import singledispatch
+from functools import singledispatch, partial
 from collections import defaultdict, OrderedDict
 
 from gem import gem, impero as imp
+from gem.node import Memoizer
 
 import islpy as isl
 import loopy as lp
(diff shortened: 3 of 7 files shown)


gem/impero_utils.py
```diff
@@ -316,3 +316,84 @@ def recurse_top(o):
         raise AssertionError("unhandled operation: %s" % type(op))
 
     return counter
+
+
+def dtype_sort_key(dtype):
+    return 1 if dtype.kind == 'c' else -1
```
wence-, line 322: Should use `numpy.find_common_dtype`

gem/refactorise.py
```diff
@@ -169,7 +169,7 @@ def stop_at(expr):
     sums = []
     for expr in compounds:
         summands = traverse_sum(expr, stop_at=stop_at)
-        if len(summands) <= 1 and not isinstance(expr, Conditional):
+        if len(summands) <= 1 and not isinstance(expr, (Conditional, MathFunction)):
```
wence-, line 172: MathFunction has a special handler (to deal with conj/real/imag).

gem/refactorise.py
```diff
@@ -211,6 +211,23 @@ def stop_at(expr):
...
+        # be dealt with specially. Just push the function onto each
+        # entry in the monomialsum of the child.
+        child_ms, = map(self, expression.children)
+        result = MonomialSum()
+        for k, v in child_ms.monomials.items():
+            result.monomials[k] = MathFunction(name, v)
+        result.ordering = child_ms.ordering.copy()
+        return result
+    else:
+        return _collect_monomials.dispatch(MathFunction.mro()[1])(expression, self)
```
wence-, line 228: singledispatch version of `super()`.

tsfc/driver.py
```diff
@@ -50,14 +50,7 @@ def compile_form(form, prefix="form", parameters=None, interface=None, coffee=Tr
     assert isinstance(form, Form)
 
     # Determine whether in complex mode:
-    # complex nodes would break the refactoriser.
     complex_mode = parameters and is_complex(parameters.get("scalar_type"))
-    if complex_mode:
-        logger.warning("Disabling whole expression optimisations"
-                       " in GEM for supporting complex mode.")
-        parameters = parameters.copy()
-        parameters["mode"] = 'vanilla'
```
wence-, line 59: Refactoriser now handles complex nodes.

tsfc/loopy.py
```diff
@@ -81,13 +110,13 @@ def generate(impero_c, args, precision, scalar_type, kernel_name="loopy_kernel",
 
     # Create arguments
     data = list(args)
-    for i, temp in enumerate(impero_c.temporaries):
+    for i, (temp, dtype) in enumerate(assign_dtypes(impero_c.temporaries, scalar_type)):
```
wence-, line 113: Type propagation.

tsfc/loopy.py
```diff
@@ -241,11 +270,6 @@ def _expression_power(expr, ctx):
 @_expression.register(gem.MathFunction)
 def _expression_mathfunction(expr, ctx):
 
-    from tsfc.coffee import math_table
-
-    math_table = math_table.copy()
-    math_table['abs'] = ('abs', 'cabs')
```
wence-, line 247: Loopy mode now has a separate math table.

tsfc/loopy.py
```diff
@@ -329,10 +354,10 @@ def _expression_conditional(expr, ctx):
 def _expression_scalar(expr, parameters):
     assert not expr.shape
     v = expr.value
-    if isnan(v):
+    if numpy.isnan(v):
         return p.Variable("NAN")
-    r = round(v, 1)
-    if r and abs(v - r) < parameters.epsilon:
+    r = numpy.round(v, 1)
+    if r and numpy.abs(v - r) < parameters.epsilon:
```
wence-, line 360: `math.isnan/round` don't work with numpy types.

tsfc/parameters.py
```diff
@@ -14,8 +14,11 @@
     # that makes compilation time much shorter.
     "unroll_indexsum": 3,
 
-    # Scalar type (C typename string)
-    "scalar_type": "double",
+    # Scalar type numpy dtype
+    "scalar_type": numpy.dtype(numpy.float64),
+
+    # So that tests pass (needs to match scalar_type)
+    "scalar_type_c": "double",
```
wence-, line 21: This is a really dumb way to do this (scalar_type_c can be expunged when coffee backend goes).

gem/impero_utils.py
```diff
@@ -316,3 +317,80 @@ def recurse_top(o):
...
+def _assign_dtype_conditional(expression, self):
+    return maxtype(map(self, expression.children[1:]))
+
+
+@_assign_dtype.register(gem.Comparison)
+@_assign_dtype.register(gem.LogicalNot)
+@_assign_dtype.register(gem.LogicalAnd)
+@_assign_dtype.register(gem.LogicalOr)
+def _assign_dtype_logical(expression, self):
+    return numpy.bool
```
wence-, line 378: Not sure these will ever cons temporaries, but we may as well handle them.

wence-: turns out they do!

gem/impero_utils.py
```diff
@@ -316,3 +317,80 @@ def recurse_top(o):
...
+    # Conservative
+    return self.scalar_type
+
+
+@_assign_dtype.register(gem.MathFunction)
+def _assign_dtype_mathfunction(expression, self):
+    if expression.name in {"real", "imag"}:
+        return self.real_type
+    else:
+        return maxtype(map(self, expression.children))
```
wence-, line 358: Is this the safe thing to do? I don't think we support any mathfunctions that complexify real values in complex mode right now.

wence-:
And now with passing firedrake again (real mode): https://jenkins.ese.ic.ac.uk:1080/blue/organizations/jenkins/firedrake/detail/wence%2Ftest-complex-tsfc/4/pipeline/67

wence-:
Reminder before merging to fix the `simplify_abs` visitor:
```python
@_simplify_abs.register(Sqrt)
def _simplify_abs_sqrt(o, self, in_abs):
    # Square root is always non-negative
    return ufl_reuse_if_untouched(o, self(o.ufl_operands[0], False))
```

gem/impero_utils.py
miklos1: What about square root of negatives?

gem/impero_utils.py
```diff
@@ -316,3 +317,80 @@ def recurse_top(o):
...
+def _assign_dtype_conditional(expression, self):
+    return maxtype(map(self, expression.children[1:]))
+
+
+@_assign_dtype.register(gem.Comparison)
+@_assign_dtype.register(gem.LogicalNot)
+@_assign_dtype.register(gem.LogicalAnd)
+@_assign_dtype.register(gem.LogicalOr)
+def _assign_dtype_logical(expression, self):
+    return numpy.int8
```
miklos1, line 378: Why not `numpy.bool_`?

gem/impero_utils.py
```diff
@@ -316,3 +317,80 @@ def recurse_top(o):
...
+    :arg scalar_type: Default scalar type.
+
+    :returns: list of tuples (expression, dtype)."""
+    mapper = Memoizer(_assign_dtype)
+    mapper.scalar_type = scalar_type
+    if scalar_type.kind == "c":
+        mapper.real_type = scalar_type.type(0).real.dtype
+    else:
+        mapper.real_type = scalar_type
+    return [(e, mapper(e)) for e in expressions]
```
miklos1, line 396: This only needs GEM and numpy, but not the Impero AST.  Nevertheless, since it's so closely coupled to code generation, I think it's fine that this visitor stays here.

gem/refactorise.py
```diff
@@ -211,6 +211,23 @@ def stop_at(expr):
...
+def _collect_monomials_mathfunction(expression, self):
+    name = expression.name
+    if name in {"conj", "real", "imag"}:
+        # These are allowed to be applied to arguments, and hence must
+        # be dealt with specially. Just push the function onto each
+        # entry in the monomialsum of the child.
+        child_ms, = map(self, expression.children)
+        result = MonomialSum()
+        for k, v in child_ms.monomials.items():
+            result.monomials[k] = MathFunction(name, v)
```
miklos1, line 224: So you only propagate these complex functions to the "rest" factors, which means you assume that "atomics" are real.  This is true for argument factorisation (which is the main use of the refactoriser), but I'm not sure that this would work correctly with tensor mode (which uses the refactoriser for a different refactorisation).

I think this special assumption should at least be noted in a comment.

gem/impero_utils.py
wence-: The dtype mapping in loopy generates `bool` for `numpy.bool_` which is not a C type.

gem/refactorise.py
wence-: `atomics` in this case (arguments) are the basis tabulations, yes? I guess the "right thing" would be to check what atomics are and propagate to both (if necessary). Or one could have a flag in the refactoriser that says "are atomics real?". WDYT?

gem/impero_utils.py
wence-: Oh yeah. Probably what I should do is handle the entire "math-table" one by one (which suggests migrating this mapper into the loopy codegen part [this is OK, since I didn't fix up the coffee backend]).

miklos1:
Lawrence Mitchell <notifications@github.com> writes:
> @wence- commented on this pull request.
> +@_collect_monomials.register(MathFunction)
> +def _collect_monomials_mathfunction(expression, self):
> +    name = expression.name
> +    if name in {"conj", "real", "imag"}:
> +        # These are allowed to be applied to arguments, and hence must
> +        # be dealt with specially. Just push the function onto each
> +        # entry in the monomialsum of the child.
> +        child_ms, = map(self, expression.children)
> +        result = MonomialSum()
> +        for k, v in child_ms.monomials.items():
> +            result.monomials[k] = MathFunction(name, v)
>
> `atomics` in this case (arguments) are the basis tabulations, yes? I
> guess the "right thing" would be to check what atomics are and
> propagate to both (if necessary). Or one could have a flag in the
> refactoriser that says "are atomics real?". WDYT?

"Atomics" are defined by the classifier function, and the classifier
functions are defined by the optimisation modes.

I just looked up the source code, and the real assumption is actually
fine for all our optimisation modes.  Vanilla does not call
refactoriser, tensor defines atomics as tabulation matrices with a
quadrature index, coffee and spectral modes define atomics as tabulation
tensors with argument indices or delta nodes with argument indices.

I guess adding a comment about the real assumption suffices for now.

gem/impero_utils.py
wence-: I fixed this in inducer/compyte#30, but it is not yet backmerged into the firedrake loopy branch.

gem/impero_utils.py
wence-: I moved it close to the mathtable in loopy.py now.

gem/refactorise.py
wence-: Added a comment.

gem/impero_utils.py
wence-: Done.

wence-:
> Reminder before merging to fix the simplify_abs visitor:

Now done.

tsfc/loopy.py
```diff
@@ -23,6 +22,116 @@
...
+    'ln': ('log', 'log'),
+    'real': (None, 'real'),
+    'imag': (None, 'imag'),
+    'conj': (None, 'conj'),
+    'erf': ('erf', None),
+    'atan_2': ('atan2', None),
+    'atan2': ('atan2', None),
+    'min_value': ('min', None),
+    'max_value': ('max', None)
+}
```
miklos1, line 53: This table now feels almost unnecessary.

- The real and complex function names are always the same for those functions which have both versions.
- ... and this target function name is in most cases the same as the source function name, except for `power`/`pow`, `ln`/`log`, `atan_2`/`atan2`, `min_value`/`min`, and `max_value`/`max`.
- Why are there two source names for atan2 by the way?  That looks like an inconsistency somewhere else in the Firedrake stack.
- Power, min value, and max value have their own GEM nodes anyway, so they need not make use of this table.
- That leaves `ln`/`log` alone where a name translation is necessary, plus the dubious atan2 case.

wence-: I checked `atan_2` is just a mistake. I refactored to remove this table. That also fixes a bug where one couldn't use `erf` (or similar) in complex mode even if the argument was wrapped in `real` to ensure realness.

wence-:
I think I've now addressed all the comments. I made enough changes that I need to run through with the real-mode version of firedrake to check, but I hope that this now at least passes tsfc tests. With the appropriate loopy branch it passes tsfc tests when the scalar type is complex too.

tsfc/coffee.py
```diff
@@ -213,7 +213,7 @@ def _expression_division(expr, parameters):
 # Copied from FFCX (ffc/language/ufl_to_cnodes.py)
 math_table = {
     'sqrt': ('sqrt', 'csqrt'),
-    'abs': ('fabs', 'cabs'),
+    'abs': ('fabs', 'abs'),
```
miklos1, line 216: COFFEE mode probably not very relevant anymore, but this change does not look right.

wence-: I cherry-picked from the sprint branch (particularly ff5193e66408d) perhaps @sv2518 can comment on why this change was needed?

wence-: Oh, I guess it's because the slate-mode kernels were being compiled in C++ mode, for which `abs` is correct. But we decided not to port slate until it is migrated to the loopy codegen path anyway, so this change is moot.

miklos1: Oh yes, complex is quite different in C and in C++...

miklos1:
So I think this is fine to be merged, though I note that the code seems to be growing in a direction where one should consider introducing GEM nodes `Real`, `Imag`, and `Conj`.  There are some 3 cases when one checks whether the name of a `MathFunction` is `in {'real', 'imag', 'conj'}`.  This change would probably increase number of code lines at this point, but it seems to be the better code organisation if this trend continues.

tsfc/coffee.py
wence-: Dropped this part.

wence-:
The abs-simplification change was incomplete and I think I've now fixed it. I'm a bit unhappy that it involves pushing this `complex_mode` flag through so many places though.

PR #214, opened 2020-05-11
Allow for ComponentTensor(NotAFlexiblyIndexed())

sv2518:
… in decompose_variable_view, so that the function can be reused for generating Slate blocks.

### DIFF
--- a/gem/gem.py
+++ b/gem/gem.py
@@ -874,10 +874,16 @@ def strides_of(shape):
 
 
 def decompose_variable_view(expression):
-    """Extract ComponentTensor + FlexiblyIndexed view onto a variable."""
-    if isinstance(expression, Variable):
+    """Extract information from a shaped node.
+       Decompose ComponentTensor + FlexiblyIndexed."""
+    if (isinstance(expression, (Variable, Inverse, Solve))):
         variable = expression
         indexes = tuple(Index(extent=extent) for extent in expression.shape)
+        dim2idxs = tuple((0, ((index, 1),)) for index in indexes)
+    elif (isinstance(expression, ComponentTensor) and
+          not isinstance(expression.children[0], FlexiblyIndexed)):
+        variable = expression
+        indexes = expression.multiindex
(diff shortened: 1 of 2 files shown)


gem/gem.py
```diff
@@ -876,7 +876,9 @@ def strides_of(shape):
 
 def decompose_variable_view(expression):
     """Extract ComponentTensor + FlexiblyIndexed view onto a variable."""
-    if isinstance(expression, Variable):
+    if (isinstance(expression, Variable) or
+        (isinstance(expression, ComponentTensor) and
+         not isinstance(expression.children[0], FlexiblyIndexed))):
```
wence-, line 881: Update docstring? This is basically treating shaped things that are not `CT(FlexiblyIndexed)` as a `Variable`? (so no decomposition)?

gem/gem.py
```diff
@@ -875,11 +875,19 @@ def strides_of(shape):
 
 
 def decompose_variable_view(expression):
-    """Extract ComponentTensor + FlexiblyIndexed view onto a variable."""
-    if isinstance(expression, Variable):
+    """Extract information from a shaped node.
+       Decompose ComponentTensor + FlexiblyIndexed."""
+    if (isinstance(expression, Variable) or
+       isinstance(expression, Inverse) or
+       isinstance(expression, Solve)):
```
wence-, line 882: `isinstance` accepts a tuple of types to check. So `isinstance(expression, (Inverse, Solve, Variable))` is probably clearer.

tsfc/loopy.py
```diff
@@ -243,6 +237,17 @@ def generate(impero_c, args, precision, scalar_type, kernel_name="loopy_kernel",
     return knl
 
 
+def create_domains(indices):
```
wence-, line 240: Docstring, now this is being exported?

Perhaps

```
Create ISL domains from indices

:arg indices: iterable of (index_name, extent) pairs
:returns: A list of ISL sets representing the iteration domain of the indices.
```

WDYT?

wence-:
Thanks.

PR #216, opened 2020-05-17
wence/feature/return accumulate

wence-:
Add ability to suppress emission of ReturnAccumulate in scheduling. This is necessary for TSFC kernels where the output tensor is not zeroed on input.

Similarly, don't do `lvalue = lvalue + rvalue` when emitting `Return` nodes in loopy output (the accumulation should have happened in the `Accumulate` part of the pair, and again this presumes that the lvalue was zero on entry.

### DIFF
--- a/gem/impero_utils.py
+++ b/gem/impero_utils.py
@@ -38,12 +38,17 @@ def preprocess_gem(expressions, replace_delta=True, remove_componenttensors=True
     return expressions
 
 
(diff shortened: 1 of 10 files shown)


wence-:
Turns out it wasn't fine (because when assembling zero-forms pyop2 doesn't send in a temporary but just the pointer to the data). So now this `Return` dichotomy is preserved optionally.

I also refactored the setting of precision/epsilon everywhere (inheriting it from numpy.finfo).

PR #218, opened 2020-06-18
Dual Evaluation with FInAT

mkan88:
Performing dual evaluation in FInAT has been a long awaited feature in firedrake. This allows interaction with dual_evaluation methods in FInAT. To be used with corresponding branches in [FInAT #57](https://github.com/FInAT/FInAT/pull/57) and [Firedrake #1743](https://github.com/firedrakeproject/firedrake/pull/1743).


tsfc/driver.py
```diff
@@ -359,19 +359,20 @@ def fn(point_set, deriv=None):
...
-            from ufl.operators import Dx, Dn, div, grad, nabla_div, nabla_grad, curl, \
-                                      rot, diff
-            deriv_dict = {'Dx': Dx, 'Dn': Dn, 'div': div, 'grad': grad, 'nabla_div': nabla_div, \
-                          'nabla_grad': nabla_grad, 'curl': curl, 'rot': rot, 'diff': diff}
+            from ufl.operators import Dn, div, grad, nabla_div, nabla_grad, curl, rot
+            deriv_oper = {'Dn': Dn, 'div': div, 'grad': grad, 'nabla_div': nabla_div,
+                          'nabla_grad': nabla_grad, 'curl': curl, 'rot': rot}
             config = kernel_cfg.copy()
             config.update(point_set=point_set)
             if deriv is None:
```
dham, line 367: If  you make the default value of `deriv` 0 instead of `None` then we can get rid of this whole if statement and just keep lines 372 and 373. Also, abbreviations are the enemy, just call it `derivative`.

tsfc/driver.py
```diff
@@ -359,19 +359,20 @@ def fn(point_set, deriv=None):
...
-            elif isinstance(deriv, (tuple, list)):
-                # TODO: How to do Dx and diff?
+            elif type(deriv) is int:
+                # Dx requires GEM indexing in FInAT
                 dexpression = expression
+                for _ in range(deriv):
+                    dexpression = grad(dexpression)
             else:
-                dexpression = deriv_dict[deriv](expression)
+                dexpression = deriv_oper[deriv](expression)
```
wence-, line 375: I _think_ that the interface for specifying the derivatives doesn't need to be this complicated. FInAT is going to be requesting derivatives based on information in the functionals, and they will always want (components) of the gradient (or higher derivatives) of the input function. So I think you just need to be able to specify how many grads to apply to each component of the `ufl_shape` of the input expression and separately which components of the resulting expression to select (or does that selection happen on the "other side")?

tsfc/driver.py
mkan88: Wouldn't just doing `grad` require more processing on the FInAT end, if say one were to use a finite element with curls or divs? Though it could be more useful if one wants to use another form language?

tsfc/driver.py
mkan88: Same as above.

tsfc/driver.py
dham: I'm not aware of a finite element whose nodes are divergences (although it would be possible to do anyway - div is just sum grad). Nodes are by definition scalar valued, so I don't think curl could come up. I think this is a kind of theoretical case...

wence-: I think there's a slight misunderstanding in what the "derivative" means here. FIAT functionals are implemented as quadrature rules that specify for each point in the quadrature rule what the weight is. Then we ask what are we integrating, and the answer is that we integrate either (components of) an expression or (components of) derivatives of an expression. The way we say what the derivatives are is that we specify (for each spatial component of the expression) what the order of the derivative is. So I think we always just need to take `grad` (or perhaps `expr.dx(...)`).

Does that make sense?

mkan88: I think its a bit clearer now. By integrating, do you simply mean the sum of the quadrature weights multiplied by the values of the nodes fed the given expression?

wence-: Yes. See the description of FIAT functionals here: https://github.com/FEniCS/fiat/blob/master/FIAT/functional.py#L39

A functional (a node) is a rule `f[expr]` that eats an expression and produces a number. They're almost all specified as an integral with some selected measure. For example, all of the point evaluation functionals are
```
f_p[expr] = \int_K \delta(x - p) expr dx = expr(p)
```
(for evaluation at the point `p`).

Many of the other functionals are moments against components of the expression against selected sets of polynomials.

So what FIAT does is encode this integration rule in a set of weights and points (via numerical quadrature) and so we have

```
f[expr] = \sum_q w_q expr(\phi_q)
```

With some potentially fancier stuff happening if you only care about a single component of the expression.

dham: ... except that delta is a distribution not a function and doesn't go inside an integral, but the core idea is right.

Since the space we are interpolating into is finite dimensional, all nodes are equivalent to integrals so we could theoretically do away with derivatives at all. However, as @rckirby pointed out, that looks a bit messy so explicitly using derivatives of the expression is probably a better move.

rckirby:
We are currently working (on a branch) on some Arnold-Winther elements.
The functionals defining constraints in this case would involve certain
moments of divergences.  So yes, they can appear ;)

tsfc/driver.py
```diff
@@ -284,8 +290,19 @@ def compile_expression_dual_evaluation(expression, to_element, coordinates, *,
...
     import loopy as lp
-    if any(len(dual.deriv_dict) != 0 for dual in to_element.dual_basis()):
-        raise NotImplementedError("Can only interpolate onto dual basis functionals without derivative evaluation, sorry!")
+
+    # TODO: Temporary fix to maintain old code, fix mapping name
+    if isinstance(element, FiniteElementBase):
+        # FInAT mapping outputs string, FIAT mapping outputs tuple
+        temp_mapping = (element.mapping,)
+        if isinstance(element, FiatElement):
+            to_element = element._element
```
ReubenHill, line 299: Note that you can end up not actually setting `to_element` here. I have tried to run this with a FInAT `QuadratureElement` and can't get past line 417.

mkan88: Where are you calling `compile_expression_dual_evaluation` from? If `interpolate` is being used it should be passed a FIAT element instead of a FInAT element for `QuadratureElement`, though I agree there should be better checking (and more consistent class types) here.

ReubenHill: From within `interpolate` - I will add the required `dual_evaluation` and `dual_basis` methods

mkan88: Only `dual_basis` should be required unless it requires some wonky delta cancellations like `TensorFiniteElement`

miklos1:
I wanted to make this proposal / suggestion much earlier, but it is still not late, since the comments I'm about to make concern the high-level code organisation, while progress seems to have been made on the individual elements in the past month.

The ugliness I am concerned with are the repeated switch-case anti-patterns, that
- Firedrake gives maybe a FIAT element, maybe a FInAT element to evaluate on
- TSFC then receives either a FIAT or a FInAT element to evaluate on
- Both Firedrake and TSFC contain ugly element lists to check whether there is special handling in FInAT.  (Why? It should be enough for FInAT to know, whether it has special handling or not...)

I suggest the following refactorings / reorganisations:
1. *Scatter the FIAT interface onto FInAT elements.*  Basically, kill `fiatinterface.py` as it exists now, and teach instead FInAT elements to construct the equivalent FIAT elements.  (This is a perfect substitute, since you will still be able to do the UFL -> FIAT element conversion with UFL -> FInAT from the `finatinterface.py`, and FInAT -> FIAT by asking the FInAT element.  But you shall not need to do that directly.)
2. *The interpolation code in Firedrake shall construct a FInAT element that is passed to TSFC's `compile_expression_dual_evaluation`.  Full stop.*  How TSFC and further FInAT deal with that FInAT element is none of Firedrake's business.
3. *In `compile_expression_dual_evaluation`, separate the parts concerning the evaluations on UFL expressions and concerning what evaluations to make for the element and how to construct the basis coefficient values.*  (I'm saying this with respect to master, this is more or less already done in this branch.)  The latter codes shall move to FInAT, the former remain in TSFC and provided to FInAT through the callback interface.  TSFC should not be concerned about the exact type of the FInAT element whose dual basis being evaluated.
4. The FInAT element base class should have the following dual evaluation code:
    - Construct the corresponding FIAT element
    - Do exactly what TSFC currently (= on master) does to evaluate on FIAT elements.  That is, move that code (mentioned in point 3) from TSFC to here.

At this point, one would have no-regression refactoring that is already a reviewable and mergeable chunk.  After this, the actual improvements on dual evaluation could be done by overriding the dual evaluation method on each element.  (In the absence of such override, control automatically falls back onto the base class which does evaluation through FIAT.)  Then later as more and more elements learn how to evaluate their dual, the need for FInAT elements knowing how to construct their corresponding FIAT element (point 1) may disappear.

tsfc/driver.py
```diff
@@ -346,7 +365,59 @@ def compile_expression_dual_evaluation(expression, to_element, coordinates, *,
                       index_cache={},
                       scalar_type=parameters["scalar_type"])
 
-    if all(isinstance(dual, PointEvaluation) for dual in to_element.dual_basis()):
+    elements_with_basis = (FiatElement, TensorFiniteElement, TensorProductElement,
+                           EnrichedElement, FlattenedDimensions, DiscontinuousElement)
+    if isinstance(element, elements_with_basis):
```
ReubenHill, line 370: Could this be generalised without causing issues to
```
    try:
        if isinstance(element.dual_basis, tuple):
          has_dual_basis = True
    except (AttributeError, NotImplementedError):
        has_dual_basis = False
    if has_dual_basis:
```
or is there a good reason for this to be maintained as a list?

wence-: These isinstance checks are an anti-pattern. I agree with Miklos' refactoring suggestion above.

mkan88:
I agree with most of Miklos' points above. However, after creating the corresponding FIAT element, I think one could create a FInAT `FiatElement` from the FIAT element and simply use its interface to perform a generic dual evaluation.

miklos1:
> However, after creating the corresponding FIAT element, I think one could create a FInAT FiatElement from the FIAT element and simply use its interface to perform a generic dual evaluation.

Where would this be?  Somewhere in TSFC or somewhere in FInAT?

miklos1:
One should weigh whether to do the refactoring above on top of this branch (moving and adjusting all new code along) or on top of master (transplanting changes as needed at first, then rebasing this branch on the refactoring). Though even in the former case, I suggest splitting the pull request, so that the refactorings could be merged before all the new functionality is worked out.

However, in any case, one should avoid endlessly rolling a big ball of changes that is never ready to be merged.  (This isn't an instance of that yet, so I am warning in advance for the sake of good software engineering.)  When this happens, the big ball of changes typically contains some stable parts which could be merged *if isolated*, as well as some more work-in-progress parts which are blocking a wholesale merge.  In such cases, one should not be afraid to slice up the PR, finalising the well thought out parts to be merged, and then rolling only a small ball of changes onwards.

mkan88:
> > However, after creating the corresponding FIAT element, I think one could create a FInAT FiatElement from the FIAT element and simply use its interface to perform a generic dual evaluation.
> 
> Where would this be? Somewhere in TSFC or somewhere in FInAT?

It could be done entirely in FInAT by passing the `dual_basis` of the FInAT `FiatElement` to perform `dual_evaluation`.

tsfc/fiatinterface.py
```diff
@@ -0,0 +1,259 @@
...
+        return cache[vector_is_mixed]
+    except KeyError:
+        pass
+
+    if element.cell() is None:
+        raise ValueError("Don't know how to build element when cell is not given")
+
+    fiat_element = convert(element, vector_is_mixed)
+    cache[vector_is_mixed] = fiat_element
+    return fiat_element
```
miklos1, line 259: No! please don't resurrect this file.

mkan88: Removed.

tsfc/driver.py
```diff
@@ -434,6 +370,55 @@ def compile_expression_dual_evaluation(expression, to_element, coordinates, *,
...
+        self.complex_mode = complex_mode
+
+    def __call__(self, point_set, derivative=0):
+        '''Wrapper function for converting UFL `expression` into GEM expression.
+
+        :param point_set: FInAT PointSet
+        :param derivative: Maximum order of differentiation
+        '''
+        from ufl.differentiation import ReferenceGrad
+        # from ufl.operators import grad
```
ReubenHill, line 400: remove unused import

tsfc/driver.py
```diff
@@ -434,6 +370,55 @@ def compile_expression_dual_evaluation(expression, to_element, coordinates, *,
...
+        self.expression = expression
+
+        # Geometric dimension or topological dimension
+        # For differentiating UFL Zero
+        # Fix for Zero expressions with no dimension
+        # TODO: Fix for topolgical dimension different from geometric
+        try:
+            self.dimension = expression.geometric_dimension()
+        except UFLException:
+            self.dimension = 0
```
ReubenHill, line 385: So this is only for when getting the derivative of a `ufl.ConstantValue` which should be a `ufl.constantvalue.Zero` of the right shape. Since you appear to be copying [these lines](https://github.com/FEniCS/ufl/blob/81d1217792e00a4af5014a2d458da94077773883/ufl/differentiation.py#L179-L187) from `ufl.differentiation.ReferenceGrad` shouldn't this query `expression.topological_dimension`? 

Furthermore, doesn't this fix, if required, belong in a UFL PR?

ReubenHill (commented):
I don't really know enough about the design intention of `driver.py` to commend on the changes there (and conclude whether this achieves it's intended goals) but have left what I hope are useful comments for discussion. 

If someone could help me better understand what's going on in this module I'd be keen to understand :)

tsfc/driver.py
```diff
@@ -351,66 +341,12 @@ def compile_expression_dual_evaluation(expression, to_element, coordinates, *,
...
-            assert qexprs.shape == ()
-            assert set(qexprs.free_indices) == set(chain(shape_indices, *argument_multiindices))
-            dual_expressions.append(qexprs)
-        basis_indices = (gem.Index(), )
-        ir = gem.Indexed(gem.ListTensor(dual_expressions), basis_indices)
+    # Call dual_evaluation method in FInAT
+    ir_shape = element.dual_evaluation(UFLtoGEMCallback(expression, kernel_cfg, complex_mode))
+    broadcast_shape = len(expression.ufl_shape) - len(element.value_shape)
+    shape_indices = tuple(gem.Index() for _ in expression.ufl_shape[:broadcast_shape])
+    basis_indices = (gem.Index(), )
```
wence-, line 348: I _think_ that the `shape_indices` part (which was a hack when finat didn't provide the dual) should go away and we should `create_element` not `create_base_element` in Firedrake when calling this function.

tsfc/driver.py
```diff
@@ -434,6 +370,55 @@ def compile_expression_dual_evaluation(expression, to_element, coordinates, *,
     return builder.construct_kernel(return_arg, impero_c, index_names)
 
 
+class UFLtoGEMCallback(object):
+    def __init__(self, expression, kernel_cfg, complex_mode):
+        # UFL expression to turn into GEM
```
wence-, line 375: Docstring.

tsfc/driver.py
mkan88: Using `expression.ufl_domain().topological_dimension()` is indeed better, though it also fails for `test_constant` in `test_interpolate.py` since the `Constant` doesn't have a UFL domain or specified dimension. Any ways around it?

Doing this in UFL is probably the better solution, but @dham and I decided not to bother with doing a UFL PR due to time constraints earlier. Is now the time for it? With @rckirby pointwise dual branch on FIAT which can express dual vectors with derivatives as weighted sums of point evaluations only, `ReferenceGrad` might not even be needed.

tsfc/driver.py
mkan88: Doesn't Firedrake already use `create_element`? I think only `compile_python_kernel` uses `create_base_element`.

I'm not too familiar with the kernel. What does the kernel do with `shape_indices` and `basis_indices`?

miklos1:
UFLtoGEMCallback does not sound like a good name.  Two things are to be
distinguished when naming things well: *what* a thing does, and *how*
thing does that.  When being asked the former question, one easily slips
into answering the latter; so it is good keep in mind that "what does it
do?" is to be understood in the sense of "what function does it serve?"
or "what purpose does it fulfil?".  These questions relate to each other
like interface and implementation, and names should rather follow the
interface rather than the implementation.

UFLtoGEMCallback seems to tell too much about the mechanics of what's
inside, rather than the purpose it serves.

mkan88:
> UFLtoGEMCallback does not sound like a good name. Two things are to be distinguished when naming things well: *what* a thing does, and *how* thing does that. When being asked the former question, one easily slips into answering the latter; so it is good keep in mind that "what does it do?" is to be understood in the sense of "what function does it serve?" or "what purpose does it fulfil?". These questions relate to each other like interface and implementation, and names should rather follow the interface rather than the implementation. UFLtoGEMCallback seems to tell too much about the mechanics of what's inside, rather than the purpose it serves. Lawrence Mitchell <notifications@github.com> writes:
> […](#)
> @wence- commented on this pull request. > @@ -434,6 +370,55 @@ def compile_expression_dual_evaluation(expression, to_element, coordinates, *, return builder.construct_kernel(return_arg, impero_c, index_names) +class UFLtoGEMCallback(object): + def __init__(self, expression, kernel_cfg, complex_mode): + # UFL expression to turn into GEM Docstring. -- You are receiving this because you commented. Reply to this email directly or view it on GitHub: [#218 (review)](https://github.com/firedrakeproject/tsfc/pull/218#pullrequestreview-493662747)

Maybe something like `GEMFunctionCallback`?

tsfc/driver.py
ReubenHill: See [this firedrake commit](https://github.com/firedrakeproject/firedrake/pull/1743/commits/1b86dbfd039cb7bd11fcf7e4ad6a1fea7a8b0f73).

ReubenHill:
Superseded by #250

PR #220, opened 2020-07-10
Do not erroneously cast to real in GEM interpreter

miklos1:
This bug triggered warning when FInAT/FInAT#58.

### DIFF
--- a/gem/interpreter.py
+++ b/gem/interpreter.py
@@ -320,11 +320,7 @@ def _evaluate_listtensor(e, self):
     """List tensors just turn into arrays."""
     ops = [self(o) for o in e.children]
     tmp = Result.empty(*ops)
-    arrs = []
-    for o in ops:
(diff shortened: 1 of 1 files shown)


miklos1:
Pushed [a firedrake branch for testing](https://jenkins.ese.ic.ac.uk:1080/blue/organizations/jenkins/firedrake/detail/miklos1%2Ftest-gem-interpreter-fix/2/pipeline/), merging if everything's fine.

miklos1:
> Pushed a firedrake branch for testing, merging if everything's fine.

Same failures as Firedrake master, so it's probably just fine, but why do we have failures on master?

wence-:
> Same failures as Firedrake master, so it's probably just fine, but why do we have failures on master?

I updated petsc without checking, they introduced a bug. It's now fixed, will push it tomorrow.

PR #225, opened 2020-08-18
Enable some zany H(div) elements  (Mardal-Tai-Winther and Arnold-Winther)

rckirby:
This takes UFL (https://github.com/FEniCS/ufl/pull/38 and FIAT), FIAT(https://github.com/FEniCS/fiat/pull/56), and FInAT (https://github.com/FInAT/FInAT/pull/63) live in the form compiler to enable Mardal-Tai-Winther and conforming and nonconforming Arnold-Winther elements.


tsfc/ufl_utils.py
```diff
@@ -333,6 +333,7 @@ def simplify_abs(expression, complex_mode):
     """Simplify absolute values in a UFL expression.  Its primary
     purpose is to "neutralise" CellOrientation nodes that are
     surrounded by absolute values and thus not at all necessary."""
+
```
miklos1, line 336: Unnecessary change.

tsfc/driver.py
```diff
@@ -342,9 +343,15 @@ def compile_expression_dual_evaluation(expression, to_element, coordinates, *,
...
                       ufl_cell=coordinates.ufl_domain().ufl_cell(),
+# <<<<<<< HEAD
+#                       precision=parameters["precision"],
+#                       argument_multiindices=argument_multiindices,
+#                       index_cache={})
+# =======
                       argument_multiindices=argument_multiindices,
                       index_cache={},
                       scalar_type=parameters["scalar_type"])
+# >>>>>>> origin/master
```
miklos1, line 354: Merge bork.

tsfc/driver.py
```diff
@@ -282,6 +282,7 @@ def compile_expression_dual_evaluation(expression, to_element, coordinates, *,
     :arg parameters: parameters object
     :arg coffee: compile coffee kernel instead of loopy kernel
     """
+
```
miklos1, line 285: Unnecessary change.

miklos1 (changes requested):

tsfc/driver.py
rckirby: Good catch, fixed.

tsfc/ufl_utils.py
rckirby: Not sure how it crept in.

tsfc/driver.py
rckirby: Not sure how it happened.  Putting it back.

miklos1:
1. I am somewhat wary of the profusion of methods on `PhysicalGeometry` since interfaces should be minimal, though this is currently an interface with a single implementation (when discounting mock implementations for test cases), so it is not actually a big deal.

2. Please edit the PR title to be more informative.

miklos1:
Pushed a temporary commit for testing with the correct branches...

miklos1 (approved):

miklos1:
Other than the concerns mentioned [above](https://github.com/firedrakeproject/tsfc/pull/225#issuecomment-676429629), I think this patch is now fine.  Of course, it has to wait until the corresponding branches elsewhere land...

miklos1 (approved):
This is fine.  I suggest merging with squashing commits, when it eventually comes to merging this.

rckirby:
> This is fine. I suggest merging with squashing commits, when it eventually comes to merging this.

I don't have write access on tsfc, so can you or Lawrence do this on the merge?

miklos1:
> I don't have write access on tsfc, so can you or Lawrence do this on the merge?

Okay, I did it, I hope things won't break...

rckirby:
Thanks.  Fingers crossed..

miklos1:
> I don't have write access on tsfc, so can you or Lawrence do this on the merge?

Though if you can push commits to branches within the TSFC repository, then you do have write access.  It's just that `master` is a protected branch, you cannot merge to master without getting an approve tick from someone else who also has write access.

PR #226, opened 2020-08-21
Interpolation onto a Vertex Only Mesh (Old)

ReubenHill:
Draft for changes needed to interpolate onto a Vertex Only Mesh from a suitable parent mesh.

Based on top of branch `dual_base_fiat` - see draft PR #218. Base should change to master when that PR is complete (or if this diverges and stops being dependent on #218).

Linked PRs: https://github.com/firedrakeproject/firedrake/pull/1815 & https://github.com/FInAT/FInAT/pull/66

### DIFF
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -410,7 +410,12 @@ class UFLtoGEMCallback(object):
         :param derivative: Maximum order of differentiation
         '''
         config = self.kernel_cfg.copy()
-        config.update(point_set=point_set)
+        if isinstance(point_set.expression, gem.Variable):
+            # config for fem.GemPointContext
+            config.update(point_indices=point_set.indices, point_expr=point_set.expression)
+        else:
+            # config for fem.PointSetContext
+            config.update(point_set=point_set)
 
         # Using expression directly changes scope
         dexpression = self.expression
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -362,6 +362,20 @@ def compile_expression_dual_evaluation(expression, element, *,
     else:
         return_arg = lp.GlobalArg("A", dtype=parameters["scalar_type"], shape=return_shape)
 
+    # TODO: Remove this horrible temporary hack
+    other_args = []
+    from finat.tensorfiniteelement import TensorFiniteElement
+    from finat.quadrature_element import QuadratureElement
+    if isinstance(element, TensorFiniteElement):
+        element = element.base_element
+    if isinstance(element, QuadratureElement):
+        expr = element._rule.point_set.expression
+        if isinstance(expr, gem.Variable):
+            if coffee:
+                other_args.append(ast.Decl(parameters["scalar_type"], ast.Symbol('X', rank=(expr.shape,))))
+            else:
+                other_args.append(lp.GlobalArg(expr.name, dtype=parameters["scalar_type"], shape=expr.shape))
+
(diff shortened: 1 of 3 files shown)


miklos1:
> I want to make a kernel for run time known points.
> I presume I ultimately want whatever represents my `gem.Variable('X', (tdim,))` to translate to a `TSFC.fem.GemPointContext` but I'll need to check this.

Yes, `GemPointContext` is exactly for evaluations at points that are only known at run-time, while `PointSetContext` is for compile-time known set of points.  However, this was designed with mere point evaluations in mind, how that exactly composes with generic dual evaluation is something I have not thought through.

ReubenHill:
> > I want to make a kernel for run time known points.
> > I presume I ultimately want whatever represents my `gem.Variable('X', (tdim,))` to translate to a `TSFC.fem.GemPointContext` but I'll need to check this.
> 
> Yes, `GemPointContext` is exactly for evaluations at points that are only known at run-time, while `PointSetContext` is for compile-time known set of points. However, this was designed with mere point evaluations in mind, how that exactly composes with generic dual evaluation is something I have not thought through.

Well for this PR at least I will just be doing point evaluations so we should be alright for now. Thanks for the heads up though

tsfc/driver.py
```diff
@@ -396,7 +396,12 @@ def __call__(self, point_set, derivative=0):
                 from ufl.differentiation import ReferenceGrad
                 # from ufl.operators import grad
                 config = kernel_cfg.copy()
-                config.update(point_set=point_set)
+                if isinstance(point_set.expression, gem.Variable):
+                    # config for fem.GemPointContext
+                    config.update(point_indices=point_set.indices, point_expr=point_set.expression)
+                else:
+                    # config for fem.PointSetContext
+                    config.update(point_set=point_set)
```
miklos1, line 404: The fact that you have to repeat this pattern four times should warn you (code smell!) that you probably should not masquerade your point expressions as point sets in the first place...

ReubenHill: Not my code - this needs to be rebased onto `dual_base_fiat` again - sorry for the confusion. See https://github.com/firedrakeproject/tsfc/pull/218/

ReubenHill: Oh wait I see what you are talking about now - I hadn't quite realised that this code (again - not actually mine)  is creating `PointSet`s every time. I think just creating a `ContextBase` to get the core `config`, then updating `kwargs` as appropriate might be the way to go? Or I can do more extensive checking of the `**kwargs` argument in `fem.compile_ufl`

wence-: Nope. I'm pretty sure @miklos1 is worried about the if isinstance repetition. That you have to update the config is probably fine. That you have to special case is a sign that the interface is badly designed.

miklos1: Yes, Lawrence is right.  Now, I cannot tell what the proper refactoring is without properly thinking the problem through, but this symptom is a clear sign, that something is wrong here.

ReubenHill: See up to date diff now that I've rebased onto the refactored  #218 (which is itself now based on the recent refactors you made @miklos1 ) - there's only one `isinstance` right now so the interface might not be the worst.

ReubenHill:
Replaced by #247 which is based on top of master

ReubenHill:
Note to self: there are no unresolved relevant comments to move to #247

PR #229, opened 2020-08-25
More magic methods for gem Nodes

wence-:
Adds a `__matmul__` method, and a `componentwise` helper so the existing
sugar also works transparently for shaped objects.

Motivated by the gem->numpy->gem dance in FInAT/FInAT#63

### DIFF
--- a/gem/gem.py
+++ b/gem/gem.py
@@ -80,25 +80,52 @@ class Node(NodeBase, metaclass=NodeMeta):
         return Indexed(self, indices)
 
     def __add__(self, other):
-        return Sum(self, as_gem(other))
+        return componentwise(Sum, self, as_gem(other))
 
     def __radd__(self, other):
         return as_gem(other).__add__(self)
 
     def __sub__(self, other):
-        return Sum(self, Product(Literal(-1), as_gem(other)))
+        return componentwise(
+            Sum, self,
+            componentwise(Product, Literal(-1), as_gem(other)))
 
     def __rsub__(self, other):
         return as_gem(other).__sub__(self)
 
     def __mul__(self, other):
-        return Product(self, as_gem(other))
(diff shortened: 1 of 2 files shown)


miklos1:
I think this is good and well motivated by the example of the FInAT PR.

rckirby:
These are welcome changes.  However, If you see my comments on the FInAT PR, some of the gem <--> numpy dance is motivated by needing to build a `V` matrix/tensor by setting specific entries or blocks in a sparse matrix.  Hence I start with a `numpy` array of zero objects and convert it (at the end) to `gem`.  I need to be able set items/slices of `ComponentTensor` to do something like `V[0:3, 0:3] = J.T @ J` (not the right math, but gives a sample of something like what is needed.  Otherwise, I may still be stuck with some gem <--> numpy business.

wence-:
If the rvalue is a gem expression but the lvalue is an array, I think that's ok. In pure gem there is wrapping in a listtensor to achieve this. 

There were just a bunch of places where it looked unnecessary, And potentially would create more temporaries in the final code.

rckirby:
How to proceed?  We could:
1.) Land this PR on tsfc and then change FInAT mappings to use it before we
merge
or
2.) Go ahead and merge the FInAT PR and after this tsfc one lands, go back
and systematically change all the zany mappings (even the existing ones).

wence-:
Maybe we should do the necessary to merge the finat PR "as is", then confirm that we can port everything we need to into just gem where possible on this branch (and make sure things work). And go round and fix all the zany elements. (So option 2).

rckirby:
OK, I'll go back over comments on the FInAT PR and address the outstanding
things not related to gem updates so we can merge.

miklos1 (approved):
I think this is fine.

Issue #232, opened 2020-10-06
Tidy up interpolation tsfc<->firedrake interface


ReubenHill:
From slack, @wence- said
> the interpolation tsfc<->firedrake interface needs to be cleaned up so that we don't have to send data-carrying objects in. Right now, we generate the list of parloop args from the coefficients that complex_expression_dual_evaluation sends back

> So if we mimic the idea for form compilation, what happens is that the initial form provides form.coefficients() in a canonical order
> and the kernel that tsfc sends back provides indices into that list of which coefficients are needed
> so you iterate over the indices and get the coefficients
> it also sends back some extra info like "do I need the cell orientations, etc..."
> one could imagine doing the same thing for interpolation kernels

So the new return values for `compile_expression_dual_evaluation` would be 
 - the kernel,
 - ast, 
 - and a bunch of switches and indices that somehow lets firedrake work out how to get the right arguments for the par loop from the expression, FInAT element and coordinates `Firedrake.Function`.

Perhaps one wouldn't need to pass in the element and coordinates at all?
```
def compile_expression_dual_evaluation(expression, element, coordinates, *,
                                       domain=None, interface=None,
                                       parameters=None, coffee=False):
```
could become
```
def compile_expression_dual_evaluation(expression, *,
                                       domain=None, interface=None,
                                       parameters=None, coffee=False):

ReubenHill:
Correction: we definitely need the `element` to know who's dual we are evaluating. We don't need `coordinates` because it's not needed for working out the pullback to the reference to cell - you can either get it from the expression or from the `domain` kwarg.

PR #233, opened 2020-10-09
Tidy up interpolation tsfc<->firedrake interface

ReubenHill:
See issue #232. Coupled with https://github.com/firedrakeproject/firedrake/pull/1861. Please feel free to bikeshed this whilst it's WIP!

### DIFF
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -332,15 +332,16 @@ def compile_expression_dual_evaluation(expression, to_element, *,
         domain = expression.ufl_domain()
     assert domain is not None
 
-    # Create a fake coordinate coefficient for a domain.
-    coords_coefficient = ufl.Coefficient(ufl.FunctionSpace(domain, domain.ufl_coordinate_element()))
-    builder.domain_coordinate[domain] = coords_coefficient
-    builder.set_cell_sizes(domain)
-
     # Collect required coefficients
+    first_coefficient_fake_coords = False
     coefficients = extract_coefficients(expression)
     if has_type(expression, GeometricQuantity) or any(fem.needs_coordinate_mapping(c.ufl_element()) for c in coefficients):
+        # Create a fake coordinate coefficient for a domain.
+        coords_coefficient = ufl.Coefficient(ufl.FunctionSpace(domain, domain.ufl_coordinate_element()))
+        builder.domain_coordinate[domain] = coords_coefficient
+        builder.set_cell_sizes(domain)
         coefficients = [coords_coefficient] + coefficients
+        first_coefficient_fake_coords = True
     builder.set_coefficients(coefficients)
 
     # Split mixed coefficients
@@ -433,7 +434,7 @@ def compile_expression_dual_evaluation(expression, to_element, *,
     # Handle kernel interface requirements
     builder.register_requirements([ir])
     # Build kernel tuple
-    return builder.construct_kernel(return_arg, impero_c, index_names)
+    return builder.construct_kernel(return_arg, impero_c, index_names, first_coefficient_fake_coords)
 
 
 def lower_integral_type(fiat_cell, integral_type):
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -267,7 +267,7 @@ def compile_integral(integral_data, form_data, prefix, parameters, interface, co
     return builder.construct_kernel(kernel_name, impero_c, index_names, quad_rule)
 
 
-def compile_expression_dual_evaluation(expression, to_element, coordinates, *,
+def compile_expression_dual_evaluation(expression, to_element, *,
                                        domain=None, interface=None,
                                        parameters=None, coffee=False):
     """Compile a UFL expression to be evaluated against a compile-time known reference element's dual basis.
@@ -276,8 +276,7 @@ def compile_expression_dual_evaluation(expression, to_element, coordinates, *,
 
     :arg expression: UFL expression
     :arg to_element: A FInAT element for the target space
-    :arg coordinates: the coordinate function
-    :arg domain: optional UFL domain the expression is defined on (useful when expression contains no domain).
+    :arg domain: optional UFL domain the expression is defined on (required when expression contains no domain).
     :arg interface: backend module for the kernel interface
     :arg parameters: parameters object
     :arg coffee: compile coffee kernel instead of loopy kernel
@@ -328,17 +327,20 @@ def compile_expression_dual_evaluation(expression, to_element, coordinates, *,
     argument_multiindices = tuple(builder.create_element(arg.ufl_element()).get_indices()
                                   for arg in arguments)
 
-    # Replace coordinates (if any)
-    domain = expression.ufl_domain()
-    if domain:
-        assert coordinates.ufl_domain() == domain
-        builder.domain_coordinate[domain] = coordinates
-        builder.set_cell_sizes(domain)
+    # Replace coordinates (if any) unless otherwise specified by kwarg
+    if domain is None:
+        domain = expression.ufl_domain()
+    assert domain is not None
+
+    # Create a fake coordinate coefficient for a domain.
+    coords_coefficient = ufl.Coefficient(ufl.FunctionSpace(domain, domain.ufl_coordinate_element()))
+    builder.domain_coordinate[domain] = coords_coefficient
+    builder.set_cell_sizes(domain)
 
     # Collect required coefficients
     coefficients = extract_coefficients(expression)
     if has_type(expression, GeometricQuantity) or any(fem.needs_coordinate_mapping(c.ufl_element()) for c in coefficients):
-        coefficients = [coordinates] + coefficients
+        coefficients = [coords_coefficient] + coefficients
     builder.set_coefficients(coefficients)
 
     # Split mixed coefficients
@@ -346,7 +348,7 @@ def compile_expression_dual_evaluation(expression, to_element, coordinates, *,
 
     # Translate to GEM
     kernel_cfg = dict(interface=builder,
-                      ufl_cell=coordinates.ufl_domain().ufl_cell(),
+                      ufl_cell=domain.ufl_cell(),
                       argument_multiindices=argument_multiindices,
                       index_cache={},
                       scalar_type=parameters["scalar_type"])
--- a/tsfc/kernel_interface/firedrake_loopy.py
+++ b/tsfc/kernel_interface/firedrake_loopy.py
@@ -17,7 +17,7 @@ from tsfc.loopy import generate as generate_loopy
 
 
 # Expression kernel description type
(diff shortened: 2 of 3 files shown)


ReubenHill:
So my broad impression is that we want to make `ExpressionKernelBuilder` more like `KernelBuilder` which seems to require much less firedrake specific data. Is that a fair interpretation? Any suggestions on how best to proceed @wence- ? I feel that what I've done so far kind of side steps the issue with inputting coordinates. They are still there just hidden within the domain.

wence-:
> So my broad impression is that we want to make ExpressionKernelBuilder more like KernelBuilder which seems to require much less firedrake specific data. Is that a fair interpretation? Any suggestions on how best to proceed @wence- ? I feel that what I've done so far kind of side steps the issue with inputting coordinates. They are still there just hidden within the domain.

Suppose I give you just the generated kernel (the loopy thing). And I say "this kernel matches up with interpolating this expression". And then the only thing I permit you to ask of the expression is "what are your coefficients" and "what is your domain". What information do I need to construct the parloop that executes the kernel?

ReubenHill:
I would need to match the domain to a PyOP2 iteration set, match the coefficients to the par loop arguments (which we ought to do with something like `coefficient_numbers` that you get when you make a kernel from compiling a form), the access descriptors for the coefficients, and any special information about how the iterset should be traversed (I presume this is what `oriented` and `needs_cell_sizes` are for?)

Sound right or wrong?

Also this seems to be an answer to a different question - I am, as far as I'm aware, not in a situation where I have such a kernel - rather I'm trying to work out how you build one. The answer to your question seems one step ahead.

ReubenHill:
> OK, I see. Basically, inside TSFC, if you want to disentangle the symbolics from the numerics you're only allowed to called UFL methods on the objects. IOW, you want to get to a situation where if you just had a fully symbolic expression, `compile_expression_dual_evaluation` would still work.
> 
> ```python
> from ufl import *
> from tsfc.finatinterface import create_element
> V = VectorElement("P", triangle, 1)
> mesh = Mesh(V)
> 
> V = FunctionSpace(mesh, FiniteElement("P", triangle, 2))
> x, y = SpatialCoordinate(mesh)
> 
> expr = x*y - y**2 + x
> element = create_element(V.ufl_element())
> stuff = compile_expression_dual_evaluation(expr, element, coffee=False)
> ```

This example should fail I believe - a function of spatial coordinates surely requires the coordinates data.

wence-:
> This example should fail I believe - a function of spatial coordinates surely requires the coordinates data.

Only when _executing_, not when compiling.

ReubenHill:
> > This example should fail I believe - a function of spatial coordinates surely requires the coordinates data.
> 
> Only when _executing_, not when compiling.

So I can execute this
```python
    V = ufl.VectorElement("P", ufl.triangle, 1)
    mesh = ufl.Mesh(V)
    V = ufl.FunctionSpace(mesh, ufl.FiniteElement("P", ufl.triangle, 2))
    v = ufl.Coefficient(V)
    expr = ufl.inner(v, v)
    element = create_element(V.ufl_element())
    ast, oriented, needs_cell_sizes, coefficients, _ = compile_expression_dual_evaluation(expr, element, coffee=False)
```
and the `ast` has no `code` property. Am I aiming for similar behaviour with an expression of spatial coordinates?

wence-:
> and the ast has no code property. Am I aiming for similar behaviour with an expression of spatial coordinates?

I think we are talking at cross-purposes.

I claim that for generation of an interpolation kernel (which is what this is doing), I should need no global (Firedrake-level) information. But rather I should only need symbolic information (what the elements are, what the expression is, etc...).

I proposed one way of checking if this is the case right now is to feed _purely symbolic_ objects (from UFL) into `compile_expression_dual_evaluation`. The kernel code that comes out should be identical to if I had fed in Firedrake objects.

ReubenHill:
> > and the ast has no code property. Am I aiming for similar behaviour with an expression of spatial coordinates?
> 
> I think we are talking at cross-purposes.
> 
> I claim that for generation of an interpolation kernel (which is what this is doing), I should need no global (Firedrake-level) information. But rather I should only need symbolic information (what the elements are, what the expression is, etc...).
> 
> I proposed one way of checking if this is the case right now is to feed _purely symbolic_ objects (from UFL) into `compile_expression_dual_evaluation`. The kernel code that comes out should be identical to if I had fed in Firedrake objects.

Okay - in a discussion with @dham he conjectured that if my expression is one of spatial coordinates or non-trivial pullback, I should need global domain information from the `domain.coordinates` coefficient (or we misunderstood each other). Why ought this not to be the case? Do the FInAT elements contain enough information already?

Put another way, how do I deal with `has_type(expression, GeometricQuantity) == true` to get past these lines:
```
    # Collect required coefficients
    coefficients = extract_coefficients(expression)
    if has_type(expression, GeometricQuantity) or any(fem.needs_coordinate_mapping(c.ufl_element()) for c in coefficients):
        coefficients = [coordinates] + coefficients
    builder.set_coefficients(coefficients)
```
([permalink](https://github.com/firedrakeproject/tsfc/blob/b043e8654a68345c524809953fc8f0705dc0d6eb/tsfc/driver.py#L338-L342))?

Does a correctly refactored `ExpressionKernelBuilder` not actually require the `coordinates` coefficient?

wence-:
> Does a correctly refactored ExpressionKernelBuilder not actually require the coordinates coefficient?

It needs a symbolic equivalent of that coefficient. A `Coefficient(coordinates.ufl_function_space())` should work fine.

You can't possibly need global domain information (numbers) when you're compiling a kernel. You can only need the symbolic information that tells you how numbers are to be interpreted.

https://github.com/firedrakeproject/tsfc/blob/b043e8654a68345c524809953fc8f0705dc0d6eb/tsfc/driver.py#L136

What does that function do?

https://github.com/firedrakeproject/tsfc/blob/b043e8654a68345c524809953fc8f0705dc0d6eb/tsfc/kernel_interface/firedrake_loopy.py#L221-L229

So it makes a symbolic representation of the coordinate field from the mesh's coordinate element. No Firedrake object required.

wence-:
Sorry, edited my comments for tetchiness.

ReubenHill:
Okay so it looks like I've got everything compiling using UFL only. Now I need to work out how to get all the necessary coefficient processing information (which I need, do I need to split coefficients) out of `compile_expression_dual_evaluation`.  I'm wary of doing much more until #234 has been reviewed since so much of what I will do will copy `KernelBuilder`.

ReubenHill:
This now works - I wonder if there's a better way of dealing with the coordinates coefficient but perhaps this is good enough for now?

tsfc/driver.py
```diff
@@ -267,7 +267,7 @@ def name_multiindex(multiindex, name):
     return builder.construct_kernel(kernel_name, impero_c, index_names, quad_rule)
 
 
-def compile_expression_dual_evaluation(expression, to_element, coordinates, *,
+def compile_expression_dual_evaluation(expression, to_element, *,
```
ReubenHill, line 270: `to_element` should eventually become a UFL cofunction whenever they are implemented

dham (approved):

tsfc/kernel_interface/firedrake_loopy.py
```diff
@@ -153,12 +153,14 @@ def register_requirements(self, ir):
...
 
-    def construct_kernel(self, return_arg, impero_c, index_names):
+    def construct_kernel(self, return_arg, impero_c, index_names, first_coefficient_fake_coords):
         """Constructs an :class:`ExpressionKernel`.
 
         :arg return_arg: loopy.GlobalArg for the return value
         :arg impero_c: gem.ImperoC object that represents the kernel
         :arg index_names: pre-assigned index names
+        :arg first_coefficient_fake_coords: If true, the kernel's first
+            coefficient is a constructed UFL coordinate field
```
wence-, line 163: TBH, this is a horrible variable name. I think what this means is that the kernel expects to be fed:

```
[mesh.coordinates] + extract_coefficients(expression)
```

Is that right?

ReubenHill: Didn't see this comment in time. Yes that is correct

ReubenHill: If you have a suggested better interface and/or variable name to say and I can make a new PR

tsfc/driver.py
```diff
@@ -328,25 +327,29 @@ def compile_expression_dual_evaluation(expression, to_element, coordinates, *,
...
     coefficients = extract_coefficients(expression)
     if has_type(expression, GeometricQuantity) or any(fem.needs_coordinate_mapping(c.ufl_element()) for c in coefficients):
-        coefficients = [coordinates] + coefficients
+        # Create a fake coordinate coefficient for a domain.
+        coords_coefficient = ufl.Coefficient(ufl.FunctionSpace(domain, domain.ufl_coordinate_element()))
+        builder.domain_coordinate[domain] = coords_coefficient
+        builder.set_cell_sizes(domain)
+        coefficients = [coords_coefficient] + coefficients
+        first_coefficient_fake_coords = True
     builder.set_coefficients(coefficients)
```
ksagiyam, line 346: The way `coordinate` is taken care of in the above is identical to how we do in `KernelBuilder.set_coordinate`, so maybe one potential future change is to make `set_coordinate` method also available in `ExpressionKernelBuilder`.

tsfc/kernel_interface/firedrake_loopy.py
sv2518: I think in various places in Firedrake we have similar variables to this e.g. for cell sizes. They are called something like `needs_cell_sizes`, maybe you just want to rename to `needs_coords`.

tsfc/driver.py
ReubenHill: I strongly considered doing this, but didn't want to make a "fake" version of that interface without recreating it more fully. Hopefully once your changes to that interface have landed this can be given another spin.

tsfc/kernel_interface/firedrake_loopy.py
```diff
@@ -173,7 +175,8 @@ def construct_kernel(self, return_arg, impero_c, index_names):
         loopy_kernel = generate_loopy(impero_c, args, self.scalar_type,
                                       "expression_kernel", index_names)
         return ExpressionKernel(loopy_kernel, self.oriented, self.cell_sizes,
-                                self.coefficients, self.tabulations)
+                                self.coefficients, first_coefficient_fake_coords,
```
ksagiyam, line 178: Another thing I think we should think about next time is how to make `ExpressionKernelBuilder` more symmetric to `KernelBuilder`. I think instead of returning `coefficients` here, we should just call `extract_coefficients(expr)` again in Firedrake (this compares to `form.coefficients()` called in `assemble.py`)  and split `Function`s if necessary when setting up `ParLoop` arguments (again, as is done in `assemble.py`).

PR #237, opened 2021-02-02
Ensure compile_ufl output has correct free indices

ReubenHill:
When an expression given to compile_ufl does not have points in the
input PointSet in its expression tree, the resulting gem expression has
missing free indices. This attempts to fix that.

Todo: 
 - [x] Fix failing tests

### DIFF
--- a/gem/optimise.py
+++ b/gem/optimise.py
@@ -102,6 +102,8 @@ def replace_indices_delta(node, self, subst):
     j = substitute.get(node.j, node.j)
     if i == node.i and j == node.j:
         return node
+    elif i == j:
+        return one
     else:
         return Delta(i, j)
 
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -377,7 +377,11 @@ def compile_expression_dual_evaluation(expression, to_element, *,
             quad_rule = QuadratureRule(point_set, to_element._weights)
             config["quadrature_rule"] = quad_rule
 
+        # evaluate expression at the points specified by point_set
         expr, = fem.compile_ufl(expression, **config, point_sum=False)
+        # the point set free indices should now be free indices of the compiled
+        # expression
+        assert set(point_set.indices) <= set(expr.free_indices)
         shape_indices = tuple(gem.Index() for _ in expr.shape)
         basis_indices = point_set.indices
         ir = gem.Indexed(expr, shape_indices)
@@ -395,7 +399,11 @@ def compile_expression_dual_evaluation(expression, to_element, *,
                 point_set = PointSet(pts)
                 config = kernel_cfg.copy()
                 config.update(point_set=point_set)
+                # evaluate expression at the points specified by point_set
                 expr, = fem.compile_ufl(expression, **config, point_sum=False)
+                # the point set free indices should now be free indices of the
+                # compiled expression
+                assert set(point_set.indices) <= set(expr.free_indices)
                 expr = gem.partial_indexed(expr, shape_indices)
                 expr_cache[pts] = expr, point_set
             weights = collections.defaultdict(list)
--- a/tsfc/fem.py
+++ b/tsfc/fem.py
@@ -3,7 +3,8 @@ geometric quantities into GEM expressions."""
 
 import collections
 import itertools
-from functools import singledispatch
+from functools import singledispatch, reduce
+import operator
 
 import numpy
 
@@ -697,6 +698,15 @@ def compile_ufl(expression, interior_facet=False, point_sum=False, **kwargs):
 
     # Translate UFL to GEM, lowering finite element specific nodes
     result = map_expr_dags(context.translator, expressions)
+    # Check that any indices from the PointSet have made it into the resulting
+    # expression. If not, blow up shape to include the missing indices using
+    # deltas - these will eventually be cancelled out but ensure each
+    # expression in our result has the expected free_indices and shape.
+    result = [reduce(operator.mul, (gem.Delta(j, j)
+                                    for j in context.point_indices
+                                    if j not in expr.free_indices),
+                     expr)
+              for expr in result]
     if point_sum:
         result = [gem.index_sum(expr, context.point_indices) for expr in result]
     return result
--- a/gem/gem.py
+++ b/gem/gem.py
@@ -792,10 +792,6 @@ class Delta(Scalar, Terminal):
         assert isinstance(i, IndexBase)
         assert isinstance(j, IndexBase)
 
-        # \delta_{i,i} = 1
-        if i == j:
-            return one
-
         # Fixed indices
         if isinstance(i, int) and isinstance(j, int):
             return Literal(int(i == j))


tsfc/fem.py
```diff
@@ -699,4 +699,17 @@ def compile_ufl(expression, interior_facet=False, point_sum=False, **kwargs):
     result = map_expr_dags(context.translator, expressions)
     if point_sum:
         result = [gem.index_sum(expr, context.point_indices) for expr in result]
+    # Check that any indices from the PointSet have made it into the resulting expression
+    for i in range(len(result)):
+        expansion_indices = ()
+        for index in context.point_indices:
+            if index not in result[i].free_indices:
+                expansion_indices = expansion_indices + (index,)
```
wence-, line 707: Strange way to write a set difference

`expand = tuple(i for i in indices if i not in free_indices)`

tsfc/fem.py
```diff
@@ -699,4 +699,17 @@ def compile_ufl(expression, interior_facet=False, point_sum=False, **kwargs):
...
+        if expansion_indices != ():
+            # blow up shape to include expansion_indices
+            deltas = gem.Delta(expansion_indices[0], expansion_indices[0])
+            # Use deltas to increase the possible rank of our expression
+            for index in expansion_indices[1:]:
+                deltas *= gem.Delta(index, index)
+            result[i] = gem.Indexed(gem.ComponentTensor(result[i] * deltas, expansion_indices), expansion_indices)
```
wence-, line 714: `functools.reduce(gem.Product, (gem.Delta(i, i) for i in indices), ())`

no need for the guard because `()` is the identity for Indexed and CT

tsfc/fem.py
ReubenHill: Best way to learn how to do something the correct way: publicly show the wrong way and await correction :)

tsfc/fem.py
ReubenHill: Apparently not? This fails with 
```
../tsfc/tsfc/fem.py:708: in compile_ufl
    deltas = functools.reduce(gem.Product, (gem.Delta(j, j) for j in expansion_indices), ())
../tsfc/gem/gem.py:48: in __call__
    obj = super(NodeMeta, self).__call__(*args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

cls = <class 'gem.gem.Product'>, a = ()
b = Delta(Index(2), Index(2))

    def __new__(cls, a, b):
>       assert not a.shape
E       AttributeError: 'tuple' object has no attribute 'shape'
```
seems I do need to guard

wence-: Oh sorry, `()` is not closed in gem.

```
deltas = () if not deltas else functools.reduce(gem.Product, deltas)
```

ReubenHill:
Sadly it seems that quite a lot of `gem` assumes that `gem.Delta(j,j)` returns `Literal(1.)` :/

gem/optimise.py
```diff
@@ -102,6 +102,8 @@ def replace_indices_delta(node, self, subst):
     j = substitute.get(node.j, node.j)
     if i == node.i and j == node.j:
         return node
+    elif i == j:
+        return one
```
ReubenHill, line 106: Okay think I've got it working with this change

tsfc/fem.py
ReubenHill: Don't quite follow your second suggestion here I'm afraid

tsfc/fem.py
```diff
@@ -697,6 +697,16 @@ def compile_ufl(expression, interior_facet=False, point_sum=False, **kwargs):
...
+    # Check that any indices from the PointSet have made it into the resulting expression
+    import functools
+    for i in range(len(result)):
+        expansion_indices = tuple(j for j in context.point_indices if j not in result[i].free_indices)
+        # blow up shape to include expansion_indices using deltas - these will
+        # eventually be cancelled out but ensure the our result has the expected
+        # free_indices and shape.
+        if expansion_indices != ():
+            deltas = functools.reduce(gem.Product, (gem.Delta(j, j) for j in expansion_indices))
+            result[i] = result[i] * deltas
```
wence-, line 709:
```
result[i] = functools.reduce(gem.Product, (gem.Delta(j, j) for j in expansion_indices), result[i])
```

I think

tsfc/fem.py
wence-: added a comment.

tsfc/fem.py
ReubenHill: Needed `operator.mul` since `gem.Product` doesn't work if `result[i]` isn't scalar but otherwise works nicely 👍

tsfc/fem.py
```diff
@@ -697,6 +697,15 @@ def compile_ufl(expression, interior_facet=False, point_sum=False, **kwargs):
 
     # Translate UFL to GEM, lowering finite element specific nodes
     result = map_expr_dags(context.translator, expressions)
+    # Check that any indices from the PointSet have made it into the resulting expression
+    import functools
+    import operator
```
wence-, line 702: Can you move these imports to top-level please?

tsfc/driver.py
```diff
@@ -377,7 +377,11 @@ def compile_expression_dual_evaluation(expression, to_element, *,
             quad_rule = QuadratureRule(point_set, to_element._weights)
             config["quadrature_rule"] = quad_rule
 
+        # evaluate expression at the points specified by point_set
         expr, = fem.compile_ufl(expression, **config, point_sum=False)
+        # the point set free indices should now be free indices of the complied
+        # expression
+        assert all(i in expr.free_indices for i in point_set.indices)
```
wence-, line 384: I think this is:
```
set(point_set.indices) <= set(expr.free_indices)
```
?
Which is clearer, I think.

tsfc/fem.py
```diff
@@ -697,6 +697,15 @@ def compile_ufl(expression, interior_facet=False, point_sum=False, **kwargs):
...
+    for i in range(len(result)):
+        expansion_indices = tuple(j for j in context.point_indices if j not in result[i].free_indices)
+        # blow up shape to include expansion_indices using deltas - these will
+        # eventually be cancelled out but ensure the our result has the expected
+        # free_indices and shape.
+        result[i] = functools.reduce(operator.mul, (gem.Delta(j, j) for j in expansion_indices), result[i])
```
wence-, line 708: Given that this loop always produces new results:

```python
result = [functools.reduce(operator.mul, (gem.Delta(j, j)
                                          for j in context.point_indices
                                          if j not in expr.free_indices),
                           expr)
          for expr in result]
```

Makes it clearer that we're modifying every result entry.

tsfc/driver.py
```diff
@@ -395,7 +399,11 @@ def compile_expression_dual_evaluation(expression, to_element, *,
                 point_set = PointSet(pts)
                 config = kernel_cfg.copy()
                 config.update(point_set=point_set)
+                # evaluate expression at the points specified by point_set
                 expr, = fem.compile_ufl(expression, **config, point_sum=False)
+                # the point set free indices should now be free indices of the
+                # complied expression
+                assert all(i in expr.free_indices for i in point_set.indices)
```
wence-, line 406: Here too, I think `assert set(point_set.indices) <= set(expr.indices)` makes the assertion clearer.

wence- (changes requested):
Minor fixes

wence-:
Very minor, your commit message for the compile_ufl change:
```
Ensure compile_ufl output has correct free indices  …
When an expression given to compile_ufl does not have points in the
input PointSet in its expression tree, the resulting gem expression has
missing free indices. This attempts to fix that.
```

I think it does fix it, so remove "This attempts to fix that" :).

wence-:
Can you add a test that this is doing TRT? e.g. that `compile_ufl(expr_that_doesn_depend_on_point_set)` has the right free indices now?

ReubenHill:
> Can you add a test that this is doing TRT? e.g. that `compile_ufl(expr_that_doesn_depend_on_point_set)` has the right free indices now?

I'm not quite sure of a nice way to set up such a test. Because of the weird interface that `fem.compile_ufl` has I need to first set up a whole kernel config which takes me close to having to reimplement much of `compile_expression_dual_evaluation`. Could the assertions do?

tsfc/fem.py
ReubenHill: Done in d711e722d19ce86ae5fca8b60e01492bd55e6ae6

tsfc/fem.py
ReubenHill: Nice. Done in d711e722d19ce86ae5fca8b60e01492bd55e6ae6

tsfc/driver.py
```diff
@@ -395,7 +399,11 @@ def compile_expression_dual_evaluation(expression, to_element, *,
                 point_set = PointSet(pts)
                 config = kernel_cfg.copy()
                 config.update(point_set=point_set)
+                # evaluate expression at the points specified by point_set
                 expr, = fem.compile_ufl(expression, **config, point_sum=False)
+                # the point set free indices should now be free indices of the
+                # complied expression
```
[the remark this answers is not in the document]
ReubenHill: fixed in d711e722d19ce86ae5fca8b60e01492bd55e6ae6

tsfc/driver.py
ReubenHill: Oh cool, much neater and introduces me to python's sets too :) . Changed in d711e722d19ce86ae5fca8b60e01492bd55e6ae6

tsfc/driver.py
```diff
@@ -377,7 +377,11 @@ def compile_expression_dual_evaluation(expression, to_element, *,
             quad_rule = QuadratureRule(point_set, to_element._weights)
             config["quadrature_rule"] = quad_rule
 
+        # evaluate expression at the points specified by point_set
         expr, = fem.compile_ufl(expression, **config, point_sum=False)
+        # the point set free indices should now be free indices of the complied
```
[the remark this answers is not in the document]
ReubenHill: Changed in d711e722d19ce86ae5fca8b60e01492bd55e6ae6

tsfc/driver.py
ReubenHill: Changed in d711e722d19ce86ae5fca8b60e01492bd55e6ae6

wence-:
> I'm not quite sure of a nice way to set up such a test. Because of the weird interface that fem.compile_ufl has I need to first set up a whole kernel config which takes me close to having to reimplement much of compile_expression_dual_evaluation. Could the assertions do?

OK, I see. yes I think that's ok.

miklos1:
This feels to me like scratching your face with your elbow behind your back.  Surely, the right way would be that the caller calmly accepts the fact, that the compiled expression may be constant along some of the point set indices, and then it deals with the situation accordingly.  Leaving "mathematical one" nodes in the expression with fake index dependencies does not sound like a clean solution.

wence-:
> This feels to me like scratching your face with your elbow behind your back. Surely, the right way would be that the caller calmly accepts the fact, that the compiled expression may be constant along some of the point set indices, and then it deals with the situation accordingly. Leaving "mathematical one" nodes in the expression with fake index dependencies does not sound like a clean solution.

So there are two parts here. The early simplification of `delta(i, i) => one` I think is kind of wrong (because then as noted `IndexSum(delta(i, i), (i, ))` produces the wrong thing.

I agree that using `delta(i, i)` as a fake way of producing a "broadcast over these indices" is a bit messy. Perhaps a better way would be to introduce a `Broadcast(expr, indices)` node and handle it later. Then there's a philosophical point of whether this should be handled inside `compile_ufl` or outside.

ReubenHill:
> This feels to me like scratching your face with your elbow behind your back. Surely, the right way would be that the caller calmly accepts the fact, that the compiled expression may be constant along some of the point set indices, and then it deals with the situation accordingly. Leaving "mathematical one" nodes in the expression with fake index dependencies does not sound like a clean solution.

Are you advocating not changing the behaviour of `fem.compile_ufl`? The trouble is that if I want to do anything with the expression after calling `fem.compile_ufl` then then the number of free indices it has is unpredictable without this fix - it's silently dependent on whether the points in the point set turn up in the expression

ReubenHill:
@dham was the person who first strongly advocated that this was a bug - he may want to add his thoughts

miklos1:
> The early simplification of `delta(i, i) => one` I think is kind of wrong (because then as noted `IndexSum(delta(i, i), (i, ))` produces the wrong thing.

I suppose `IndexSum(one, (i,))` should simplify to `Literal(i.extent)`, though I admit in other cases it is useful to have an `index_sum` helper function which just ignores indices not present in the integrand.

> I agree that using `delta(i, i)` as a fake way of producing a "broadcast over these indices" is a bit messy. Perhaps a better way would be to introduce a `Broadcast(expr, indices)` node and handle it later.

Perhaps.  Where is it used?  See more in another answer below.

miklos1:
> The trouble is that if I want to do anything with the expression after calling `fem.compile_ufl` then then the number of free indices it has is unpredictable without this fix - it's silently dependent on whether the points in the point set turn up in the expression

I would need to see how you exactly use `compile_ufl`, but my preliminary opinion is that this is indeed the correct behaviour.
Cases like this have been dealt with in TSFC: for example, when interpolating `grad(f)` to a vector-DG1 space, it might happen that `grad(f)` is cellwise constant because `f` was linear.  In this case the same values are written in each target DG node of the cell.  `impero_utils.Assignment` accepts right-hand sides whose free indices are a proper subset of the free indices of the left-hand side.

dham:
> > The early simplification of `delta(i, i) => one` I think is kind of wrong (because then as noted `IndexSum(delta(i, i), (i, ))` produces the wrong thing.
> 
> I suppose `IndexSum(one, (i,))` should simplify to `Literal(i.extent)`, though I admit in other cases it is useful to have an `index_sum` helper function which just ignores indices not present in the integrand.
> 
> > I agree that using `delta(i, i)` as a fake way of producing a "broadcast over these indices" is a bit messy. Perhaps a better way would be to introduce a `Broadcast(expr, indices)` node and handle it later.
> 
> Perhaps. Where is it used? See more in another answer below.

I think the `index_sum` is a red herring here. The question to ask yourself is, what are the shape and free indices of `delta(i, i)`  and `one`. I think that the former is scalar with one free index and the latter is scalar with no free indices. This makes the simplification invalid.

ReubenHill:
> > The trouble is that if I want to do anything with the expression after calling `fem.compile_ufl` then then the number of free indices it has is unpredictable without this fix - it's silently dependent on whether the points in the point set turn up in the expression
> 
> I would need to see how you exactly use `compile_ufl`, but my preliminary opinion is that this is indeed the correct behaviour.
> Cases like this have been dealt with in TSFC: for example, when interpolating `grad(f)` to a vector-DG1 space, it might happen that `grad(f)` is cellwise constant because `f` was linear. In this case the same values are written in each target DG node of the cell. `impero_utils.Assignment` accepts right-hand sides whose free indices are a proper subset of the free indices of the left-hand side.

See [this comment](https://github.com/FInAT/FInAT/pull/57#issuecomment-746247537) in https://github.com/FInAT/FInAT/pull/57 which I am trying to tidy and land. 

I am currently putting together a tensor of weights `Q` which I right multiply by a vector of function point evaluations (the function of interest here being whatever the ufl expression we are compiling with `gem_expr = fem.compile_ufl(ufl_expression,...)` is and the points are some `FInAT.PointSet` `point_set`). 

If the `ufl_expression` function/expression happens to not be dependent on the points given to it, instead of getting `gem_expr` to be a vector of point evaluations (where some of the free indices are equal to `point_set.indices`) I get a scalar. To do my tensor contraction I then need to have a special case to deal with this quirk of `fem.compile_ufl`.

It seems to me better to get `fem.compile_ufl` to return something with the expected free indices.

miklos1:
> I think the `IndexSum` is a red herring here. The question to ask yourself is, what are the shape and free indices of `delta(i, i)` and `one`. I think that the former is scalar with one free index and the latter is scalar with no free indices. This makes the simplification invalid.

Well, yes and no.  From a formal point of view, you are right, since `Delta(i, i)` is an expression which contains `i` without `i` being bound, so it does have`i` as a free index, while `one` does not have `i` as a free index.  However, they are substantially equivalent, since for all valid values of `i`: `Delta(i, i)` = `one`.  So, considering the purpose of filling the result tensor with the correct values, replacing `Delta(i, i)` with `one` is legit.

To apply optimisations, that invariably simplify the expression, as early as possible, is *usually* the right thing to do.  The benefits are potentially unlocking further optimisations after simplification (which means better optimised output code), and less work for further traverals to do, due to the smaller resulting expression (which means faster compilation).

Currently, for interpolation, we just let assignment do the broadcasting.  Form assembly with fancy optimisations can be even more complicated.  For example, the mass matrix of a vector-CG element is the outer product of the scalar mass matrix with an identity matrix, which means we only broadcast along the diagonal loop, but not onto the entire (dense) result tensor.  Is this new use case somehow more special than the previous ones?

miklos1:
> I am currently putting together a tensor of weights `Q` which I right multiply by a vector of function point evaluations (the function of interest here being whatever the ufl expression we are compiling with `gem_expr = fem.compile_ufl(ufl_expression,...)` is and the points are some `FInAT.PointSet` `point_set`).
> 
> If the `ufl_expression` function/expression happens to not be dependent on the points given to it, instead of getting `gem_expr` to be a vector of point evaluations (where some of the free indices are equal to `point_set.indices`) I get a scalar. To do my tensor contraction I then need to have a special case to deal with this quirk of `fem.compile_ufl`.
> 
> It seems to me better to get `fem.compile_ufl` to return something with the expected free indices.

I don't really see why some missing indices would get in the way of some tensor contraction.  Since you say in the linked comment that `Q` is compile-time constant, I would go even further: check which point indices are not being used in the result of `compile_ufl`, and pre-contract the `Q` matrix along those indices.  Then you have spared some loops at run time.

dham:
> > I think the `IndexSum` is a red herring here. The question to ask yourself is, what are the shape and free indices of `delta(i, i)` and `one`. I think that the former is scalar with one free index and the latter is scalar with no free indices. This makes the simplification invalid.
> 
> Well, yes and no. From a formal point of view, you are right, since `Delta(i, i)` is an expression which contains `i` without `i` being bound, so it does have`i` as a free index, while `one` does not have `i` as a free index. However, they are substantially equivalent, since for all valid values of `i`: `Delta(i, i)` = `one`. So, considering the purpose of filling the result tensor with the correct values, replacing `Delta(i, i)` with `one` is legit.
> 
> To apply optimisations, that invariably simplify the expression, as early as possible, is _usually_ the right thing to do. The benefits are potentially unlocking further optimisations after simplification (which means better optimised output code), and less work for further traverals to do, due to the smaller resulting expression (which means faster compilation).
> 
> Currently, for interpolation, we just let assignment do the broadcasting. Form assembly with fancy optimisations can be even more complicated. For example, the mass matrix of a vector-CG element is the outer product of the scalar mass matrix with an identity matrix, which means we only broadcast along the diagonal loop, but not onto the entire (dense) result tensor. Is this new use case somehow more special than the previous ones?

Hmmm. The problem is that you get intermediate values that don't have the expected shape, and you then have to somehow implicitly know what has disappeared. In this case, compile_ufl is being used to evaluate an expression at a bunch of points. The expression might also have shape because it's vector or tensor-valued. If some of the shape/indices disappear because of this sort of index cancellation, how do I safely know which indices have disappeared? What if it's a 3D vector valued expression being evaluated at 3 points? 

I think for this to be safe you have to have explicit broadcasting of some sort. You can then drop the broadcasts before you go to Impero in order to avoid loops, but while you're manipulating GEM I think you need to maintain consistent shapes.

miklos1:
> The problem is that you get intermediate values that don't have the
> expected shape, and you then have to somehow implicitly know what has
> disappeared. In this case, compile_ufl is being used to evaluate an
> expression at a bunch of points. The expression might also have shape
> because it's vector or tensor-valued. If some of the shape/indices
> disappear because of this sort of index cancellation, how do I safely
> know which indices have disappeared? What if it's a 3D vector valued
> expression being evaluated at 3 points?

I think we shall consider whether shapes (say, `A` being a 3 x 3 matrix)
and free indices (say, `e` being a scalar expression with free indices
`i` and `j` which both have extent 3) are equivalent and interchangeable,
or somehow different.  Considering an expression in isolation, these two
ways of looking at tensor nature seem interchangeable, for they are
convertible: you can wrap up free indices into a tensor shape with
`ComponentTensor` and you can turn tensor shapes into scalars with free
indices by indexing (`Indexed`) them with ‘free’ (i.e. loop) indices.

This consideration, however, raises the question, why do we have both if
they are both expressing the same thing?  Is it not redundant to have
both?  Is it just an artefact of coming from UFL, which also has both?
There is more to it than that.

Shapes and free indices differ significantly when we look at them under
the aspect of substitutability (Ersetzbarkeit) in a surrounding
expression.  Consider an expression such as Ax, uA, or uAx, where A is a
matrix, and u and x are vectors.  Then, say, we are able to reduce the
rank of A (to a vector or a scalar), and we want to substitute this
“optimised” A into the original expression.  This cannot be done
correctly without updating the surrounding vector operations, that is
whether they are an inner product, outer product, MatVec, MatMult,
multiplication by scalar etc.

This is not a problem, however, for free indices, since indices are
explicitly named rather than merely having an implicit order.  Consider
u_i * A_{ij} * x_j.  If manage to reduce A’s rank, we can simply
substitute the result back, and u_i * a_i * x_j, u_i * a_j * x_j, or u_i
* a * x_j are all correctly substituted expressions because the matching
indices were explicitly named.  It could be that the result of such a
substitution gives reduced rank in the outer expression too, but that
just means we’ve done even more optimisation.

This is why GEM is absolutely strict about shapes; a scalar [shape =
()], a vector of length 1 [shape = (1,)], and a 1 x 1 matrix [shape =
(1, 1)] are all distinct shapes, and they are in no way interchangeable
(unlike MATLAB, for example).  Free indices, however, are handled in a
much more loose way.  In fact, the reason GEM exists, and I could not
just use a subset of UFL was that in UFL, substituting something with
additional free indices (say, a coefficient with something that depends
on the quadrature index) may be illegal, because in UFL free indices are
really just a different way of looking at shapes.

So the convention I have usually used in API design was to define as
shape that which is absolutely non-negotiable, and define as free
indices those which are possible dependencies on loop indices.

> If some of the shape/indices disappear because of this sort of index
> cancellation, how do I safely know which indices have disappeared?

For shapes, this ought not happen, because that’s problematic.  For free
indices: they are “named”, so you can just look at the expression and
tell which ones have disappeared.

> What if it's a 3D vector valued expression being evaluated at 3
> points?

It could be a reasonable approach to define the 3D vector value (or
whatever shape it has) as the shape of the expression, and communicate
the point dependency as free indices, so they can disappear as a
dependency if optimisation was successful.

ReubenHill:
We have come to the conclusion that this PR is not necessary, but that the behaviour of GEM with respect to shape being sacred and indices being discardable needs to be documented in a design document for TSFC. I will make an issue.

tsfc/driver.py
```diff
@@ -377,7 +377,11 @@ def compile_expression_dual_evaluation(expression, to_element, *,
             quad_rule = QuadratureRule(point_set, to_element._weights)
             config["quadrature_rule"] = quad_rule
 
+        # evaluate expression at the points specified by point_set
         expr, = fem.compile_ufl(expression, **config, point_sum=False)
+        # the point set free indices should now be free indices of the compiled
+        # expression
+        assert set(point_set.indices) <= set(expr.free_indices)
```
ReubenHill, line 384: @wence- there was some discussion yesterday of adding some kind of assertion along these lines, but I can't work out what it should be. Given that the point set indices _can_ disappear, and we don't know what other free indices there might be at this point, what's there left to assert?

wence-: What other free indices might there be?

wence-: I think `expr.free_indices \in point_set.indices`?

ReubenHill: I _think_ if your expression has an argument there is a free index for that. Might be remembering wrong though

wence-: Oh yeah, so what we're expecting is `expr.free_indices <= set(chain(point_set.indices, *argument_multiindices)))` I think

ReubenHill: See https://github.com/firedrakeproject/tsfc/pull/241

PR #241, opened 2021-02-17
check free indices after calling fem.compile_ufl

ReubenHill:
Following discussion in https://github.com/firedrakeproject/tsfc/pull/237 (specifically [this comment](https://github.com/firedrakeproject/tsfc/pull/237#discussion_r574501314)), this asserts that we get what we expect from fem.compile_ufl.


wence- (approved):
I think this looks right.

ReubenHill:
FYI I don't have permissions to merge

PR #247, opened 2021-05-07
Runtime point set FInAT QuadratureElement interpolation

ReubenHill:
Allows interpolation onto FInAT `QuadratureElement`s with runtime point set expressions as their quadrature rule point set. 
Note that runtime tabulation `gem.Variable`s should be named with the prefix `"rt_"` to trigger their runtime tabulation.
Also adds compatibility to the new FInAT QuadratureElement API introduced in https://github.com/FInAT/FInAT/pull/81 .

All these changes are temporary fixes until FInAT supports dual evaluation (see WIP #218), **with the exception** of the changes made to `compile_ufl` in`tsfc/fem.py` which fix `compile_ufl` such that context `kwargs` for a `GemPointContext` to be specified.

These changes are needed to allow interpolation from an arbitrary function function space onto a P0DG function space on a Firedrake `VertexOnlyMesh`.

This replaces #226 and sits on top of master, avoiding dependency on #218 (dual evaluation with FInAT which needs lots of work).

Linked PRs
 - https://github.com/firedrakeproject/firedrake/pull/2059
 - https://github.com/FInAT/FInAT/pull/81

### DIFF
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -204,8 +204,8 @@ def compile_integral(integral_data, form_data, prefix, parameters, interface, co
         config = kernel_cfg.copy()
         config.update(quadrature_rule=quad_rule)
         expressions = fem.compile_ufl(integrand,
-                                      interior_facet=interior_facet,
-                                      **config)
+                                      fem.PointSetContext(**config),
+                                      interior_facet=interior_facet)
         reps = mode.Integrals(expressions, quadrature_multiindex,
                               argument_multiindices, params)
         for var, rep in zip(return_variables, reps):
@@ -386,8 +386,8 @@ def compile_expression_dual_evaluation(expression, to_element, *,
             # UnknownPointSet point set used to create the
             # QuadratureElement rule.
             point_set = finat_to_element._rule.point_set
-            # config for fem.GemPointContext
             config.update(point_indices=point_set.indices, point_expr=point_set.expression)
+            context = fem.GemPointContext(**config)
         else:
             qpoints = []
             # Everything is just a point evaluation.
@@ -408,7 +408,9 @@ def compile_expression_dual_evaluation(expression, to_element, *,
                 quad_rule = QuadratureRule(point_set, to_element._weights)
                 config["quadrature_rule"] = quad_rule
 
-        expr, = fem.compile_ufl(expression, **config, point_sum=False)
+            context = fem.PointSetContext(**config)
+
+        expr, = fem.compile_ufl(expression, context, point_sum=False)
         # In some cases point_set.indices may be dropped from expr, but nothing
         # new should now appear
         assert set(expr.free_indices) <= set(chain(point_set.indices, *argument_multiindices))
@@ -437,10 +439,12 @@ def compile_expression_dual_evaluation(expression, to_element, *,
                     assert isnan(pts).all()
                     point_set = finat_to_element._rule.point_set
                     config.update(point_indices=point_set.indices, point_expr=point_set.expression)
+                    context = fem.GemPointContext(**config)
                 else:
                     point_set = PointSet(pts)
                     config.update(point_set=point_set)
-                expr, = fem.compile_ufl(expression, **config, point_sum=False)
+                    context = fem.PointSetContext(**config)
+                expr, = fem.compile_ufl(expression, context, point_sum=False)
                 # In some cases point_set.indices may be dropped from expr, but
                 # nothing new should now appear
                 assert set(expr.free_indices) <= set(chain(point_set.indices, *argument_multiindices))
--- a/tsfc/driver.py
+++ b/tsfc/driver.py
@@ -6,7 +6,7 @@ import sys
 from functools import reduce
 from itertools import chain
 
-from numpy import asarray, allclose
+from numpy import asarray, allclose, isnan
 
 import ufl
 from ufl.algorithms import extract_arguments, extract_coefficients
@@ -22,8 +22,9 @@ import FIAT
 from FIAT.reference_element import TensorProductCell
 from FIAT.functional import PointEvaluation
 
-from finat.point_set import PointSet
+from finat.point_set import PointSet, UnknownPointSet
 from finat.quadrature import AbstractQuadratureRule, make_quadrature, QuadratureRule
+from finat.quadrature_element import QuadratureElement
 
 from tsfc import fem, ufl_utils
 from tsfc.finatinterface import as_fiat_cell
@@ -286,7 +287,8 @@ def compile_expression_dual_evaluation(expression, to_element, *,
 
     # Just convert FInAT element to FIAT for now.
     # Dual evaluation in FInAT will bring a thorough revision.
-    to_element = to_element.fiat_equivalent
+    finat_to_element = to_element
+    to_element = finat_to_element.fiat_equivalent
 
     if any(len(dual.deriv_dict) != 0 for dual in to_element.dual_basis()):
         raise NotImplementedError("Can only interpolate onto dual basis functionals without derivative evaluation, sorry!")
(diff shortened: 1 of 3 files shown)


tsfc/kernel_interface/firedrake_loopy.py
```diff
@@ -153,14 +153,16 @@ def register_requirements(self, ir):
...
+    def construct_kernel(self, return_arg, impero_c, index_names, first_coefficient_fake_coords, other_args=None):
         """Constructs an :class:`ExpressionKernel`.
 
         :arg return_arg: loopy.GlobalArg for the return value
         :arg impero_c: gem.ImperoC object that represents the kernel
         :arg index_names: pre-assigned index names
         :arg first_coefficient_fake_coords: If true, the kernel's first
             coefficient is a constructed UFL coordinate field
+        :arg other_args: A list of other loopy.GlobalArgs to include in
+            the kernel (TEMPORARY).
```
ReubenHill, line 165: Add issue and remove `(TEMPORARY)`.

tsfc/kernel_interface/firedrake_loopy.py
```diff
@@ -171,7 +173,10 @@ def construct_kernel(self, return_arg, impero_c, index_names, first_coefficient_
         args.extend(self.kernel_args)
         for name_, shape in self.tabulations:
             args.append(lp.GlobalArg(name_, dtype=self.scalar_type, shape=shape))
-
+        # TODO: deal with this properly, probably by making sure self.tabulations has the correct stuff in it!
```
ReubenHill, line 176: remove comment

tsfc/kernel_interface/firedrake_loopy.py
ReubenHill: note ` - appended to the argument list in the order given`

tsfc/kernel_interface/firedrake_loopy.py
ReubenHill: Better yet - investigate this

tsfc/kernel_interface/firedrake_loopy.py
ReubenHill: (investigate first!)

tsfc/kernel_interface/firedrake_loopy.py
ReubenHill: Removed

tsfc/kernel_interface/firedrake_loopy.py
ReubenHill: removed

tsfc/driver.py
```diff
@@ -398,9 +414,18 @@ def compile_expression_dual_evaluation(expression, to_element, *,
             try:
                 expr, point_set = expr_cache[pts]
             except KeyError:
-                point_set = PointSet(pts)
+                if runtime_point_interpolation:
+                    assert isnan(pts).all()
```
ReubenHill, line 418: add comment, either here or at the top when getting the fiat equivalent - "Until FInAT dual evaluation is done, FIAT QuadratureElements with UnknownPointSingleton point sets advertise NaNs as their points for each node in the dual basis.

tsfc/driver.py
```diff
@@ -357,7 +359,26 @@ def compile_expression_dual_evaluation(expression, to_element, *,
...
+    # FInAT QuadratureElement will create the necessary gem code.
+    from finat.tensorfiniteelement import TensorFiniteElement
+    runtime_quadrature_rule = (
+        isinstance(finat_to_element, QuadratureElement) or
+        (
+            isinstance(finat_to_element, TensorFiniteElement) and
+            isinstance(finat_to_element.base_element, QuadratureElement)
+        ) and
+        isinstance(finat_to_element._rule.point_set.expression, gem.Variable)
+    )
```
miklos1, line 375: This is such an anti-pattern... can you refactor please?

tsfc/driver.py
```diff
@@ -357,7 +359,26 @@ def compile_expression_dual_evaluation(expression, to_element, *,
...
+
+    point_evaluation_nodes_only = (
+        all(isinstance(dual, PointEvaluation) for dual in to_element.dual_basis()) and not
+        runtime_quadrature_rule
+    )
+    if point_evaluation_nodes_only:
```
miklos1, line 381: And this is a similar, but pre-existing anti-pattern.  I wish to see this gone, too.

tsfc/driver.py
ReubenHill: To what? A `QuadratureElement` with `gem.Variable` expression defines doing runtime quadrature which, without changing quite a lot, requires flagging at the relevant places in the code. Might as well only create that flag once. Besides this is all a hack until this dual evaluation code gets sucked into FInAT (which is not the purpose of this PR).

tsfc/driver.py
ReubenHill: See previous comment - I'm not going to refactor what is going to be replaced soon anyway and is beyond the scope of this PR.

tsfc/driver.py
ReubenHill: Also writing "this is an anti pattern" is about as not-instructive as it gets: please at least tell me _what_ the anti pattern is that you have observed otherwise you might as well be telling me "you suck and I don't like it and I won't tell you how to improve yourself"

ReubenHill: Also writing "this is an anti pattern" is about as not-instructive as it gets: please at least tell me _what_ the anti pattern is that you have observed otherwise you might as well be telling me "you suck and I don't like it and I won't tell you how to improve yourself"

ReubenHill: @miklos1 sorry if that was a bit intense - I am trying very hard, am keen to learn, and definitely value your input (particularly given you are no longer actively working on TSFC and FInAT!) Just please do remember here that there is an actual human at the other end reading these comments and who's (often difficult and iterative) work you are criticising in a public forum. :-)

wence-: The pattern is, I guess, that everything that we're interpolating into `is-a` finat element. Ideally therefore once inside we shouldn't need to check what type of finat element it is, since all elements should offer the same API.

As you note, there was already a place where this is not true (below), in that the finat element comes in, but what we actually inspect is the FIAT element. Now, every FIAT element `has-a` dual, but not all duals are equivalent for the code generation path right now, so we special case the PointEvaluation-only cases.

With these new runtime quadrature elements, some finat elements now no longer satisfy the `has-a` fiat element property (at least, they have one, but it doesn't have a usable dual), so new special casing is introduced.

The bigger next step is to (as noted) refactor and provide a flat interface to dual evaluation from finat that is robust (and doesn't need special casing).

I think it makes sense to temporarily introduce this code (so that there's a testable code-path that new improvements can be judged against), with the intention of taking it out again once we have finat duals properly in place.

tsfc/fem.py
```diff
@@ -694,7 +694,13 @@ def take_singleton(xs):
...
 
 def compile_ufl(expression, interior_facet=False, point_sum=False, **kwargs):
-    context = PointSetContext(**kwargs)
+
+    # kwargs must be context kwargs for a PointSetContext or GemPointContext
+    point_expr = kwargs.get("point_expr")
+    if point_expr:
+        context = GemPointContext(**kwargs)
+    else:
+        context = PointSetContext(**kwargs)
```
wence-, line 703: I suspect it makes more sense to refactor and make the caller responsible for constructing the context. That way we're not reliant on `point_expr` never being a kwarg for the `PointSetContext` case.

So:

```
def compile_ufl(expression, context, interior_facet=False, point_sum=False):
   """Translate a UFL expression to GEM.

   :arg expression: ...
   :arg context: translation context.
   ...
   """
   ...
```

tsfc/driver.py
miklos1: [Switch-case code smell.](https://wiki.c2.com/?SwitchStatementsSmell)  The issue is that you are dispatching on class and object internals, and then you are doing things differently based on that.  Essentially you are observing the guts of a finite element because things are done at the wrong place.  If the internals of the finite element matter that much as to do what to do, then that functionality should live on the class, rather than here.

The canonical (or rather, mechanical) refactoring would be to ask the finite element whether its quadrature rule is run time, or even better, move all that functionality that depends on it being run time to the element classes.

miklos1: > Also writing "this is an anti pattern" is about as not-instructive as it gets: please at least tell me _what_ the anti pattern is that you have observed otherwise you might as well be telling me "you suck and I don't like it and I won't tell you how to improve yourself"

Sorry, since I generally don't have much time for code reviews, and tend to optimize and omit lengthy explanations unless they are proved to be necessary, such as being asked for.  This may sometimes come across as rude...

ReubenHill: Thanks for the input both, and for the lengthy explanations. So the conclusion seems to be that providing a FInAT dual evaluation interface will solve this (i.e. #218 ) so I will regard this conversation as resolved for now.

miklos1:
@wence-, @ReubenHill: So we all agree that some of these changes add more hacks and that they need to be refactored.  The disagreement seems to be whether to refactor immediately or later.  I am generally for gradual, incremental improvements, rather than piling hacks which will ostensibly be cleaned up by a grand refactoring that will happen at a later date. So I would immediately push core parts of dual evaluations onto the elements, even if the API introduced at this point does not handle every future case, it merely need to support current functionality.  This API could then be improved later as needed and fit, rather than piling hacks now and clearing them up later.

However, core parts of this are already implemented in the dual evaluation draft pull requests.  Is there anything wrong with those?  If those PRs merely contain extra stuff (beyond the basic refactoring) that is not ready yet, I would just separate the refactoring part from the rest, get that merged, and then this 0D-mesh interpolation could be added more clearly, without having introduce (temporary) hacks.

wence-:
> However, core parts of this are already implemented in the dual evaluation draft pull requests. Is there anything wrong with those?

Unfortunately the data structures exposed by the dual in those PRs need a lot of work. They unroll all the structure (which was the point in the first place).

ReubenHill:
> > However, core parts of this are already implemented in the dual evaluation draft pull requests. Is there anything wrong with those?
> 
> Unfortunately the data structures exposed by the dual in those PRs need a lot of work. They unroll all the structure (which was the point in the first place).

They also introduced a pretty horrendous performance bug which which I don't believe I have managed to fully expunge as yet. The whole thing was becoming a bit of a massive task and I wanted to separate out my immediate concern (allowing for compile time dual evaluation onto meshes meshes of disconnected vertices) from the broader dual-evaluation-in-FInAT work.

ReubenHill:
All comments addressed and all tests passing - ready for re review.

tsfc/fem.py
```diff
@@ -693,8 +693,17 @@ def take_singleton(xs):
...
 
 
-def compile_ufl(expression, interior_facet=False, point_sum=False, **kwargs):
-    context = PointSetContext(**kwargs)
+def compile_ufl(expression, context, interior_facet=False, point_sum=False):
+    """Translate a UFL expression to GEM.
+
+    :arg expression: The UFL expression to compile.
+    :arg context: translation context - either a :class:`GemPointContext`
+        or :class:`PointSetContext`
```
miklos1, line 701: I like that this `context` has been made explicit compared to the previous version.

miklos1:
I wanted to have a go at doing some of the essential refactorings that I have been suggesting, because I still think that this code could be in a better shape, and following the previous discussions that is how I could demonstrate that.  Or, perhaps, that is how I could come to realise that the suggested refactorings are, in fact, not that easy to do.  In any case, that is how I could bring this issue constructively forward.

However, I have not had the time, and will not have at least until Monday or Wednesday (still uncertain after that), so I am afraid you have to proceed with this without me.

dham (approved):

ReubenHill:
Cross branch references removed - awaiting policy decision on merge vs squash merge (merge preferable here)

PR #251, opened 2021-06-20
Benchmarking - Flop counter 

AWhitmell:
Created and implemented Flop counting functions with the end goal of creating a full benchmarking function within the Firedrake project


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.pythonPath": "/home/andrew/firedrake/bin/python3"
+}
```
wence-, line 3: Please don't commit editor-specific files. You can fix this by doing `git rm` of this file and committing that deletion (and then squashing with 992002d).

flops/flop_relationship.py
```diff
@@ -0,0 +1,79 @@
...
+def print_loopy(kernel):
+    print(loopy.generate_code_v2(kernel.ast).device_code())
+
+
+plt.figure(figsize=(14,7))
+plt.plot(K, new_flops, label='New')
+plt.plot(K, old_flops, label='Old')
+plt.xlabel("Degree", fontsize=16)
+plt.ylabel("Flops", fontsize=16)
+plt.legend(loc="best")
```
wence-, line 79: Can you turn this into a test (to live in `tests/test_flop_counts.py`)? It should assert that the impero flop counts match, sufficiently well, the loopy counts.

gem/flop_count.py
```diff
@@ -0,0 +1,172 @@
+import gem.gem as gem
```
wence-, line 1: Short module-level docstring?

wence- (changes requested):
Thanks, some minor comments.

Can you additionally migrate the `test_sum_factorisation.py` flop counting to use these new flop counts (rather than the COFFEE `EstimateFlops` visitor)?

wence-:
We can redo the commit history to remove the PDF and example files separately.

wence-:
Can you fix the flake8 linting errors? See https://github.com/firedrakeproject/tsfc/pull/251/checks?check_run_id=2873140680

wence-:
Wrong source branch, see #252.

PR #252, opened 2021-06-23
Andrew/benchmarking

AWhitmell:


tests/test_sum_factorisation.py
```diff
@@ -172,4 +173,4 @@ def test_vector_laplace_action(cell, order):
 if __name__ == "__main__":
     import os
     import sys
-    pytest.main(args=[os.path.abspath(__file__)] + sys.argv[1:])
+    pytest.main(args=[os.path.abspath(__file__)] + sys.argv[1:])
```
wence-, line 176: Don't remove the trailing newline.

wence-:
Can you check the `tests/test_count_flops.py` test? it was written when `count_flops` was expecting a list of `gem` expressions, but now it expects a single `Impero` tree. It might no longer be possible to do this.

gem/flop_count.py
```diff
@@ -0,0 +1,192 @@
...
+        return flops(expression, parameters)
+
+
+def count_flops(impero_c):
+    """An approximation to flops required for a scheduled impero_c tree.
+
+    :arg impero_c: a :class:`~.Impero_C` object.
+    :returns: approximate flop count for the tree.
+    """
+    return statement(impero_c.tree, impero_c)
```
wence-, line 192: I think not being able to count flops should be a soft failure, so can you wrap this in a try/except block catching the possible exceptions it might raise? return zero in the except.

wence- (dismissed):
I think this is now good, will just check the firedrake test suite then merge. Thanks!

wence- (dismissed):
I think I have squashed the remaining bugs.

wence- (approved):

PR #257, opened 2021-08-19
Propagate argument information for kernel

connorjward:
The key thing I am trying to accomplish here is to add an `arguments` attribute to the TSFC kernel that Firedrake can query to determine codegen-type information to pass through to PyOP2 (`dim` and `arity` mainly).

I have added a `KernelArg` class hierarchy to encapsulate this information because it allows us to do some nice stuff with `functools.singledispatch` in Firedrake. We can straightforwardly convert between TSFC argument types and PyOP2 argument types in a way where the argument ordering does not matter (!) (e.g. [here](https://github.com/firedrakeproject/firedrake/pull/2172/files#diff-a17f9f7d347d1e36d5b04e849d399f23cfbfdb41a93d88ebad3cf4b4b0638ffcR750)).

This PR goes with https://github.com/OP2/PyOP2/pull/624 and https://github.com/firedrakeproject/firedrake/pull/2172.

Any comments would be appreciated. Crucially, am I on the right track here?


tsfc/finatinterface.py
```diff
@@ -330,3 +330,18 @@ def create_base_element(ufl_element, **kwargs):
...
+
+def split_shape(finat_element):
+    """Split a FInAT element's index_shape into its 'basis' and 'node' shapes where the
+    former describes the number and layout of nodes and the latter describes the local
+    shape at each node.
+    """
+    # Only tensor elements have node-local shape
+    if isinstance(finat_element, finat.TensorFiniteElement):
+        basis_shape = finat_element.index_shape[:-len(finat_element._shape)]
+        node_shape = finat_element._shape
```
wence-: I think this is wrong if transpose=True is passed to the TFE constructor?

Probably breaks too if one can make enriched of tensor fe, but that is verboten right now.

tsfc/kernel_args.py
```diff
@@ -0,0 +1,76 @@
...
+class TabulationKernelArg(KernelArg):
+    ...
+
+
+class ExteriorFacetKernelArg(KernelArg):
+    ...
+
+
+class InteriorFacetKernelArg(KernelArg):
+    ...
```
wence-, line 76: Would it make sense for these to all be enum instances, rather than classes? Then you can handle the cases in a match/case and static analysis can say something useful about whether everything is handled.

connorjward: Possibly. I chose this route for two reasons:
- I can use `functools.singledispatch` to handle them.
- I can imagine a future where TSFC might attach more functionality to these classes (e.g. to generate the loopy arg from a FInAT element).

tsfc/finatinterface.py
connorjward: I've removed this change. You must have been looking at an older commit.

dham (approved):

PR #258, opened 2021-08-19
optimise: option to ignore indices in sum factorisation

wence-:
For optimisation of dual evaluation, we want to sum factorise a
contraction, but know some sets of indices we just need to treat with
delta elimination. Allow caller to exclude these indices from the
factorisation routines to avoid N! behaviour for large N.

### DIFF
--- a/gem/optimise.py
+++ b/gem/optimise.py
@@ -481,12 +481,17 @@ def traverse_sum(expression, stop_at=None):
     return result
 
 
-def contraction(expression):
+def contraction(expression, ignore=None):
     """Optimise the contractions of the tensor product at the root of
     the expression, including:
 
     - IndexSum-Delta cancellation
     - Sum factorisation
+
+    :arg ignore: Optional set of indices to ignore when applying sum
+        factorisation (otherwise all summation indices will be
+        considered). Use this if your expression has many contraction
+        indices.
 
     This routine was designed with finite element coefficient
     evaluation in mind.
@@ -498,7 +503,12 @@ def contraction(expression):
     def rebuild(expression):
         sum_indices, factors = delta_elimination(*traverse_product(expression))
         factors = remove_componenttensors(factors)
-        return sum_factorise(sum_indices, factors)
+        if ignore is not None:
+            extra = tuple(i for i in sum_indices if i in ignore)
+            to_factor = tuple(i for i in sum_indices if i not in ignore)
+            return IndexSum(sum_factorise(to_factor, factors), extra)
+        else:
+            return sum_factorise(sum_indices, factors)
 
     # Sometimes the value shape is composed as a ListTensor, which
     # could get in the way of decomposing factors.  In particular,


gem/optimise.py
```diff
@@ -498,7 +503,12 @@ def contraction(expression):
     def rebuild(expression):
         sum_indices, factors = delta_elimination(*traverse_product(expression))
```
ReubenHill, line 504: Given your commit comment
> we want to sum factorise a
contraction, but know some sets of indices we just need to treat with
delta elimination. 

why doesn't this line already remove the indices that you want to ignore?

wence-: I do want to do delta elimination for all indices.

wence-: and the indices that I'm ignoring for the purposes of sum factorisation I still need lying around so that I can contract over them.

ReubenHill: Okay, and I presume the issue here is that `optimise.contraction` is called too early? I.e. before the delta elimination of all the indices occurs?

wence-: No, `gem.optimise.contraction` effectively does a slightly generalised version of:

```
contraction_indices, factors = delta_elimination(expr)
expr = sum_factorise(factors, contraction_indices)
```

But, the implementation of `sum_factorise` determines the best contraction order by considering all N! permutations of the `contraction_indices`. So it has a cut-off of "if you asked me to optimise more than 6 indices, I will complain".

In the cases we want to optimise here, I think that the caller of `gem.optimise.contraction` has information which allows us to know which indices we don't think are important and therefore indicate to the sum factorisation routine that they should not be considered.

Recall that `sum_factorise(factors, indices)` does an operation minimal implementation of `IndexSum(reduce(Product, factors), indices)`, so if we split the indices into ones we want to optimise and a rest then we need to do `IndexSum(sum_factorise(factors, to_optimise), rest)`.

ReubenHill: if `expr` has deltas in it (which is what we worry about here - when the shape indices of `Q` in your new FInAT dual evaluation changes are shape indices of deltas/identities), why can't we just call `delta_elimination(expr)` before calling `optimise.contraction(expr)`?

Or perhaps I'm still not understanding?

ReubenHill: PS I think the commit message ought to be reworded to make it clear that this avoids hitting the sum-factorisation index limit

wence-: shape indices do not always come from deltas, sometimes they come from the finite element actually having shape. I ran into this problem interpolating into a Regge space from a TensorFiniteElement.

ReubenHill: Okay then that seems reasonable.

ReubenHill (approved):
I'm happy. Just need to reword the commit message to make it clear that this avoids hitting the sum-factorisation index limit but I can do this once merged

PR #262, opened 2021-09-02
Update interpolation API

wence-:
Extend the classes of pullbacks handled in interpolation to include all tensor elements.

To do this we need to send the UFL element into `compile_expression` rather than the finat element (since finat elements don't know about symmetries [they don't need to]).

Linked PRs

- https://github.com/FEniCS/ufl/pull/68
- https://github.com/firedrakeproject/firedrake/pull/2202


ReubenHill:
What problem is this intended to solve?

wence-:
If you want a `VectorElement(RaviartThomas, ...)` which is used in mixed formulations of elasticity with weakly imposed symmetry, then you probably want to be able to interpolate into that space.

tsfc/ufl_utils.py
```diff
@@ -392,45 +400,73 @@ def apply_mapping(expression, element, domain):
...
+            # For each unique piece in reference space
+            if ri in seen:
+                continue
+            seen.add(ri)
+            # Get the physical space piece
+            piece = [flat[gm*gi + j] for j in range(gm)]
+            piece = as_tensor(numpy.asarray(piece).reshape(vs))
+            # get into reference space
+            piece = apply_mapping(piece, sub_elem, mesh)
+            assert piece.ufl_shape == rvs
```
ksagiyam, line 463: So this is a check for vector valued elements on manifold, I think.

ksagiyam (approved):
Looks very comprehensive to me.

tsfc/ufl_utils.py
```diff
@@ -345,12 +345,23 @@ def simplify_abs(expression, complex_mode):
...
+    :arg domain: Optional domain to provide in case expression does
+         not contain a domain (used for constructing geometric quantities).
+
+    :arg expression: The expression to map with shape element.value_shape()
+    :arg element: The UFL element of the target space
+    :arg domain: domain to consider.
+    :returns: A new UFL expression with shape element.reference_value_shape()
+    :raises NotImplementedError: If we don't know how to apply the
+        inverse of the pullback.
+    :raises ValueError: If we get shape mismatches.
```
ReubenHill, line 364: looks like you've accidentally entered the list of arguments twice here

PR #267, opened 2022-02-15
TSFC loopy: Profile local kernels.

sv2518:


tsfc/loopy.py
```diff
@@ -15,11 +15,41 @@
...
+        # Petsc functions
+        start_event = "PetscLogEventBegin"
+        end_event = "PetscLogEventEnd"
+        register_event = "PetscLogEventRegister"
+        # The variables
+        event_id_var_name = "event_id_" + kernel_name
+        event_name = "Log_Event_" + kernel_name
+        # Logging registration # TODO offport this so we don't register every single time
+        prepend = [lp.CInstruction("", "PetscLogEvent "+event_id_var_name+";"),
+                   lp.CInstruction("", register_event+"(\""+event_name+"\", PETSC_OBJECT_CLASSID, &"+event_id_var_name+");")]
```
sv2518, line 37: Depending on if we decide that the faffing with declaring the events outside the C kernels in python and setting the variables in the DLLs is worth it, we want to remove declaration in 36 and the registration in 37 and instead declare the event in the preamble.

tsfc/loopy.py
```diff
@@ -221,6 +254,9 @@ def generate(impero_c, args, scalar_type, kernel_name="loopy_kernel", index_name
                            seq_dependencies=True, silenced_warnings=["summing_if_branches_ops"],
                            lang_version=(2018, 2))
 
+    if PETSc.Log.getActive():
+        knl = lp.register_preamble_generators(knl, [_PreambleGen("#include <petsclog.h>")])
```
sv2518, line 258: Depending on if we decide that the faffing with declaring the events outside the C kernels in python and setting the event ids from the outside in the DLLs is worth it, we want to declare the events here in the preamble.

PR #268, opened 2022-02-16
Remove coefficients from ExpressionKernel

connorjward:
This should be useful to do because UFL coefficients are large data structures which until now have inhibited caching of the generated `ExpressionKernel`.

### DIFF
--- a/tests/test_dual_evaluation.py
+++ b/tests/test_dual_evaluation.py
@@ -12,7 +12,7 @@ def test_ufl_only_simple():
     W = V
     to_element = create_element(W.ufl_element())
     kernel = compile_expression_dual_evaluation(expr, to_element, W.ufl_element())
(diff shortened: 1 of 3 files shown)


wence- (dismissed):
Yes I think this looks right.

wence-:
Does this need to go concurrently with something in Firedrake?

connorjward:
> Does this need to go concurrently with something in Firedrake?

Yes. https://github.com/firedrakeproject/firedrake/pull/2348/commits/1c18e827d319e57fbdbb3e581c411ed779ef1cc8 from https://github.com/firedrakeproject/firedrake/pull/2348.

connorjward:
It also needs this [bugfix](https://github.com/OP2/PyOP2/pull/650)

sv2518 (approved):

PR #271, opened 2022-03-01
Ksagiyam/only use enabled components

ksagiyam:
See https://github.com/firedrakeproject/firedrake/pull/2364.

### DIFF
--- a/gem/gem.py
+++ b/gem/gem.py
@@ -22,7 +22,7 @@ from numbers import Integral, Number
 import numpy
 from numpy import asarray
 
-from gem.node import Node as NodeBase
+from gem.node import Node as NodeBase, traversal
 
 
 __all__ = ['Node', 'Identity', 'Literal', 'Zero', 'Failure',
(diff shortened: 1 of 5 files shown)


gem/gem.py
```diff
@@ -1041,3 +1041,12 @@ def as_gem(expr):
...
         raise ValueError("Do not know how to convert %r to GEM" % expr)
+
+
+def collect_node_type(expressions, node_type):
+    """Collects Node objects of type node_type in expressions."""
+    vset = set()
+    for node in traversal(expressions):
+        if isinstance(node, node_type):
+            vset.add(node)
+    return tuple(vset)
```
wence-, line 1052: Call this `extract_type` and `traversal` is already unique so:
```
tuple(node for node in traversal(expressions) if isinstance(node, klass))
```

is sufficient.

ksagiyam: Thanks! I'll fix.

tsfc/kernel_interface/firedrake_loopy.py
```diff
@@ -311,16 +322,35 @@ def construct_kernel(self, name, ctx):
...
+        funarg = self.generate_arg_from_expression(expr)
+        args.append(kernel_args.CoordinatesKernelArg(funarg))
         if oriented:
-            args.append(self.cell_orientations_arg)
+            funarg = self.generate_arg_from_expression(self._cell_orientations, dtype=numpy.int32)
+            args.append(kernel_args.CellOrientationsKernelArg(funarg))
         if needs_cell_sizes:
-            args.append(self.cell_sizes_arg)
-        args.extend(self.coefficient_args)
+            funarg = self.generate_arg_from_expression(self._cell_sizes)
```
sv2518, line 341: Can you add a comment?

PR #286, opened 2023-01-24
Fix interpolation for Arguments of zany spaces

wence-:
Previously we only checked if any coefficients in the source expression needed a coordinate mapping, but we should also check if any arguments do.

### DIFF
--- a/tests/test_dual_evaluation.py
+++ b/tests/test_dual_evaluation.py
@@ -59,3 +59,13 @@ def test_ufl_only_shape_mismatch():
     assert to_element.value_shape == (2,)
     with pytest.raises(ValueError):
         compile_expression_dual_evaluation(expr, to_element, W.ufl_element())
+
+
+def test_interpolate_zany_argument():
+    mesh = ufl.Mesh(ufl.VectorElement("P", ufl.triangle, 2))
+    V = ufl.FunctionSpace(mesh, ufl.FiniteElement("P", ufl.triangle, 1))
+    Q = ufl.FunctionSpace(mesh, ufl.FiniteElement("Argyris", ufl.triangle, 5))
+    expr = ufl.TestFunction(Q)
+    to_element = create_element(V.ufl_element())
+    kernel = compile_expression_dual_evaluation(expr, to_element, V.ufl_element())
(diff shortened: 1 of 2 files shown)


tsfc/fem.py
```diff
@@ -246,12 +248,64 @@ def physical_vertices(self):
...
+
+@needs_coordinate_mapping.register(finat.mixed.MixedSubElement)
+def _needs_coordinate_mapping_finat_mixed(element):
+    return needs_coordinate_mapping(element.element)
+
+
+@needs_coordinate_mapping.register(finat.HDivElement)
+@needs_coordinate_mapping.register(finat.HCurlElement)
+def _needs_coordinate_mapping_finat_hdivcurl(element):
+    return needs_coordinate_mapping(element.wrappee)
```
wence-, line 308: I put this here because it meant I didn't need a PR in finat as well, but probably finat should offer a `traverse` function that yields all the elements in a structured element. Everything needs special-casing because elements don't have regular "children" slots.

e.g. ideally you would just write (if you don't care about ordering)

```python
def traverse(element):
    yield element
    yield from map(traverse, element.children)
```

tests/test_dual_evaluation.py
```diff
@@ -59,3 +59,16 @@ def test_ufl_only_shape_mismatch():
...
+                         [lambda x: x, ufl.VectorElement, ufl.TensorElement],
+                         ids=["scalar", "vector", "tensor"])
+def test_interpolate_zany_argument(modifier):
+    mesh = ufl.Mesh(ufl.VectorElement("P", ufl.triangle, 2))
+    V = ufl.FunctionSpace(mesh, modifier(ufl.FiniteElement("P", ufl.triangle, 1)))
+    Q = ufl.FunctionSpace(mesh, modifier(ufl.FiniteElement("Argyris", ufl.triangle, 5)))
+    expr = ufl.TestFunction(Q)
+    to_element = create_element(V.ufl_element())
+    kernel = compile_expression_dual_evaluation(expr, to_element, V.ufl_element())
+    assert kernel.needs_external_coords
```
wence-, line 74: Not checking for correctness, just that a kernel actually comes out the other end.

PR #288, opened 2023-02-14
[Don't merge yet] Replace ufl_domain with extract_unique_domain

APaganini:
and ufl.domains() with extract_domains(expr). Both extract_domains and extract_unique_domain() are to be imported from ufl.domain. 

This pull request is the tsfc counterpart of [firedrake/firedrake#2754](https://github.com/firedrakeproject/firedrake/pull/2754)


JDBetteridge (approved):
LGTM

PR #294, opened 2023-05-10
Constant refactoring

JDBetteridge:


tsfc/kernel_interface/common.py
```diff
@@ -432,6 +435,24 @@ def check_requirements(ir):
     return cell_orientations, cell_sizes, tuple(sorted(rt_tabs.items()))
 
 
+def prepare_constant(constant):
+    """Bridges the kernel interface and the GEM abstraction for
+    CONSTANTS.
+
+    FIXME
```
dham, line 442: FIXME

tsfc/kernel_interface/firedrake_loopy.py
```diff
@@ -352,6 +368,7 @@ def construct_kernel(self, name, ctx, log=False):
         # in the list of kernel arguments by investigating
         # `Kernel.coefficient_numbers`.
         # Add return arg
+        # import pdb; pdb.set_trace()
```
dham, line 371: !!!!!!!!!!!!!

tsfc/driver.py
```diff
@@ -142,6 +142,7 @@ def compile_integral(integral_data, form_data, prefix, parameters, interface, *,
     builder.set_coordinates(mesh)
     builder.set_cell_sizes(mesh)
     builder.set_coefficients(integral_data, form_data)
+    builder.set_constants(form_data.constants)  # FIXME: should attach constants to integral data instead
```
connorjward, line 145: This could do with a more instructive comment about why this is the case. We want to attach to integral data since we don't want to be passing constants to kernels that don't need them. This is also probably a `TODO` instead of a `FIXME`.

Aside: I would love to have this refactored at some point so we don't need to distinguish here between coefficients and constants. At the point where we care about them as kernel arguments they are equivalent.

tsfc/kernel_interface/common.py
```diff
@@ -47,6 +47,9 @@ def __init__(self, scalar_type, interior_facet=False):
         # Coefficients
         self.coefficient_map = collections.OrderedDict()
 
+        # not coefficients
```
connorjward, line 50:
```suggestion
        # Constants
```

tsfc/ufl_utils.py
```diff
@@ -467,3 +475,11 @@ def apply_mapping(expression, element, domain):
...
         raise ValueError(f"Mismatching reference shapes, got {rexpression.ufl_shape} expected {element.reference_value_shape()}")
     return rexpression
+
+
+class TSFCConstantMixin:
+    """ Mixin class to identify Constants """
+
+    def __init__(self):
+        """ Set boolean """
+        self.TSFCConstant = True
```
connorjward, line 485: Is this needed? Surely `isinstance` is sufficient for any sort of use case.

connorjward (changes requested):

PR #309, opened 2024-03-27
introduce MixedMesh

ksagiyam:


gem/gem.py
```diff
@@ -1044,3 +1044,17 @@ def as_gem(expr):
...
+    variables = set()
+    for node in traversal(expressions):
+        if isinstance(node, Variable):
+            variables.add(node)
+        elif isinstance(node, Indexed):
+            for idx in node.multiindex:
+                if isinstance(idx, VariableIndex):
+                    v, = extract_variables((idx.expression, ))
+                    variables.add(v)
+    return tuple(variables)
```
wence-, line 1060: sets are not ordered. Is that a problem?

ksagiyam: No. Fixed.

PR #312, opened 2024-05-06
Add H(curl div) sobolev space and the covariant contravariant Piola mapping

pbrubeck:


rckirby (approved):
Seems right as long as the Firedrake tests pass?

PR #316, opened 2024-08-19
Cleanup variants

pbrubeck:
Tidying up and adding "demkowicz" and "fdm" variants for DG on simplices.


rckirby (approved):
I think this is fine?  Do we have any test coverage to make sure we're not breaking anything?

rckirby:
> I think this is fine? Do we have any test coverage to make sure we're not breaking anything?

https://github.com/firedrakeproject/firedrake/pull/3741

PR #317, opened 2024-09-03
generalise VariableIndex and FlexiblyIndexed

ksagiyam:
Generalise:

- `VariableIndex` to allow for `VariableIndex(some_general_expr)`,
- `FlexiblyIndexed` to allow for `FlexiblyIndexed((offset_expr, ((index_or_variableindex, stride_expr), )), ...)`.

The above changes then let us permute quadrature points based on the provided orientation data, giving us a way to enable interior facet integration on hex.

Potential TODO:

- enable full index arithmetic (and retire `FlexiblyIndexed`).
- attach `dtype` to each `Terminal` and let operators inherit `dtype` from children.

Depend on :
https://github.com/FInAT/FInAT/pull/138
https://github.com/firedrakeproject/fiat/pull/83

### DIFF
--- a/gem/node.py
+++ b/gem/node.py
@@ -2,6 +2,7 @@
 expression DAG languages."""
 
 import collections
+import gem
 
 
 class Node(object):
@@ -99,8 +100,23 @@ class Node(object):
         return hash((type(self),) + self._cons_args(self.children))
 
 
+def _make_traversal_children(node):
+    if isinstance(node, (gem.Indexed, gem.FlexiblyIndexed)):
+        # Include child nodes hidden in index expressions.
+        return node.children + node.indirect_children
+    else:
+        return node.children
+
+
 def pre_traversal(expression_dags):
-    """Pre-order traversal of the nodes of expression DAGs."""
+    """Pre-order traversal of the nodes of expression DAGs.
+
+    Notes
+    -----
+    This function also walks through nodes in index expressions
+    (e.g., `VariableIndex`s); see ``_make_traversal_children()``.
+
+    """
     seen = set()
     lifo = []
     # Some roots might be same, but they must be visited only once.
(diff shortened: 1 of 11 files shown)


tsfc/kernel_interface/firedrake_loopy.py
```diff
@@ -259,14 +259,26 @@ def __init__(self, integral_data_info, scalar_type,
...
+            self._entity_orientation = {
+                '+': gem.VariableIndex(gem.Indexed(facet_orientation, (0,))),
+                '-': gem.VariableIndex(gem.Indexed(facet_orientation, (1,)))
+            }
         elif integral_type == 'interior_facet_horiz':
             self._entity_number = {'+': 1, '-': 0}
+            facet_orientation = gem.Variable('facet_orientation', (1,), dtype=gem.uint_type)  # base mesh entity orientation
+            self._entity_orientation = {
+                '+': gem.VariableIndex(gem.Indexed(facet_orientation, (0,))),
+                '-': gem.VariableIndex(gem.Indexed(facet_orientation, (0,)))
```
dham, line 280: are these really both zero?

dham (changes requested):

tsfc/kernel_interface/firedrake_loopy.py
ksagiyam: They indeed are. This is for `interior_facet_horiz`, so when we walk up the column, facet orientation is always the same as the base cell orientation.

gem/node.py
```diff
@@ -99,6 +100,13 @@ def get_hash(self):
         return hash((type(self),) + self._cons_args(self.children))
 
 
+def _make_traversal_children(node):
+    if isinstance(node, (gem.Indexed, gem.FlexiblyIndexed)):
+        return node.children + node.indirect_children
+    else:
+        return node.children
```
wence-, line 107: Note: I'm not sure there is necessarily a better way of doing what you're trying to do, but the fact that the `indirect_children` must be handled specially in every traversal is slightly concerning. It's very easy to forget to do it, I would think.

ksagiyam: I think the issue is that there is a gap between the primal DAG and the DAG inside `VariableIndex`, as `children` of `Indexed` object does not include `multiindex`.

I at least added some comments in the docstrings.

ksagiyam:
Reviewed in a Firedrake meeting.

PR #321, opened 2024-10-18
Remove old warning about COFFEE

connorjward:


JDBetteridge (approved):
The timing code nearby looks sus, but you didn't touch that so I won't request changes!

PR #326, opened 2024-11-07
Refine dtype logic for Power node

angus-g:
We know how to optimise `0 ** x` and `x ** 0` to a `Zero` or `Constant(1)` node, but we can't assume that `x` is an instance of `Constant` with a defined dtype.


ksagiyam:
Unfortunately, we need a more robust fix. Please see if https://github.com/firedrakeproject/tsfc/pull/327 fixes your issue.

PR #327, opened 2024-11-07
Ksagiyam/attach dtype to nodes

ksagiyam:
Follow up https://github.com/firedrakeproject/tsfc/pull/317.

Attach `dtype` consistently to every node, and fix treatment of `Zero()**{Min, Max}Value()`.

Add test in Firedrake: https://github.com/firedrakeproject/firedrake/pull/3855

### DIFF
--- a/gem/gem.py
+++ b/gem/gem.py
@@ -38,7 +38,7 @@ __all__ = ['Node', 'Identity', 'Literal', 'Zero', 'Failure',
            'Inverse', 'Solve', 'extract_type', 'uint_type']
 
 
-uint_type = numpy.uintc
+uint_type = numpy.dtype(numpy.uintc)
 
 
 class NodeMeta(type):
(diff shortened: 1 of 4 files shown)


gem/gem.py
```diff
@@ -388,18 +419,17 @@ class Remainder(Scalar):
     def __new__(cls, a, b):
         assert not a.shape
         assert not b.shape
-        # TODO: Attach dtype property to Node and check that
-        #       numpy.result_dtype(a.dtype, b.dtype) is uint type.
-        #       dtype is currently attached only to {Constant, Variable}.
+        dtype = Node.inherit_dtype_from_children([a, b])
+        if dtype != uint_type:
```
dham, line 423:
```suggestion
        if dtype is not uint_type:
```

dham (changes requested):
just fix the "!=" to "is not" for types.

ksagiyam:
It turned out that object equal caused some pickling tests to fail:
```
import numpy as np 
import pickle 


obj1 = np.dtype(int) 
obj2 = np.dtype(int) 
print(obj1 == obj2)  # True 
print(obj1 is obj2)  # True 
obj1_pickled = pickle.dumps(obj1) 
obj1_loaded = pickle.loads(obj1_pickled) 
print(obj1_loaded == obj1)  # True 
print(obj1_loaded is obj1)  # False 
```
So probably we should just use `==` instead of object equal.

dham (approved):

PR #330, opened 2024-11-18
Remove unnecessary loopy transform that produces warnings

connorjward:
I *think* that this will help fix a lot of the warnings we have been seeing.

### DIFF
--- a/tsfc/loopy.py
+++ b/tsfc/loopy.py
@@ -264,9 +264,6 @@ def generate(impero_c, args, scalar_type, kernel_name="loopy_kernel", index_name
         preambles=preamble
     )
 
-    # Prevent loopy interchange by loopy
-    knl = lp.prioritize_loops(knl, ",".join(ctx.index_extent.keys()))
-
     return knl, event_name
 
 


tsfc/loopy.py
```diff
@@ -264,9 +264,6 @@ def generate(impero_c, args, scalar_type, kernel_name="loopy_kernel", index_name
         preambles=preamble
     )
 
-    # Prevent loopy interchange by loopy
-    knl = lp.prioritize_loops(knl, ",".join(ctx.index_extent.keys()))
```
kaushikcfd, line 268: This could have some performance degradation. For example, for kernels of the form:
```python
u[n,j] = 1.
```
it is preferred to have "j"-loop nested within "n"-loop. This preferred nesting can be guaranteed by performing appropriate calls to `lp.prioritize_loops`. Without those priorities loopy might generated code with the loops permuted in an unfavorable manner.

connorjward: Ah good to know! I should dig into how we build this `ctx.index_extent` because it is including 'sibling' loop nests when it shouldn't.

connorjward:
Replaced by #333.

tsfc/loopy.py
connorjward: #333 has a 'proper' solution to this. Thanks for flagging it.

PR #333, opened 2024-11-27
Redo loopy loop priorities

connorjward:
The previous solution to avoiding loop interchanges did not account for having multiple loop nests.


pbrubeck (approved):
LGTM

